PR #8, opened 2026-07-21
feat(auth): harden credential and registry handling

### LINKED ISSUE #7: can't log into codex
On Ubuntu 24

 Signed in to OpenAI ChatGPT
│
▲  Could not save OAuth tokens to Keychain — Secret Service daemon is not running (start GNOME Keyring or KWallet)

camjac251:
## Summary

- Add an external credential-helper protocol with executable validation, bounded runtime, helper-bound references, and verified writes.
- Serialize registry writes and recover dead or expired lock holders without removing a replacement lock.
- Represent anonymous access explicitly so stale environment credentials cannot override no-auth providers, no placeholder secrets are persisted, and anonymous discovery omits credential headers.
- Normalize unambiguous existing no-auth entries and delete helper-backed credentials when their final provider reference is removed.

This combines the overlapping work from #13 and #14 into one branch as requested.

Fixes #7.

## Test plan

- [x] 96 affected tests across credential helpers, registry CRUD, anonymous endpoints, refresh, catalog display, and lock handling
- [x] 612 broader non-listener tests
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] GitHub Actions full test suite

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -154,10 +154,12 @@ const oauthRefreshInflight = new Map<string, Promise<string | null>>();
 export type ParsedAuthRef =
   | { kind: 'keyring'; account: string }
   | { kind: 'helper'; helperId: string; account: string }
-  | { kind: 'env'; varName: string };
+  | { kind: 'env'; varName: string }
+  | { kind: 'none' };
 
 /** Parse registry credential references. */
 export function parseAuthRef(authRef: string): ParsedAuthRef | null {
+  if (authRef === 'none:anonymous') return { kind: 'none' };
   if (authRef.startsWith('keyring:')) {
     const account = authRef.slice('keyring:'.length);
     return account ? { kind: 'keyring', account } : null;
@@ -319,10 +321,12 @@ export async function resolveProviderCredential(
   authRef: string,
   diag?: (msg: string) => void,
 ): Promise<string | null> {
+  const parsed = parseAuthRef(authRef);
+  if (parsed?.kind === 'none') return null;
+
   const namespaced = readEnvCredential(clodexKeyEnvVar(providerId));
   if (namespaced) return namespaced;
 
-  const parsed = parseAuthRef(authRef);
   if (!parsed) return null;
 
   if (parsed.kind === 'env') {
@@ -338,7 +342,12 @@ export async function resolveProviderOAuthAccountId(
   diag?: (msg: string) => void,
 ): Promise<string | undefined> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind === 'env' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
+  if (
+    !parsed
+    || parsed.kind === 'env'
+    || parsed.kind === 'none'
+    || !oauthProviderIdFromAccount(parsed.account)
+  ) return undefined;
   const raw = await readStoredCredential(parsed, diag);
   return parseStoredOAuthCredential(raw)?.accountId;
 }
@@ -348,7 +357,12 @@ export async function resolveProviderOAuthProviderData(
   diag?: (msg: string) => void,
 ): Promise<Record<string, unknown> | undefined> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind === 'env' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
+  if (
+    !parsed
+    || parsed.kind === 'env'
+    || parsed.kind === 'none'
+    || !oauthProviderIdFromAccount(parsed.account)
+  ) return undefined;
   const raw = await readStoredCredential(parsed, diag);
   return parseStoredOAuthCredential(raw)?.providerData;
 }
@@ -422,7 +436,7 @@ export async function saveProviderCredential(
   diag?: (msg: string) => void,
 ): Promise<boolean> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind === 'env') return false;
+  if (!parsed || parsed.kind === 'env' || parsed.kind === 'none') return false;
   const written = await writeStoredCredential(parsed, key, diag);
   if (!written) return false;
   const readBack = await readStoredCredential(parsed, diag);
@@ -437,7 +451,7 @@ export async function probeProviderCredentialStore(
   diag?: (msg: string) => void,
 ): Promise<boolean> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind === 'env') return false;
+  if (!parsed || parsed.kind === 'env' || parsed.kind === 'none') return false;
   const probeAccount = `${parsed.account}::probe::${randomUUID()}`;
   const probeRef: StoredCredentialRef = parsed.kind === 'helper'
     ? { kind: 'helper', helperId: parsed.helperId, account: probeAccount }
@@ -469,7 +483,6 @@ export async function deleteProviderCredential(
   diag?: (msg: string) => void,
 ): Promise<boolean> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind === 'env') return false;
+  if (!parsed || parsed.kind === 'env' || parsed.kind === 'none') return false;
   return deleteStoredCredential(parsed, diag);
 }
-
--- a/src/env.ts
+++ b/src/env.ts
@@ -1,5 +1,11 @@
 // src/env.ts
 import { CONFLICTING_ENV_VARS } from './constants.js';
+import { randomUUID } from 'node:crypto';
+import {
+  deleteCredentialHelperAccount,
+  readCredentialHelperAccount,
+  writeCredentialHelperAccount,
+} from './credential-helper.js';
 import { claudeCodeClientModelId, stripOneMContextSuffix } from './context-model-id.js';
 import { resolveContextWindow } from './context-window.js';
 import {
@@ -147,9 +153,10 @@ const oauthRefreshInflight = new Map<string, Promise<string | null>>();
 
 export type ParsedAuthRef =
   | { kind: 'keyring'; account: string }
+  | { kind: 'helper'; helperId: string; account: string }
   | { kind: 'env'; varName: string };
 
-/** Parse registry authRef strings like `keyring:provider:openai` or `env:OPENAI_API_KEY`. */
+/** Parse registry credential references. */
 export function parseAuthRef(authRef: string): ParsedAuthRef | null {
   if (authRef.startsWith('keyring:')) {
     const account = authRef.slice('keyring:'.length);
@@ -159,6 +166,8 @@ export function parseAuthRef(authRef: string): ParsedAuthRef | null {
     const varName = authRef.slice('env:'.length);
     return varName ? { kind: 'env', varName } : null;
   }
+  const helper = /^helper:v1:([0-9a-f]{64}):(.+)$/s.exec(authRef);
+  if (helper) return { kind: 'helper', helperId: helper[1]!, account: helper[2]! };
   return null;
 }
 
@@ -239,6 +248,7 @@ async function deleteKeyringAccount(account: string, diag?: (msg: string) => voi
   try {
     const { Entry } = await import('@napi-rs/keyring');
     const value = new Entry(KEYRING_SERVICE, account).getPassword();
+    if (value === null) return true;
     if (value?.startsWith(KEYRING_CHUNK_PREFIX)) {
       const chunkCount = Number(value.slice(KEYRING_CHUNK_PREFIX.length));
       for (let i = 0; i < chunkCount; i++) {
@@ -253,7 +263,57 @@ async function deleteKeyringAccount(account: string, diag?: (msg: string) => voi
   }
 }
 
-/** Resolve a provider secret from authRef (env → keyring). */
+type StoredCredentialRef = Extract<ParsedAuthRef, { kind: 'keyring' | 'helper' }>;
+
+function storedCredentialAuthRef(ref: StoredCredentialRef): string {
+  return ref.kind === 'helper'
+    ? `helper:v1:${ref.helperId}:${ref.account}`
+    : `keyring:${ref.account}`;
+}
+
+async function readStoredCredential(
+  ref: StoredCredentialRef,
+  diag?: (msg: string) => void,
+): Promise<string | null> {
+  if (ref.kind === 'keyring') return readKeyringAccount(ref.account, diag);
+  try {
+    return await readCredentialHelperAccount(ref.account, ref.helperId);
+  } catch (err) {
+    diag?.(err instanceof Error ? err.message : String(err));
+    return null;
+  }
+}
+
+async function writeStoredCredential(
+  ref: StoredCredentialRef,
+  value: string,
+  diag?: (msg: string) => void,
+): Promise<boolean> {
+  if (ref.kind === 'keyring') return writeKeyringAccount(ref.account, value, diag);
+  try {
+    await writeCredentialHelperAccount(ref.account, value, ref.helperId);
+    return true;
+  } catch (err) {
+    diag?.(err instanceof Error ? err.message : String(err));
+    return false;
+  }
+}
+
+async function deleteStoredCredential(
+  ref: StoredCredentialRef,
+  diag?: (msg: string) => void,
+): Promise<boolean> {
+  if (ref.kind === 'keyring') return deleteKeyringAccount(ref.account, diag);
+  try {
+    await deleteCredentialHelperAccount(ref.account, ref.helperId);
+    return true;
+  } catch (err) {
+    diag?.(err instanceof Error ? err.message : String(err));
+    return false;
+  }
+}
+
+/** Resolve a provider secret from a namespaced env var or its configured store. */
 export async function resolveProviderCredential(
   providerId: string,
   authRef: string,
@@ -269,7 +329,7 @@ export async function resolveProviderCredential(
     return readEnvCredential(parsed.varName);
   }
 
-  return readProviderSecret(parsed.account, diag);
+  return readProviderSecret(parsed, diag);
 }
 
 /** Read OAuth metadata retained alongside the access token. */
@@ -278,8 +338,8 @@ export async function resolveProviderOAuthAccountId(
   diag?: (msg: string) => void,
 ): Promise<string | undefined> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind !== 'keyring' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
-  const raw = await readKeyringAccount(parsed.account, diag);
+  if (!parsed || parsed.kind === 'env' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
+  const raw = await readStoredCredential(parsed, diag);
   return parseStoredOAuthCredential(raw)?.accountId;
 }
 
@@ -288,8 +348,8 @@ export async function resolveProviderOAuthProviderData(
   diag?: (msg: string) => void,
 ): Promise<Record<string, unknown> | undefined> {
   const parsed = parseAuthRef(authRef);
-  if (!parsed || parsed.kind !== 'keyring' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
-  const raw = await readKeyringAccount(parsed.account, diag);
+  if (!parsed || parsed.kind === 'env' || !oauthProviderIdFromAccount(parsed.account)) return undefined;
+  const raw = await readStoredCredential(parsed, diag);
   return parseStoredOAuthCredential(raw)?.providerData;
 }
 
@@ -309,13 +369,14 @@ function decodeProviderSecret(raw: string | null): string | null {
   return trimmed;
 }
 
-async function refreshOAuthKeyringAccount(
-  account: string,
+async function refreshOAuthStoredCredential(
+  ref: StoredCredentialRef,
   providerId: string,
   raw: string,
   diag?: (msg: string) => void,
 ): Promise<string | null> {
-  const existing = oauthRefreshInflight.get(account);
+  const authRef = storedCredentialAuthRef(ref);
+  const existing = oauthRefreshInflight.get(authRef);
   if (existing) return existing;
 
   const work = (async (): Promise<string | null> => {
@@ -326,7 +387,8 @@ async function refreshOAuthKeyringAccount(
     try {
       const refreshed = await refreshStoredOAuthCredential(providerId, cred);
       const json = oauthCredentialToKeychainJson(refreshed);
-      await writeKeyringAccount(account, json, diag);
+      const saved = await saveProviderCredential(authRef, json, diag);
+      if (!saved) throw new Error('Could not persist refreshed OAuth credential');
       return refreshed.access;
     } catch (err) {
       diag?.(err instanceof Error ? err.message : String(err));
@@ -335,21 +397,21 @@ async function refreshOAuthKeyringAccount(
     }
   })();
 
-  oauthRefreshInflight.set(account, work);
+  oauthRefreshInflight.set(authRef, work);
   try {
     return await work;
   } finally {
-    oauthRefreshInflight.delete(account);
+    oauthRefreshInflight.delete(authRef);
   }
 }
 
-async function readProviderSecret(account: string, diag?: (msg: string) => void): Promise<string | null> {
-  const raw = await readKeyringAccount(account, diag);
+async function readProviderSecret(ref: StoredCredentialRef, diag?: (msg: string) => void): Promise<string | null> {
+  const raw = await readStoredCredential(ref, diag);
   if (!raw) return null;
 
-  const oauthProviderId = oauthProviderIdFromAccount(account);
+  const oauthProviderId = oauthProviderIdFromAccount(ref.account);
   if (oauthProviderId && raw.trim().startsWith('{')) {
-    return refreshOAuthKeyringAccount(account, oauthProviderId, raw, diag);
+    return refreshOAuthStoredCredential(ref, oauthProviderId, raw, diag);
   }
   return decodeProviderSecret(raw);
 }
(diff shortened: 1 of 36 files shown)


src/env.ts
```diff
@@ -253,7 +263,57 @@ async function deleteKeyringAccount(account: string, diag?: (msg: string) => voi
...
+  }
+}
+
+async function deleteStoredCredential(
+  ref: StoredCredentialRef,
+  diag?: (msg: string) => void,
+): Promise<boolean> {
+  if (ref.kind === 'keyring') return deleteKeyringAccount(ref.account, diag);
+  try {
+    await deleteCredentialHelperAccount(ref.account, ref.helperId);
```
bman654, line 311: The helper delete path here is correct and complete — but nothing ever reaches it during provider removal.

`removeProviderFromRegistry` still gates credential deletion on the ref being a *keyring* ref:

```ts
// src/registry/crud.ts:41 — not touched by this PR
const parsed = parseAuthRef(removedProvider.authRef);
const shouldDelete = !credentialStillReferenced(removedProvider.authRef, registry.providers);
if (shouldDelete && parsed?.kind === 'keyring') {
  credentialDeleted = await deleteProviderCredential(removedProvider.authRef);
}
```

For a `helper:v1:...` ref that condition is false, so `deleteProviderCredential` is never called and this `deleteStoredCredential` branch never runs.

**Repro:** with `CLODEX_CREDENTIAL_HELPER` pointing at a helper, add a provider, then `clodex providers remove openai`. It exits 0 and reports success, but `provider:openai` is still present in the helper store — the API key (or OAuth refresh token) is orphaned, with no remaining registry reference and no CLI path to delete it.

**Suggested fix:** drop the `kind === 'keyring'` restriction so both kinds dispatch through `deleteProviderCredential` (keep the `credentialStillReferenced` shared-ref guard as-is).

Worth adding an integration test alongside it — the suite currently has no coverage for removing a helper-backed provider, which is why this slipped through.

bman654 (commented):
Reviewed this against `main` in an isolated worktree. `pnpm typecheck`, `pnpm build`, and `pnpm test` all pass (634 tests / 64 files, +14 over the 620/63 baseline), and I found no non-opt-in regression: with `CLODEX_CREDENTIAL_HELPER` unset, generated refs stay `keyring:`, and existing keyring reads plus the legacy `~/.relay-ai` fallback behave exactly as before.

I also smoke-tested the helper backend against a real fake helper script, including the failure paths, and the security posture looks good:

- an injection-shaped account name produced no shell side effect (`shell: false` confirmed)
- a helper exiting non-zero surfaced only `credential helper get failed with exit code 7` — secret markers on stdout/stderr did not leak, even under `--trace`
- a missing binary gave a clean `CLODEX_CREDENTIAL_HELPER is not an executable file`
- a hanging helper was killed by the timeout rather than wedging the invocation

One functional issue worth fixing before this lands — inline comment below on the removal path leaving credentials orphaned in the helper store.

Heads-up on coordination, unrelated to the code here: #13 and #14 independently rework the same registry/auth-ref surface as this PR (#13 adds a `kind:'none'` variant to the same union, #14 wraps several of these same functions in a registry write lock). This PR merges cleanly onto current `main`, but both of those will conflict once it lands and will need a rebase.

bman654 (changes requested):
Thanks for this submission.  can you fix the cleanup logic then I'll merge

src/env.ts
camjac251: The missing removal dispatch is fixed in `a37ade2`. Provider removal now sends every unreferenced auth reference through `deleteProviderCredential`, which rejects environment and no-auth references while handling keyring and helper backends. A real-helper integration test writes the credential, removes the provider, and verifies both the registry entry and helper-store value are gone.

The same branch now contains the overlapping #13 and #14 changes requested for a single review.

src/registry/lock.ts
```diff
@@ -0,0 +1,339 @@
...
+  const snapshot: RegistryLockSnapshot = {
+    raw,
+    device: stats.dev,
+    inode: stats.ino,
+    modifiedAt: stats.mtimeMs,
+  };
+  const owner = parseLockOwner(raw);
+  if (owner) {
+    const expired = now - owner.startedAt >= LOCK_MAX_HOLD_MS;
+    return alive(owner.pid) && !expired ? null : snapshot;
```
bman654: **Blocker — this evicts a *live* lock owner, and the evicted owner still writes. Net effect is a silent lost update.**

The age check is the right instinct, but nothing ties a subsequent write back to the lock instance that authorised it. Once a live owner is reaped at `LOCK_MAX_HOLD_MS`, a second writer acquires, and the original owner — still running, still holding state it loaded *before* eviction — proceeds to `saveRegistry` anyway.

Reproduced: live expired owner holding a pre-lock snapshot → second writer acquired and saved → original owner saved → **the second writer's update disappeared.**

**This one is my fault, and I'm sorry** — my review on #14 proposed almost exactly this code and asserted the existing raw/dev/inode/mtime snapshot made the reap safe. That was wrong. The snapshot prevents deleting the *wrong file*; it does nothing to stop an evicted owner from writing stale state. You implemented the suggestion faithfully and it was incomplete.

**Suggested fix — a fencing token, which also resolves the `removeStaleLock` race below.** Write a unique nonce into the lock record at acquisition and return it in the lease. Immediately before `saveRegistry`, re-read the lock and confirm the nonce (and inode) still match; if not, the lease was lost — throw rather than write.

```ts
// on acquire
const nonce = randomUUID();
writeLockRecord({ pid, startedAt: now, nonce });

// inside withRegistryWriteLock, immediately before the callback's write
if (readLockOwner(lockPath)?.nonce !== lease.nonce) {
  throw new RegistryLockLostError();   // detected + safe, instead of silent loss
}
```

A microscopic TOCTOU remains between validate and write, but this shrinks the window from "ten minutes of keychain prompt" to microseconds — and combined with moving credential I/O out of the critical section (see my comment on `crud.ts`), the realistic eviction scenario goes away entirely.

src/registry/lock.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  // A process can be between exclusive creation and writing its owner record.
+  // Only reap malformed locks after a long grace period.
+  return now - snapshot.modifiedAt >= INVALID_LOCK_STALE_MS ? snapshot : null;
+}
+
+function removeStaleLock(
```
bman654, line 203: **Blocker — check-then-unlink race: the reaper can delete a *replacement* lock.**

Between the `statSync()` snapshot comparison and `unlinkSync()`, the stale owner can release its lock and a brand-new owner can acquire one. The reaper then unlinks the new owner's lock, and two writers proceed believing they hold it.

`tryAcquireReaperGuard` correctly prevents reaper-vs-reaper collisions, but not reaper-vs-new-owner.

**Same fix as the fencing token above** — this is really one defect with two faces. Once every writer validates its nonce immediately before writing, a wrongly-deleted lock can no longer cause a silent double-write: only the process whose nonce is actually on disk proceeds, and the other fails loudly.

Worth fixing them together rather than separately; a purely path-based reap can't be made fully atomic, so the durable answer is making writes verify ownership rather than making deletion perfect.

src/registry/lock.ts
```diff
@@ -0,0 +1,339 @@
...
+  };
+  const owner = parseLockOwner(raw);
+  if (owner) {
+    const expired = now - owner.startedAt >= LOCK_MAX_HOLD_MS;
+    return alive(owner.pid) && !expired ? null : snapshot;
+  }
+
+  // A process can be between exclusive creation and writing its owner record.
+  // Only reap malformed locks after a long grace period.
+  return now - snapshot.modifiedAt >= INVALID_LOCK_STALE_MS ? snapshot : null;
```
bman654: **Major — a crash between exclusive creation and the owner-record write leaves an empty lock that blocks everything for 10 minutes.**

If a process dies in that window, the lock file exists but has no parseable owner, so it falls through to `INVALID_LOCK_STALE_MS` and is treated as validly occupied for the full grace period. Every writer then waits the 30s `DEFAULT_WAIT_MS` and fails.

The sharper consequence: an unmigrated-registry **read** synchronously acquires this lock, so during the outage a stalled lock adds ~30s to *every* registry-loading invocation. Measured: **30,026 ms**. That includes `clodex-claude`, which runs once per spawned agent process — so a single crash at the wrong moment stalls every agent launch for ten minutes.

**Suggested fix:** make acquisition atomic — write the owner record into a temp file and `rename()` it into place, so the lock either exists complete or doesn't exist at all. Failing that, drop `INVALID_LOCK_STALE_MS` for the empty case specifically: a zero-length lock has no owner to protect and can be reaped immediately.

src/registry/crud.ts
```diff
@@ -21,37 +22,40 @@ export async function removeProviderFromRegistry(
...
-
-  return {
-    removed: true,
-    id,
-    name: removedProvider.name,
-    credentialDeleted,
-  };
+
+    const [removedProvider] = registry.providers.splice(index, 1);
+    saveRegistry(registry);
```
bman654, line 50: **Major — credential I/O runs inside the registry write lock's critical section.**

`deleteProviderCredential` here, and `saveProviderCredential` at `add-template.ts:121` and `custom-endpoint.ts:190`, all execute while the lock is held. Both can be slow in ways the lock never anticipated:

- a macOS keychain unlock prompt is **unbounded user interaction** — it waits on a human
- a credential helper is a subprocess with a 10s timeout

While one process sits on that, concurrent clodex processes exhaust the 30s wait and fail. Worse, a prompt outliving `LOCK_MAX_HOLD_MS` triggers exactly the eviction scenario in my `lock.ts` comment — the lock is reaped from under a live holder, and the write that follows is unguarded.

`provider-auth.ts:159-165` already does this correctly: probe and save the credential **before** taking the lock, then hold the lock only for the registry mutation. Applying that pattern uniformly here would make the layering consistent and remove the slowest operations from the critical section.

**Separately, and explicitly *not* your bug:** the ordering in this function — `saveRegistry()` before credential deletion, with success reported regardless of the result — is pre-existing on `main` and behaves the same way there. It means a failed delete orphans the secret with no registry reference left to find it by. I'm tracking that as our own follow-up; mentioning it only so the two aren't confused. Happy for you to leave it alone.

src/registry/materialize.ts
```diff
@@ -83,9 +82,14 @@ export function cachedModelToLocal(
   };
 }
 
-function providerAllowsAnonymousFreeModels(provider: RegistryProvider): boolean {
-  const template = getTemplateById(provider.templateId) ?? getTemplateById(provider.id);
-  return template?.anonymousFreeModels === true;
+export function isAnonymousProvider(
```
bman654, line 85: **Minor — custom endpoints created by shipped clodex aren't matched by this, so they don't migrate.**

This requires `authType === 'none'`, but the entries that actually exist on disk today don't have it. On current `main`, `custom-endpoint.ts:152` turns a blank key into the literal string `'local'`, stores that in the keychain (`:209`), and persists an entry with **no `authType` field at all** (the `authType` at `:164` is set on the template object passed to the model fetcher, not on the saved entry — easy to misread).

So anyone who added an Ollama / LM Studio / vLLM endpoint with released clodex has: `authType` absent, `authRef` a normal keyring ref, credential `"local"`. Verified against a hand-written registry: the no-credential entry vanished from the catalog, and the `'local'` entry came back as an *authenticated* provider whose API key is the word `local`.

To be clear this is **not** relay-ai legacy state — it's current clodex format, so it won't be swept up by the legacy removal we're planning separately.

**Fix if it's easy:** also treat `authType` absent + credential `'local'` as anonymous, and normalise it to the new representation on load.

**If that turns out to be awkward, fail closed instead** — the probability of anyone hitting this is low, so a clear error beats a silent wrong state. Refuse to materialise the ambiguous entry and tell the user exactly how to recover, e.g.:

```
Provider 'my-ollama' uses an outdated no-auth format and can't be loaded.
Re-add it:  clodex providers remove my-ollama && clodex providers add
```

What we want to avoid is the current silent outcomes: disappearing from the catalog with no explanation, or showing up as authenticated with a bogus key that 401s at request time.

src/registry/provider-auth.ts
```diff
@@ -142,19 +152,26 @@ export async function authenticateProvider(
     throw new Error('OAuth sign-in is only available for openai (ChatGPT Plus/Pro).');
   }
 
+  const authRef = oauthAuthRef(registryId);
+  let storeDiagMsg = '';
+  const storeReady = await probeProviderCredentialStore(authRef, (msg) => { storeDiagMsg = msg; });
+  if (!storeReady) {
```
bman654, line 163: **Minor — behavior change worth documenting rather than reverting.**

This now hard-fails when the credential store probe fails, where `main` warned and continued with a non-persistent session. I think the new behavior is *better* — silently proceeding to lose credentials is worse than refusing — so I'd keep it.

But it does block users whose keychain is unavailable (headless Linux, CI containers, SSH sessions without an unlocked keyring) who previously got a degraded-but-working session.

**Suggestion — docs, not code:** have the error name the escape hatch, and cover it in `docs/credential-helpers.md` plus the release notes:

```
Credential store is unavailable — <diag>
Set CLODEX_CREDENTIAL_HELPER to use an external credential store instead.
```

That turns a dead end into a signpost, which is all this needs.

bman654 (changes requested):
thanks for combining.  Claude found a few more major edge cases and some minor documentation issues.

src/registry/lock.ts
camjac251: Fixed in `4cf938a`. Each acquired lock now exposes a lease carrying its UUID token plus device and inode identity. `saveRegistry` requires the matching active lease and validates it both before preparing the write and immediately before the rename. Unique per-write temporary files also prevent one writer from replacing another writer's prepared payload.

The regression keeps the successor lock active while an expired original owner resumes, then verifies that the stale save throws `RegistryLockLostError` and cannot overwrite the successor's registry state.

src/registry/lock.ts
camjac251: Handled by the same ownership fencing in `4cf938a`. Stale removal and normal lease release compare the token, device, and inode before unlinking. More importantly, every registry publish validates the current on-disk lease immediately before rename, so a writer whose lock was removed or replaced fails instead of silently publishing stale state.

The unavoidable filesystem stat-to-unlink micro-window remains, but it no longer permits an affected writer to commit without detecting the lost lease.

src/registry/lock.ts
camjac251: Fixed in `4cf938a`. Acquisition writes and fsyncs the complete owner record to a unique sibling inode, then uses a hard link to publish the canonical lock path atomically without replacing an existing owner. A crash before the link leaves the canonical path absent; a crash after it leaves a complete, parseable record.

Empty or malformed canonical locks are now recoverable immediately, and the lock tests include a zero-length recovery case.

src/registry/crud.ts
camjac251: Fixed in `4cf938a`. Slow keychain and helper operations no longer run under the global registry lock. Credential mutations use a separate hashed, per-reference cross-process lock, while the global lock is held only for short registry checks and publication.

The same-reference lock spans credential save plus registry commit for template, custom endpoint, and OAuth writes. Removal rechecks registry references under that lock before deleting, and OAuth refresh rereads the credential after acquiring it. Regression coverage includes concurrent template adds, remove and re-add interleaving, direct helper writes, and refresh versus replacement.

src/registry/materialize.ts
camjac251: Fixed in `4cf938a`. Materialization now recognizes the exact current-main shape: absent `authType`, a `custom-openai` or `custom-anthropic` template, `keyring:provider:<same-id>`, and a stored credential equal to `local`. That runtime provider is normalized to an empty key, `none:anonymous`, and `authType: none`.

An ambiguous `local` value with a mismatched reference fails closed, a missing credential remains unavailable, and a real credential is preserved. The tests cover all four cases.

src/registry/provider-auth.ts
camjac251: Fixed in `4cf938a`. The preflight error now preserves the credential backend diagnostic and directs the user to set `CLODEX_CREDENTIAL_HELPER` to an absolute helper path. `docs/credential-helpers.md` documents the fail-closed behavior and retry path.

`CHANGELOG.md` is managed by release-please and has no pending section, so the conventional `fix(auth)` commit supplies the release-note input instead of editing previously released history.

src/registry/lock.ts
bman654: Reopening this — the lease machinery is a real improvement and the basic eviction case now throws `RegistryLockLostError` correctly, but there's a residual window. The final `assertRegistryWriteOwnership` and the `renameSync` in `saveRegistry` are two separate operations, so an eviction can still land between them. I've left the detailed repro and fix options as a new comment on `src/registry/io.ts:160`. Verified by experiment, not just reading — details there.

src/registry/lock.ts
bman654: Reopening — same root cause as the `io.ts:160` residual. The token+device+inode comparison before unlink is good, but the compare-to-unlink is still non-atomic: a reaper can unlink a replacement lock created after its final `stat`. On its own that's now "loud" (the wrongly-unlinked owner fails its pre-rename ownership check) — *except* when it composes with the `io.ts:160-161` rename gap, where the displaced owner can still publish silently. Fixing the `saveRegistry` publish atomically (see that comment) closes both; this thread can resolve once that does.

src/registry/io.ts
```diff
@@ -143,9 +154,18 @@ export function saveRegistry(registry: ProviderRegistry, path = getProvidersPath
       // backup is best-effort
     }
   }
-  const tmp = `${path}.tmp`;
-  writeSecureFile(tmp, payload);
-  renameSync(tmp, path);
+  const tmp = `${path}.${process.pid}.${randomUUID()}.tmp`;
+  try {
+    writeSecureFile(tmp, payload);
+    assertRegistryWriteOwnership(path);
```
bman654, line 160: **Residual lost-update — the final ownership check and the rename aren't atomic.**

```ts
writeSecureFile(tmp, payload);
assertRegistryWriteOwnership(path);   // ← validates the lease
renameSync(tmp, path);               // ← ...but eviction can land in between
```

These are two syscalls. Reproduced with a deterministic delay injected between them: the owner passed `assertRegistryWriteOwnership`, a successor then evicted it (age reap) + acquired + published, and the original's `renameSync` executed anyway and clobbered the successor. Final on-disk value was the stale owner's. So the window is now microseconds instead of ten minutes — big improvement — but it's not closed, and the failure is silent.

**Why this is genuinely hard:** no single syscall does "rename A→B only if file C still holds my token." `rename()` is atomic for the *content* swap, and `link()`/`O_EXCL` give you atomic create-if-absent — but neither expresses a cross-file conditional. So there are two clean ways out:

**Option 1 — never evict a live owner; make the wedge case loud instead.** Reap only on dead PID (ESRCH), and for a live-but-stuck holder, bound the wait and fail with an actionable message (`another clodex process (pid N) holds the registry lock`). This is exactly the "pid + ESRCH liveness, loser skips with a notice" pattern CLAUDE.md documents for `patch.lock` / `server-runtime.lock`. If a live owner can't be evicted, check-then-rename needs no atomicity — nobody can invalidate you mid-write. Simplest, and it matches the rest of the repo. The cost is giving up automatic recovery from a genuinely-hung process (now a manual kill), which for a developer CLI is a fine trade.

**Option 2 — compare-and-swap on the registry content.** Keep age-based eviction, but make correctness independent of lock timing: embed a monotonic version (or content hash) in `providers.json`. A writer reads version N, and the publish is conditional on the on-disk version still being N at rename time; if a successor bumped it to N+1, the evicted writer's CAS fails and it retries from fresh state. The lock becomes an optimization; the version check is the actual guarantee. More moving parts, but it preserves auto-recovery.

I'd lean Option 1 for clodex — it's less code, kills the whole class of race, and is consistent with the existing locks. Your call on whether automatic hung-process recovery is worth Option 2's complexity.

src/registry/crud.ts
```diff
@@ -21,37 +26,63 @@ export async function removeProviderFromRegistry(
...
+  const authRefToDelete = removal.authRefToDelete;
+  if (authRefToDelete) {
+    await withCredentialMutationLock(authRefToDelete, async () => {
+      const referencedAgain = await withRegistryWriteLock(() =>
+        credentialStillReferenced(
+          authRefToDelete,
+          loadRegistry().providers,
+        ));
+      if (referencedAgain) return;
+      removal.result.credentialDeleted = await deleteProviderCredential(
```
bman654, line 76: **Provider removal still orphans the credential when the backend delete fails.**

The registry deletion is committed (`saveRegistry` inside the write lock, above) before this `deleteProviderCredential` runs, and its `false` return is stored but not surfaced — the function still returns `removed: true`, and `providers-command.ts:316-320` prints "Removed" and exits 0 regardless.

Reproduced with a fake helper whose `delete` exits 9: command printed **"Removed OpenAI"**, exit 0, `providers.json` emptied, and `sk-remains` left in the helper store — with the registry reference already gone, there's no path back to it.

Full disclosure: this one was originally on *our* side of the ledger — it's pre-existing on `main` and I'd flagged it as our follow-up, not yours. You took a run at it as part of this consolidation, which is why I'm raising it here rather than leaving it silent. Two ways to close it:

- delete the credential (under the credential lock) **before** committing the registry removal, so a failure aborts with the entry still intact and retryable; or
- keep the current order but surface the `false` — report the orphaned reference and how to retry, rather than reporting success.

Either is fine; the thing to avoid is the silent "Removed" while the secret persists. Genuinely optional for this PR if you'd rather I take it as a separate fix — just say so.

tests/registry-lock.test.ts
```diff
@@ -0,0 +1,480 @@
...
+      () => {
+        expect(() =>
+          saveRegistry(emptyRegistry(), otherRegistryPath),
+        ).toThrow(RegistryLockLostError);
+      },
+      { lockPath },
+    );
+  });
+
+  it('rejects an expired owner before it can overwrite a newer registry', async () => {
```
bman654: **Test-coverage gap on the parts that matter most.** Not blocking, but worth closing given how subtle this code is.

I mutation-tested the new fixes — reverted each in the source and checked whether a test went red:

- ✅ zero-length/incomplete-lock recovery — **caught** (`:380`)
- ✅ credential I/O moved outside the registry lock — **caught** (`registry-crud.test.ts:97`)
- ✅ A6 `'local'` recognition — **caught**, and the fixture uses the *real* shipped shape (absent `authType`, credential `'local'`), not a fabricated one — nice
- ❌ **pre-rename fence** (`io.ts:160`) — reverting it left all 16 lock tests green. `:420` only exercises the *first* lease check, not the final one before rename
- ❌ **hard-link atomic acquisition** — replacing it with `open(..., 'wx')` kept all lock tests green
- ❌ **per-reference credential lock** — making it a pass-through kept all 31 caller tests green (they mock both locks)
- ❌ **false credential-deletion success** — kept 26 tests green

The through-line: every contention test coordinates promises **inside one Vitest process**, and the caller suites mock the locks away — so cross-process serialization, which is the entire point of this work, is never actually exercised. One test that spawns two real `node` subprocesses racing on the same `CLODEX_HOME` (both provider writes, and an evicted-owner-must-not-publish case) would pin the behavior the single-process tests can't, and would have caught the `io.ts:160` residual.

bman654 (changes requested):
Thanks for turning this around so thoroughly — consolidating the three PRs and reworking the locking was a lot of work, and most of it landed. I re-reviewed the new commit (`4cf938a`) with several focused passes plus a live smoke test, and verified the fixes by experiment rather than just reading the replies.

**Independently verified fixed (these threads can stay resolved):**

- **Credential I/O out of the global lock** — confirmed on every write path; a 5s-blocked helper let an unrelated registry write finish in ~50ms. And the two-lock design is **deadlock-free**: every nested site takes credential→registry in the same order, no reverse nesting. That was my main worry about adding a second lock — it's clean.
- **`'local'` custom-endpoint migration** — all four cases correct, and the test uses the real shipped shape.
- **Fail-closed store-probe** — error names `CLODEX_CREDENTIAL_HELPER`, docs updated.
- **Empty/incomplete-lock 30s stall** — fully fixed via the fsync'd hard-link acquisition; malformed locks now recover in ~40ms, and 20 concurrent writers produced 20/20 unique updates with zero leaked lock/temp files.

**Smoke test:** built this branch and ran real `clodex claude -p` requests through both providers — `luna` (openai-oauth bridge) and `sonnet` (Anthropic passthrough). Both returned cleanly, and the registry was intact with no leftover lock/temp artifacts afterward. The basic bridge works end-to-end on this branch.

**What still needs a look — I've reopened two threads and added three comments:**

1. **`io.ts:160` (blocker)** — the fencing lease closed the 10-minute window but not the last one: the final ownership check and the `renameSync` are non-atomic, so an eviction can still slip between them and silently overwrite a successor's write. Reproduced deterministically. The comment lays out two clean fixes (I lean toward "never evict a live owner", which also matches the existing `patch.lock` pattern).
2. **`crud.ts:71` (major)** — removal still reports success and orphans the secret when the backend delete fails. This was originally our pre-existing issue, not yours; you took a run at it, so I'm flagging it rather than leaving it silent. Optional for this PR.
3. **Test gap (non-blocking)** — the fixes that are correct aren't pinned: reverting the pre-rename fence, the hard-link acquisition, and the credential lock all leave the suite green, because everything runs in one process with the locks mocked. Details in the comment.

Also two low-priority notes for whenever: the hard-link acquisition now makes hard-link support mandatory (FAT/exFAT and some network filesystems will reject every acquire — worth a fallback or at least a documented requirement), and a `SIGKILL` between `link()` and the `finally` cleanup can leave a stray `.tmp` sibling that nothing scavenges.

Really solid progress — the remaining items are narrow. Happy to take the `crud.ts` orphan fix off your plate as a separate change if you'd rather keep this PR focused on the locking.

src/registry/lock.ts
camjac251: Updated in `9794240`. A parseable lock owned by a live PID is now retained regardless of age. That removes the path that could invalidate a live writer after its final ownership check and before registry publication. Contention remains bounded, and the timeout identifies the owner PID. The new subprocess regression ages a live holder past the former threshold, verifies that a contender cannot acquire or replace its token, then lets the original holder publish successfully.

src/registry/lock.ts
camjac251: Updated in `9794240`. The reaper now removes only malformed locks or locks owned by a dead PID. It never age-evicts a live owner. This keeps the compare-to-unlink residual from composing with normal live-owner publication. The cross-process test confirms that the aged live holder retains its token, the contender times out with the holder PID, and the holder completes and cleans up normally.

src/registry/io.ts
camjac251: Updated in `9794240` using the live-owner policy suggested here. Registry locks no longer expire while their owner PID is alive, so another process cannot invalidate the writer between the final ownership check and `renameSync`. Dead and malformed locks remain reclaimable. Live contention still has a bounded timeout and now reports the owning PID. The regression exercises the full holder and contender sequence in separate processes.

src/registry/crud.ts
camjac251: Updated in `9794240`. If the credential backend returns `false`, removal now returns an actionable cleanup error containing the exact credential reference. The command logs the error, exits 1, and does not print the success message. Credential I/O remains outside the global registry lock and under its per-reference mutation lock. Registry-level and command-level tests cover the retained credential result and CLI behavior.

tests/registry-lock.test.ts
camjac251: Added the requested coverage in `9794240`: separate holder and contender processes for live-owner retention, separate processes for per-reference credential serialization, a paused acquisition proving the canonical lock path is not visible until the complete record is ready, and deterministic lease replacement after the registry temporary write proving the final publication fence rejects ownership loss. Credential deletion failure is covered at both result and command levels. Focused runs pass 18/18 lock tests and 15/15 removal and command tests. The README also documents the hard-link filesystem requirement and harmless crash-left temporary-file cleanup.

bman654 (approved):
All findings from the three review rounds are now verified fixed. Final verification on `9794240`:

- **Lost-update closed for real:** re-ran the exact repro that broke the previous round — writer paused between the final ownership assertion and `renameSync`, lock aged 24h, contender racing. The contender timed out naming the owner PID, the token/inode never changed, and the paused writer published correctly. No age-based reclamation of a live-PID lock remains anywhere in the tree.
- **Mutation-tested the whole set:** all five critical mutations (pre-rename fence, hard-link acquisition, credential-lock pass-through, swallowed deletion failure, re-introduced live-owner eviction) now fail a named test. The cross-process tests spawn real Node child processes on shared filesystem locks — no mocks in the worker.
- **Behavior:** dead-PID reclaim 35ms, zero-length reclaim 29ms, reads lock-free under a held write lock, 12/12 concurrent writers survived with zero leaked artifacts, ~10ms uncontended lock cost.
- **Live smoke test on this head:** real `clodex claude -p` through both the openai-oauth bridge and Anthropic passthrough — both clean, registry intact afterward.

The live-owner invariant note in `CLAUDE.md` and the hard-link/tmp documentation in the README are exactly the right places for those.

Thanks for the persistence across three substantive revisions — the final state (fencing leases + never-evict-live + genuine cross-process tests) is stronger than what either of us started with. Approving.

PR #10, opened 2026-07-21
fix(routing): prevent configured route bypasses

camjac251:
## Summary

- resolve saved endpoint aliases through the real catalog route identifier
- retain unavailable alias targets in the reserved namespace so they fail closed
- canonicalize configured and requested route identifiers across context suffixes and provider prefixes
- inspect supported compressed request bodies before selective routing and reject unreadable compressed input locally
- preserve compressed first-party payload bytes, encoding, and authentication during pass-through
- reject unavailable configured routes before default or first-party forwarding with one recovery message
- record local route rejection without upstream provider attribution
- merge uppercase and lowercase proxy bypass lists without dropping existing entries

## Routing guarantees

- available short aliases reach their configured catalog route
- unavailable aliases cannot fall through to the default route
- unavailable canonical identifiers cannot pass through with client credentials
- supported canonical variants resolve to the configured route
- messages and token-count endpoints enforce the same reservation boundary
- unrelated proxy bypass entries survive environment merging

## Test plan

- [x] production route-resolution wiring regression
- [x] compressed pass-through byte-equality regression
- [x] focused endpoint and selective-proxy regressions
- [x] complete suite: 79 files and 888 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] secret scan

### DIFF
--- a/docs/background-agents.md
+++ b/docs/background-agents.md
@@ -13,6 +13,7 @@ This page explains how to bridge **every** Claude Code process on your machine �
 - Setting **`CLAUDE_CODE_PROCESS_WRAPPER`** to `clodex-claude` makes Claude Code invoke it as `clodex-claude <claude-binary-path> <args...>` for every process it spawns — agents view sessions and background agents are bridged automatically.
 - For your own terminal sessions, run **`clodex-claude`** instead of `claude` — same auto-discovery, no port or CA path to hardcode anywhere.
 - Set **`CLODEX_REQUIRE_SERVER=1`** in an isolated routed profile when bypassing Clodex must be impossible. `clodex-claude` then exits with an error if no advertised server passes its process and TCP checks. The default remains fail-open for ordinary installations.
+- Proxy-mode wrapper launches remove Anthropic entries from `NO_PROXY` and `no_proxy`, so inherited bypass settings cannot silently skip the selective proxy.
 
 ## Setup steps
 
--- a/docs/background-agents.md
+++ b/docs/background-agents.md
@@ -13,7 +13,7 @@ This page explains how to bridge **every** Claude Code process on your machine â
 - Setting **`CLAUDE_CODE_PROCESS_WRAPPER`** to `clodex-claude` makes Claude Code invoke it as `clodex-claude <claude-binary-path> <args...>` for every process it spawns â€” agents view sessions and background agents are bridged automatically.
 - For your own terminal sessions, run **`clodex-claude`** instead of `claude` â€” same auto-discovery, no port or CA path to hardcode anywhere.
 - Set **`CLODEX_REQUIRE_SERVER=1`** in an isolated routed profile when bypassing Clodex must be impossible. `clodex-claude` then exits with an error if no advertised server passes its process and TCP checks. The default remains fail-open for ordinary installations.
-- Proxy-mode wrapper launches remove Anthropic entries from `NO_PROXY` and `no_proxy`, so inherited bypass settings cannot silently skip the selective proxy.
+- Proxy-mode wrapper launches remove Anthropic entries from the union of `NO_PROXY` and `no_proxy`, while preserving unrelated bypasses. The manual server output prints the same adjusted values for users who export the proxy settings themselves.
 
 ## Setup steps
 
@@ -28,6 +28,8 @@ This page explains how to bridge **every** Claude Code process on your machine â
    ```bash
    clodex server --proxy
    ```
+
+   If the server prints `NO_PROXY` and `no_proxy`, export those adjusted values with the proxy and CA settings. Empty values intentionally clear inherited wildcard or Anthropic bypasses.
 
    Bridging only happens while this server is running. When it is not, `clodex-claude` falls back cleanly and launches `claude` with an untouched environment.
 
--- a/src/http-proxy/index.ts
+++ b/src/http-proxy/index.ts
@@ -106,6 +106,7 @@ export async function startConfiguredHttpProxy(
     port,
     routes: loaded.routes,
     modelAliases: loaded.aliases,
+    reservedModelIds: loaded.unavailableAliases.map(alias => alias.name),
     debug,
     debugLogPath,
     inferenceLogPath,
--- a/src/http-proxy/index.ts
+++ b/src/http-proxy/index.ts
@@ -6,7 +6,7 @@ import { fetchProviderCatalog, resolveLocalProviderApiKey } from '../provider-ca
 import { providersForTarget } from '../target-compatibility.js';
 import type { ProxyRoute } from '../proxy.js';
 import { buildHttpProxyRoutes, type HttpProxyRouteResult } from './routes.js';
-import { startHttpProxy, type HttpProxyHandle } from './server.js';
+import { startHttpProxy, type HttpProxyHandle, type HttpProxyOptions } from './server.js';
 import { ensureHttpProxyCaBundle } from './ca.js';
 import { registerServerRuntimeState, unregisterServerRuntimeState } from '../server-runtime.js';
 import {
@@ -14,6 +14,7 @@ import {
   getSessionLogPath,
   writeProxyLifecycleLog,
 } from '../trace-log.js';
+import { removeAnthropicProxyBypass } from '../wrapper-env.js';
 
 export interface LoadedHttpProxyRoutes extends HttpProxyRouteResult {
   favoriteCount: number;
@@ -93,15 +94,15 @@ export function reportSkippedHttpProxyFavorites(loaded: LoadedHttpProxyRoutes): 
   }
 }
 
-export async function startConfiguredHttpProxy(
+export function buildConfiguredHttpProxyOptions(
+  loaded: LoadedHttpProxyRoutes,
   port: number,
   debug = false,
   inferenceLogPath = getInferenceRequestLogPath(),
   debugLogPath?: string,
   webSocketDiagnosticsLogPath?: string,
-): Promise<{ handle: HttpProxyHandle; loaded: LoadedHttpProxyRoutes }> {
-  const loaded = await loadHttpProxyRoutes();
-  const handle = await startHttpProxy({
+): HttpProxyOptions {
+  return {
     host: '127.0.0.1',
     port,
     routes: loaded.routes,
@@ -111,7 +112,42 @@ export async function startConfiguredHttpProxy(
     debugLogPath,
     inferenceLogPath,
     webSocketDiagnosticsLogPath,
-  });
+  };
+}
+
+export function formatHttpProxyEnvironmentLines(
+  handle: Pick<HttpProxyHandle, 'port' | 'caCertPath'>,
+  baseEnv: NodeJS.ProcessEnv = process.env,
+): string[] {
+  const env = { ...baseEnv };
+  const hadNoProxy = env['NO_PROXY'] !== undefined || env['no_proxy'] !== undefined;
+  removeAnthropicProxyBypass(env);
+  return [
+    `  HTTPS_PROXY=http://127.0.0.1:${handle.port}`,
+    `  HTTP_PROXY=http://127.0.0.1:${handle.port}`,
+    `  NODE_EXTRA_CA_CERTS=${handle.caCertPath}`,
+    ...(hadNoProxy
+      ? [`  NO_PROXY=${env['NO_PROXY'] ?? ''}`, `  no_proxy=${env['no_proxy'] ?? ''}`]
+      : []),
+  ];
+}
+
+export async function startConfiguredHttpProxy(
+  port: number,
+  debug = false,
+  inferenceLogPath = getInferenceRequestLogPath(),
+  debugLogPath?: string,
+  webSocketDiagnosticsLogPath?: string,
+): Promise<{ handle: HttpProxyHandle; loaded: LoadedHttpProxyRoutes }> {
+  const loaded = await loadHttpProxyRoutes();
+  const handle = await startHttpProxy(buildConfiguredHttpProxyOptions(
+    loaded,
+    port,
+    debug,
+    inferenceLogPath,
+    debugLogPath,
+    webSocketDiagnosticsLogPath,
+  ));
   handle.caCertPath = ensureHttpProxyCaBundle(
     handle.caCertPath,
     process.env['NODE_EXTRA_CA_CERTS'],
@@ -165,9 +201,7 @@ export async function runHttpProxyServerCommand(
   });
   console.log('');
   console.log(pc.bold(pc.green('clodex proxy-mode server running')));
-  console.log(`  HTTPS_PROXY=http://127.0.0.1:${handle.port}`);
-  console.log(`  HTTP_PROXY=http://127.0.0.1:${handle.port}`);
-  console.log(`  NODE_EXTRA_CA_CERTS=${handle.caCertPath}`);
+  for (const line of formatHttpProxyEnvironmentLines(handle)) console.log(line);
   console.log(`  Request log: ${handle.inferenceLogPath}`);
   if (handle.webSocketDiagnosticsLogPath) {
     console.log(`  WebSocket diagnostics: ${handle.webSocketDiagnosticsLogPath}`);
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -143,6 +143,8 @@ export interface HttpProxyOptions {
   routes: ProxyRoute[];
   /** Short incoming model names mapped to canonical adapter route ids. */
   modelAliases?: ResolvedHttpProxyAlias[];
+  /** Configured local model ids that must never fall through to Anthropic. */
+  reservedModelIds?: string[];
   debug?: boolean;
   /** Per-process translated-adapter debug log used when debug is enabled. */
   debugLogPath?: string;
@@ -661,13 +663,18 @@ function forwardPlainHttp(req: http.IncomingMessage, res: http.ServerResponse): 
 export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpProxyHandle> {
   const certificates = ensureHttpProxyCertificates();
   const routesById = new Map<string, ProxyRoute>();
+  const reservedModelIds = new Set<string>();
   for (const route of options.routes) {
     for (const id of routeLookupIds(route.aliasId)) routesById.set(id, route);
   }
   for (const alias of options.modelAliases ?? []) {
+    for (const id of routeLookupIds(alias.name)) reservedModelIds.add(id);
     const route = routesById.get(alias.routeId);
     if (!route) continue;
     for (const id of routeLookupIds(alias.name)) routesById.set(id, route);
+  }
+  for (const modelId of options.reservedModelIds ?? []) {
+    for (const id of routeLookupIds(modelId)) reservedModelIds.add(id);
   }
   const anthropicOrigin = new URL(options.anthropicOrigin ?? 'https://api.anthropic.com');
   let adapter: ProxyHandle | null = options.adapterHandle ?? null;
@@ -714,18 +721,25 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       const claudeSessionId = parsed
         ? extractClaudeSessionId(parsed, claudeSessionIdHeader)
         : undefined;
+      const requestedModel = typeof parsed?.model === 'string' ? parsed.model : undefined;
+      const unresolvedRoutedModel = !route && requestedModel !== undefined && (
+        requestedModel.startsWith('clodex:') || reservedModelIds.has(requestedModel)
+      );
+      const requestedProvider = unresolvedRoutedModel
+        ? (requestedModel?.startsWith('clodex:') ? requestedModel.split(':')[1] || 'unknown' : 'unknown')
+        : 'anthropic';
 
       if (messagesEndpoint === 'messages' && options.inferenceLogPath) {
         const provider = route
           ? (route.providerId ?? route.aliasId.split(':')[1] ?? 'unknown')
-          : 'anthropic';
+          : requestedProvider;
         writeInferenceRequestLog(options.inferenceLogPath, {
           requestId,
           claudeSessionId,
           modelId: typeof parsed?.model === 'string' ? parsed.model : 'unknown',
           effort: parsed ? anthropicEffortFromRequest(parsed) : undefined,
           provider,
-          route: route ? 'translated' : 'passthrough',
+          route: route || unresolvedRoutedModel ? 'translated' : 'passthrough',
           stream: Boolean(parsed?.stream),
           requestPreview: getLatestMessagePreview(parsed?.messages, parsed?.system),
         });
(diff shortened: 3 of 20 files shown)


src/proxy.ts
```diff
@@ -365,7 +368,14 @@ export function startProxyCatalog(
       const relayRequestId = Array.isArray(relayRequestIdRaw) ? relayRequestIdRaw[0] : relayRequestIdRaw;
 
       // Per-request route resolution: look up the alias, fall back to default
-      const route = lookupRoute(byAlias, originalModel) ?? defaultRoute;
+      const resolvedRoute = lookupRoute(byAlias, originalModel);
+      const configuredModelUnavailable = typeof originalModel === 'string'
+        && (originalModel.startsWith('clodex:') || configuredAliasNames.has(originalModel));
```
bman654: **major:** This half of the guard is unreachable from any production call site. `configuredAliasNames` is built solely from the `modelAliases` parameter, and the production endpoint launcher (`src/cli.ts` ~565-575) never passes it — while the MITM caller only passes aliases that are guaranteed resolvable (`buildHttpProxyRoutes` partitions unresolvable ones into `unavailableAliases`), so `!resolvedRoute && configuredAliasNames.has(...)` can't both be true via that path either. Net effect: a patched Claude Code sending a saved alias like `sol` in **endpoint mode** still silently falls through to `defaultRoute` — the exact bypass class this PR targets. (The new test only reaches this branch by passing an unresolvable alias by hand — a call shape no production code produces.)

Fix options: wire the preference aliases (including currently-unavailable ones) into the `startProxyCatalog` call in `cli.ts`, or drop the alias half here and keep only the `clodex:` prefix check until the endpoint path actually supplies aliases.

src/http-proxy/server.ts
```diff
@@ -713,18 +720,25 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       const claudeSessionId = parsed
         ? extractClaudeSessionId(parsed, claudeSessionIdHeader)
         : undefined;
+      const requestedModel = typeof parsed?.model === 'string' ? parsed.model : undefined;
+      const unresolvedRoutedModel = !route && requestedModel !== undefined && (
+        requestedModel.startsWith('clodex:') || reservedModelIds.has(requestedModel)
```
bman654: **major:** The reservation check tests the RAW request value against a set pre-expanded from the configured side, but `routeLookupIds`' expansion isn't closed under the normalization the resolver applies: for alias `sol` the set holds `sol`, `sol[1m]`, `models/sol` — not `sol[1M]` (the `[1m]` suffix is case-insensitive per `src/context-model-id.ts`) or `models/sol[1m]`. Those variants miss both route resolution and this set → still pass through to Anthropic (here) or `defaultRoute` (endpoint proxy, same construction in `src/proxy.ts`). Normalizing the *incoming* model id and checking the bare form would close both sites at once.

Related inverse bug on this line: `routesById.get(...)` is also a raw lookup, so a **valid** canonical id sent as `clodex:openai:gpt-5.6[1M]` misses the route map, then the `clodex:` prefix check rejects it — a false 400 for a supported spelling (the endpoint proxy's `lookupRoute` normalizes and doesn't have this problem).

nit: `'clodex:'` here can be `HTTP_PROXY_MODEL_PREFIX` from `./routes.js` (same package; `src/proxy.ts` can't import it without a cycle, which is worth a comment there).

src/http-proxy/server.ts
```diff
@@ -713,18 +720,25 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       const claudeSessionId = parsed
         ? extractClaudeSessionId(parsed, claudeSessionIdHeader)
         : undefined;
+      const requestedModel = typeof parsed?.model === 'string' ? parsed.model : undefined;
```
bman654, line 747: **major (pre-existing, but it punctures this PR's claim):** if the client sends a compressed body (`Content-Encoding: gzip` etc.), `JSON.parse` on the raw bytes fails, `parsed` is undefined, so `requestedModel` is undefined — the guard can't fire and the body is forwarded to Anthropic with the client's credentials, even if it contains `model: "clodex:…"`. Routing has the same blind spot, so this isn't a regression, but the repo already has `decodeBody` support for gzip/deflate/brotli/zstd in `src/http-utils.ts` — worth either decoding before the model sniff or explicitly documenting compressed bodies as out of scope of the bypass guarantee.

docs/background-agents.md
```diff
@@ -13,6 +13,7 @@ This page explains how to bridge **every** Claude Code process on your machine 
 - Setting **`CLAUDE_CODE_PROCESS_WRAPPER`** to `clodex-claude` makes Claude Code invoke it as `clodex-claude <claude-binary-path> <args...>` for every process it spawns — agents view sessions and background agents are bridged automatically.
 - For your own terminal sessions, run **`clodex-claude`** instead of `claude` — same auto-discovery, no port or CA path to hardcode anywhere.
 - Set **`CLODEX_REQUIRE_SERVER=1`** in an isolated routed profile when bypassing Clodex must be impossible. `clodex-claude` then exits with an error if no advertised server passes its process and TCP checks. The default remains fail-open for ordinary installations.
+- Proxy-mode wrapper launches remove Anthropic entries from `NO_PROXY` and `no_proxy`, so inherited bypass settings cannot silently skip the selective proxy.
```
bman654: **major (docs/UX gap):** this line covers wrapper launches, but the *manual* flow remains bypassable: `clodex server --proxy` prints `HTTPS_PROXY`/`HTTP_PROXY`/`NODE_EXTRA_CA_CERTS` for the user to export (and README's proxy-mode section documents the same), with no mention of scrubbing inherited `NO_PROXY`/`no_proxy`. A user with `NO_PROXY=*` (or `api.anthropic.com`) who follows those instructions exactly still bypasses the MITM. Suggest: have the server's printed export block include an adjusted `NO_PROXY` (reusing `removeAnthropicProxyBypass` on the server's view of the env), and add a sentence to the README/this doc for the manual path.

src/http-proxy/server.ts
```diff
@@ -733,17 +747,37 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
...
+          route: route || unresolvedRoutedModel ? 'translated' : 'passthrough',
           headers: req.headers,
           body: parsed ? parsed as unknown as Record<string, unknown> : {},
         });
       }
 
+      if (unresolvedRoutedModel) {
+        const message = `Clodex model route is unavailable: ${requestedModel}`;
+        if (messagesEndpoint === 'messages' && options.inferenceLogPath) {
+          writeInferenceResponseErrorLog(options.inferenceLogPath, {
```
bman654: **minor:** two log-semantics issues worth a deliberate decision rather than a ride-along: (1) this synthesizes an `upstream_error` entry for a request that never left the machine — a local policy rejection isn't an upstream error, and anyone correlating that event with provider health gets a false signal; (2) rejected requests are now labeled `route: 'translated'` with a guessed provider (`groq` parsed from a typo'd id, `'unknown'` otherwise) where they previously logged as `passthrough`/`anthropic` — a log-schema change not mentioned in the commit body. A distinct event name (e.g. `route_unavailable`) or an explicit note in the commit body would settle both.

src/wrapper-env.ts
```diff
@@ -10,6 +10,34 @@ import type { ServerRuntimeState } from './server-runtime.js';
 const PROXY_ENV_VARS = ['HTTPS_PROXY', 'HTTP_PROXY', 'https_proxy', 'http_proxy'] as const;
 export const REQUIRE_SERVER_ENV = 'CLODEX_REQUIRE_SERVER';
 
+export function removeAnthropicProxyBypass(env: NodeJS.ProcessEnv): void {
+  const noProxy = env['NO_PROXY'] ?? env['no_proxy'];
```
bman654: **minor:** the `NO_PROXY ?? no_proxy` selection means that when both variables are set with different values, entries present only in the non-selected one are silently dropped (e.g. `NO_PROXY=localhost` + `no_proxy=corp.internal` → both become `localhost`, and `corp.internal` traffic now routes through the clodex proxy). This is faithful to the extracted per-session logic, but `computeWrapperEnv` newly applies it to every wrapped process, widening the exposure. Merging the union of both variables before filtering would preserve unrelated bypasses.

tests/proxy.test.ts
```diff
@@ -110,6 +112,103 @@ describe('SDK anonymous route handling', () => {
 });
 
 describe('catalog model aliases', () => {
+  it('rejects unresolved configured model ids without using the default route', async () => {
```
bman654, line 259: **tests:** this harness mocks `node:http` wholesale, captures the internal handler, and hand-rolls req/res doubles — bypassing Node's request parsing and the real response path, and coupling the test to the handler's implementation shape. This same file already has the real-listener pattern (`postToProxy` + the catalog-alias tests just below); a real-listener version of this test would assert actual wire behavior, including the response body (currently only `statusCode` is checked — the Anthropic error shape/`invalid_request_error` type is never pinned). Also: `vi.resetModules`/`vi.doMock`/`vi.stubGlobal` run before the `try`, so a failure during setup skips `doUnmock`/`unstubAllGlobals` and leaks into subsequent tests.

Smaller gaps worth covering while here: the guard on `count_tokens` requests (both proxies), the guard with `modelAliases` undefined, and the MITM side with a `missing-route[1m]` suffixed variant (only the endpoint test covers suffixes).

src/http-proxy/index.ts
```diff
@@ -102,6 +102,7 @@ export async function startConfiguredHttpProxy(
     port,
     routes: loaded.routes,
     modelAliases: loaded.aliases,
+    reservedModelIds: loaded.unavailableAliases.map(alias => alias.name),
```
bman654, line 110: **tests:** this mapping — the only production source of `reservedModelIds` — has no test: the server-level test injects `reservedModelIds: ['missing-route']` manually, so deleting this line would break the feature while every test stays green. A test through `startConfiguredHttpProxy` (or a focused unit on the loaded-config → options mapping) that proves an unavailable alias from `buildHttpProxyRoutes` actually lands in the reject set would pin the plumbing.

src/proxy.ts
```diff
@@ -365,7 +368,14 @@ export function startProxyCatalog(
...
+      if (!resolvedRoute && configuredModelUnavailable) {
+        anthropicError(res, 400, `Model '${originalModel}' is unavailable`);
```
bman654: **nit:** two different wordings for the same condition (`Model '<id>' is unavailable` here vs `Clodex model route is unavailable: <id>` in the MITM), and neither gives the user a path forward. For the wrapper/background-agent case there's no startup warning visible, so this 400 is all they see for a removed provider, deleted key, dropped favorite, or stale patch alike. Suggest one shared message with a hint, e.g. "…run `clodex models --list` to see available routes, or `clodex patch` to refresh baked-in aliases."

bman654 (changes requested):
Thanks — the direction here is exactly right, and the review panel (routing correctness, proxy-env, tests, architecture) unanimously endorsed fail-closed: the old behavior shipped the full conversation plus the user's Anthropic OAuth token to api.anthropic.com under a bogus model id, and a local 400 is strictly better. Also verified on this branch: the full suite passes in my environment (69 files / 693 tests, listener suites included) plus `pnpm typecheck`; the NO_PROXY extraction is behaviorally identical to the removed inline block; the wrapper stays dependency-free (confirmed via a scratch build — no heavy imports leak into the wrapper bundle); real Anthropic model traffic is untouched, preserving the selective-MITM contract; and the branch merges cleanly with both current `main` and #21 — including the semantic check that #21's retained anonymous routes can never land in this PR's `reservedModelIds`.

**Why request-changes:** the implementation currently misses its own goal in four places, one of them structural:

1. **(major)** The endpoint-proxy alias guard is unreachable from production — `clodex claude` endpoint launches never pass `modelAliases` into `startProxyCatalog`, so the alias-name bypass the title targets still exists in endpoint mode. Inline comment on `src/proxy.ts`.
2. **(major)** The guard sets aren't closed under the routing normalization rules — variants like `sol[1M]` or `models/sol[1m]` still fall through. Inline comment on `src/http-proxy/server.ts`.
3. **(major, pre-existing)** Compressed request bodies bypass the guard (and routing) entirely. Inline comment.
4. **(major, docs/UX)** The documented manual `clodex server --proxy` export flow still inherits `NO_PROXY` bypasses. Inline comment on `docs/background-agents.md`.

Plus test-shape asks (the mocked `node:http` harness, and the untested production `unavailableAliases → reservedModelIds` mapping) and two minor log/env issues — all inline. None of this needs a redesign; the shape of the change is right.

src/proxy.ts
camjac251: Addressed in `063ea47`. The production endpoint launcher now passes every saved preference alias into `startProxyCatalog`, including aliases whose targets are absent. Request lookup and the configured-alias reservation set use the same canonical key, so an unavailable saved alias is rejected before default-route fallback.

src/http-proxy/server.ts
camjac251: Addressed in `063ea47`. `normalizeRouteLookupId()` strips the context suffix case-insensitively and normalizes the `models/` prefix. Both proxy route maps, alias targets, reservation sets, and incoming request IDs use that key. Tests cover uppercase suffixes, prefixed suffix variants, and a valid canonical ID that must resolve rather than return 400.

src/http-proxy/server.ts
camjac251: Addressed in `063ea47`. The selective proxy now reuses `decodeRequestBody()` before model inspection. Supported compressed routed requests are decoded for the local adapter hop, compressed unavailable routes are rejected locally, and unreadable compressed bodies return a local 400 without forwarding client credentials.

docs/background-agents.md
camjac251: Addressed in `063ea47`. Manual proxy output now includes adjusted uppercase and lowercase bypass values when either was inherited, including explicit empty values when wildcard-only input must be cleared. The main guide and background-process guide document exporting those values with the proxy and CA settings.

src/http-proxy/server.ts
camjac251: Addressed in `063ea47`. Local policy rejection now records a distinct `route_unavailable` event before ordinary request logging, with no provider or translated/passthrough attribution. Tests verify it does not emit a false upstream error or a normal routed request entry.

src/wrapper-env.ts
camjac251: Addressed in `063ea47`. Uppercase and lowercase bypass lists are merged, trimmed, filtered, and deduplicated before the same result is written back to both spellings. Unrelated entries present in only one spelling are retained.

tests/proxy.test.ts
camjac251: Addressed in `063ea47` and `eace413`. The mocked HTTP harness was replaced with real-listener coverage that pins the exact 400 error shape and verifies no upstream call. It covers messages, token counting, aliases omitted, suffix case, and prefixed variants. Describe-level cleanup now restores global stubs even when listener startup fails, while each opened listener is closed through `finally`.

src/http-proxy/index.ts
camjac251: Addressed in `063ea47`. `buildConfiguredHttpProxyOptions()` is now the production loaded-config-to-server-options boundary. Its focused test proves that every unavailable alias is copied into `reservedModelIds`, so removing that mapping breaks the test.

src/proxy.ts
camjac251: Addressed in `063ea47`. Both proxy paths now use `routeUnavailableMessage()` and return the same error wording. The response includes concrete list and refresh commands so a background or wrapper request has a recovery path without relying on startup output.

src/proxy.ts
```diff
@@ -365,7 +370,19 @@ export function startProxyCatalog(
       const relayRequestId = Array.isArray(relayRequestIdRaw) ? relayRequestIdRaw[0] : relayRequestIdRaw;
 
       // Per-request route resolution: look up the alias, fall back to default
-      const route = lookupRoute(byAlias, originalModel) ?? defaultRoute;
+      const resolvedRoute = typeof originalModel === 'string'
+        ? lookupRoute(byAlias, originalModel)
+        : undefined;
+      const configuredModelUnavailable = typeof originalModel === 'string'
+        && (
+          originalModel.startsWith('clodex:')
```
bman654: The incoming id is normalized for the route/alias lookup above, but this fail-closed guard tests the **raw** `originalModel`. `models/clodex:openai:missing[1M]` normalizes to `clodex:openai:missing` for the lookup, misses, then fails this raw `startsWith('clodex:')` check — and falls through to the default route. Checking the normalized lookup key here closes the last spelling in the equivalence class. Worth adding real-listener messages + count_tokens cases for a prefixed unavailable canonical id (including uppercase `[1M]`).

src/http-proxy/server.ts
```diff
@@ -701,18 +712,59 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
...
       }
       const claudeSessionIdHeader = Array.isArray(req.headers['x-claude-code-session-id'])
         ? req.headers['x-claude-code-session-id'][0]
         : req.headers['x-claude-code-session-id'];
       const claudeSessionId = parsed
         ? extractClaudeSessionId(parsed, claudeSessionIdHeader)
         : undefined;
+      const requestedModel = typeof parsed?.model === 'string' ? parsed.model : undefined;
+      const unresolvedRoutedModel = !route && requestedModel !== undefined && (
+        requestedModel.startsWith(HTTP_PROXY_MODEL_PREFIX)
```
bman654: Same hole on the MITM side: `requestedModel.startsWith(HTTP_PROXY_MODEL_PREFIX)` tests the raw id after the normalized lookup missed, so `models/clodex:…` spellings of an unavailable canonical route forward the original body — with the client's Anthropic authorization — upstream. Apply the prefix check to the normalized key here too.

bman654 (changes requested):
Re-reviewed at eace413 against our previous review. Substantial and careful revision — all four majors were engaged head-on, and the test rewrite from the mocked `node:http` harness to real listeners is a clear upgrade (exact 400 shapes pinned, no-upstream-call proven, suffix/prefix variants covered, stub cleanup fixed in eace413). Verified locally: typecheck + full suite green (69 files / 699 tests).

Verified addressed: endpoint alias reservation is now live production code (`cli.ts` passes all saved aliases into `startProxyCatalog`, absent targets stay reserved); `normalizeRouteLookupId()` closes the alias cases from our review including valid `clodex:…[1M]` resolution; compressed bodies are decoded before model inspection with routed/reserved/unreadable all handled locally and passthrough still forwarding the original bytes; `NO_PROXY` guidance covers both spellings including wildcard-only; `route_unavailable` trace event; `buildConfiguredHttpProxyOptions()` is now a tested production boundary; unified error wording.

One hole remains in the normalization closure — the fail-closed prefix check still inspects the raw id (inline comments). It's the same fail-open class this PR exists to remove, so requesting one more round; the fix looks small.

Two non-blocking coverage notes:
- Byte-for-byte passthrough is pinned only for uncompressed first-party bodies; a compressed first-party byte-equality case would directly guard the invariant that decoding never touches passthrough.
- The endpoint behavior tests construct `startProxyCatalog(..., modelAliases)` directly, so the `cli.ts` plumbing that feeds it could later regress without a test failing. A thin test through the production wiring would close that.

Rebase notes: #23 is landing ahead of this, which leaves a trivial import conflict in `src/http-proxy/server.ts` (keep both its `listenTcpServer` import and your routing imports). Merge-tree also shows a content conflict with #16 in `src/proxy.ts` — whichever revision resolves it, this PR's route-unavailable rejection must stay ahead of #16's token-count/refresh handling.

src/proxy.ts
camjac251: Addressed in `b7e7a7b`. The endpoint guard now checks `normalizeRouteLookupId(originalModel)`, so the compatibility prefix and context suffix are removed before testing the reserved namespace. Real-listener regressions cover messages and token counting with `models/clodex:...[1M]` spelling and verify that no default-route request is made.

src/http-proxy/server.ts
camjac251: Addressed in `b7e7a7b`. The selective-proxy guard now applies the reserved-prefix check to `normalizeRouteLookupId(requestedModel)`. Regressions cover messages and token counting with the prefixed uppercase-suffix spelling and verify that neither an upstream request nor client authorization reaches the origin.

bman654 (changes requested):
Thanks — the remaining major from last round is genuinely fixed: both guards now normalize before the reserved-namespace check (`proxy.ts` + `http-proxy/server.ts`), `models/clodex:...[1M]` is rejected fail-closed, and the real-listener regressions are non-tautological. 

One new **medium** to resolve before merge, in endpoint switch-menu mode:

`cli.ts:1231` feeds saved aliases into `startProxyCatalog` as `modelAliasTarget(alias)` → `clodex:<providerId>:<modelId>`, but the catalog's routes are keyed by `aliasModelId` → `anthropic-<slug>__<modelId>`. So `lookupRoute(byAlias, alias.routeId)` never matches: the alias name lands in `configuredAliasNames` (reserved) but is never added to `byAlias` (routed). Net effect — in endpoint mode, a request by an alias's short name (e.g. a patched binary's `model: sol`) hits the new fail-closed guard and returns **400 even when the alias's target model is in the catalog**. That contradicts `abc12ce` ("feed all saved aliases into endpoint routing"): the reserve half works, the route half can't.

Fix: pass the resolved catalog `aliasId` (e.g. `resolveRoute(alias.providerId, alias.modelId)?.aliasId`) as `routeId`, not `modelAliasTarget(alias)`.

Two open test-coverage items from last round are exactly what hide this, worth closing alongside: (1) the endpoint alias tests construct routes whose `aliasId` is already `clodex:...`, so alias `routeId`s match in-test but never in the real `cli.ts` path; (2) compressed first-party passthrough byte-equality is still untested.

bman654 (approved):
Approving at f1071bb.

The endpoint switch-menu medium from last round is fixed correctly. `resolveCatalogModelAliases()` now maps each saved alias through the same `resolveRoute` that builds the catalog, so the alias's `routeId` is the catalog route's `aliasId` (`anthropic-<slug>__<model>[1m]`) — the exact key `byAlias` is keyed on — instead of the `clodex:<provider>:<model>` canonical id that never matched. `cli.ts` passes that resolver in, so the namespace lines up end to end. Targets absent from the catalog fall back to `modelAliasTarget()` and stay reserved-but-unrouted, i.e. still fail closed. Verified the prior guards are untouched (this commit only touches `catalog.ts` + `cli.ts` + tests).

The two test-coverage gaps we flagged are both closed:
- The alias integration test in `proxy.test.ts` no longer hand-builds `clodex:`-keyed routes — it builds them through the real `makeRouteResolver`/`resolveCatalogModelAliases` and asserts the routeId is `anthropic-test-provider__solver-v1[1m]`, so it exercises the real cli.ts namespace.
- New `resolveCatalogModelAliases` unit tests assert the resolved id and explicitly that it is *not* the canonical id, plus the reserved-fallback case.
- New compressed first-party passthrough test proves byte-equality (`receivedBody.equals(compressedBody)`) plus encoding + auth preserved.

Verified locally on a branch merging this over current `main`: `pnpm typecheck` clean, full suite green (79 files / 889 tests).

Land-order note (no change needed here): this merges cleanly with `main`. If it lands alongside #26, there's one mechanical conflict in `src/http-proxy/server.ts` at the translated-route `forwardToAdapter` call — this PR switches the body arg to `adapterBody`, #26 adds `claudeSessionId` to the lifecycle object and a trailing `() => shuttingDown`. Whoever lands second combines all three; I verified the combined form typechecks and the suite stays green.

PR #11, opened 2026-07-21
fix(transport): terminate rejected connection upgrades

camjac251:
## Summary

- Convert rejected connection upgrades into status-aware terminal stream errors.
- Drain rejected responses and discard failed connections so clients do not hang or reuse them.
- Preserve structured upstream errors through forced streaming.

## Test plan

- [x] Focused transport and stream tests: 89 passed.
- [x] Isolated non-listener suite: 571 passed.
- [x] `pnpm typecheck`
- [x] `pnpm build`


bman654:
thanks for these fixes.

PR #13, opened 2026-07-21
fix(auth): represent anonymous access explicitly

camjac251:
## Summary

- Persist anonymous access as an explicit credential reference.
- Keep anonymous providers from consulting stale environment or stored credentials.
- Preserve free-only filtering and display anonymous access clearly.

## Test plan

- [x] Focused credential and registry tests: 64 passed.
- [x] Isolated non-listener suite: 575 passed.
- [x] `pnpm typecheck`
- [x] `pnpm build`

### DIFF
--- a/src/registry/add-template.ts
+++ b/src/registry/add-template.ts
@@ -94,7 +94,9 @@ export async function addProviderFromTemplate(
     };
   }
 
-  const authRef = `keyring:provider:${template.id}`;
+  const authRef = trimmedKey
+    ? `keyring:provider:${template.id}`
+    : 'none:anonymous';
   const saved = trimmedKey ? await saveProviderCredential(authRef, trimmedKey) : true;
   if (!saved) {
     return {
@@ -118,7 +120,10 @@ export async function addProviderFromTemplate(
     name: template.name,
     enabled: true,
     authRef,
-    authType: template.authType,
+    authType: trimmedKey ? template.authType : 'none',
+    ...(!trimmedKey && template.anonymousFreeModels
+      ? { subscriptionFilter: 'free' as const }
+      : {}),
     api: {
       npm: template.npm,
       url: fetched.baseUrl,
--- a/src/registry/materialize.ts
+++ b/src/registry/materialize.ts
@@ -8,7 +8,6 @@ import { normalizeGoogleDisplayName, normalizeGoogleModelId } from './google-mod
 import { findModelsDevModel } from './models-dev.js';
 import type { CachedModel, ProviderRegistry, RegistryProvider } from './types.js';
 import { isValidProviderId } from './validate.js';
-import { getTemplateById } from '../provider-templates.js';
 import { classifyFreeStatus, isFreeStatus } from '../free-models.js';
 
 export type CredentialResolver = (provider: RegistryProvider) => string | null;
@@ -83,11 +82,6 @@ export function cachedModelToLocal(
   };
 }
 
-function providerAllowsAnonymousFreeModels(provider: RegistryProvider): boolean {
-  const template = getTemplateById(provider.templateId) ?? getTemplateById(provider.id);
-  return template?.anonymousFreeModels === true;
-}
-
 function materializeOne(
   provider: RegistryProvider,
   resolveCredential: CredentialResolver,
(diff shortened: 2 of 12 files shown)


src/registry/add-template.ts
```diff
@@ -94,7 +94,9 @@ export async function addProviderFromTemplate(
     };
   }
 
-  const authRef = `keyring:provider:${template.id}`;
+  const authRef = trimmedKey
+    ? `keyring:provider:${template.id}`
+    : 'none:anonymous';
```
bman654, line 99: **The explicit representation is wired into a path that can't currently fire, while the path that *can* is left a no-op — including what looks like a half-finished edit.**

Neither shipped template permits anonymous access (`src/provider-templates.ts:31-43` — `openai` is `authType: 'api'`, `openai-oauth` is `'oauth'`), so `trimmedKey` is never empty here and this `'none:anonymous'` branch is currently unreachable. That's fork residue: the optional-key template flags still referenced in `provider-catalog.ts:37` (`apiKeyOptional`, `anonymousFreeModels`) are vestigial from relay-ai, which shipped providers this one doesn't.

The only reachable way to create a no-auth provider is a custom endpoint with a blank key — and that path (`src/registry/custom-endpoint.ts`, not touched by this PR) already gets the metadata right but never applies the new ref:

```ts
// :164 — correct
authType: apiKey === 'local' ? 'none' : 'api',

// :193 — both branches are identical
authRef: apiKey === 'local' ? `keyring:provider:${providerId}` : `keyring:provider:${providerId}`,
```

That ternary is exactly where `'none:anonymous'` belongs. As written it's a no-op branch, which reads like the edit was started and not finished.

Consequently the behavior this PR is named for is still broken for the only users who can hit it:

- `:152` still coerces a blank key to the literal string `'local'`, and `:208-209` **stores that placeholder in the OS keychain as if it were a real credential**
- because `resolveProviderCredential` checks `CLODEX_KEY_<PROVIDER>` *before* the keyring, a stale namespaced env credential still silently overrides an explicit no-auth choice — and gets transmitted to what is typically a local Ollama/LM Studio endpoint
- the keychain accumulates junk entries that aren't credentials

**Suggested fix:** set `authRef: 'none:anonymous'` in that `:193` ternary, skip the `:208` `saveProviderCredential` call entirely when `apiKey === 'local'`, and drop the `'local'` sentinel in favor of the `authType`/`authRef` pair you've introduced here.

**Test gap:** the new tests pass only because they construct fabricated optional templates absent from the shipped catalog, so the reachable path stays uncovered. Worth a behavioral test that sets `CLODEX_KEY_<PROVIDER>`, creates an anonymous custom endpoint, and asserts the env credential is ignored.

src/registry/materialize.ts
```diff
@@ -97,29 +91,30 @@ function materializeOne(
   if (!isValidProviderId(provider.id)) return null;
 
   const freeOnly = provider.subscriptionFilter === 'free';
-  const apiKey = resolveCredential(provider) ?? '';
-  const anonymousFreeOnly = !apiKey.trim() && providerAllowsAnonymousFreeModels(provider);
+  const anonymous = provider.authRef === 'none:anonymous' && provider.authType === 'none';
```
bman654, line 94: **Minor — no migration for entries written in the old representation.**

Materialization now requires the exact `none:anonymous` + `authType: 'none'` pair, but nothing migrates providers persisted under the previous scheme (a `keyring:provider:<id>` ref with no stored credential). Such an entry won't match `anonymous`, will resolve to an empty `apiKey`, and gets silently dropped by the `!apiKey.trim() && !anonymous` guard below — the provider just disappears from the catalog after upgrade, with no warning.

I confirmed this by hand-writing an old-style no-key registry entry: it parsed fine but failed to materialize, and the server exited 1 with "No providers configured."

No released clodex template could have produced such an entry, so supported installs aren't affected today — but it does break the generalized behavior your own new tests exercise. Either add a migration in the registry load path or accept the legacy shape as anonymous.

bman654 (changes requested):
#8 #14 and this PR are all changing the same files and it's going to be a string of merge conflicts.  I'll happily accept a single PR with all the changes combined along with the fixes requested in each PR.

bman654 (changes requested):
please address the comments.  consider combining with #14 and #8 PRs for less merge conflict

src/registry/add-template.ts
camjac251: Handled in `a37ade2` on #8, the combined branch requested for these overlapping changes. Blank custom-endpoint keys now produce `authRef: 'none:anonymous'` with `authType: 'none'`; no credential or `local` placeholder is written; namespaced environment credentials are ignored; and anonymous model discovery omits `x-api-key`. Tests cover the stored registry shape, absence of a credential write, environment isolation, and the request header.

src/registry/materialize.ts
camjac251: Handled in `a37ade2` on #8. The runtime now accepts only the unambiguous prior no-auth shape, `authType: 'none'` with `keyring:provider:<same-id>`, skips credential and environment resolution for it, and normalizes the materialized reference to `none:anonymous`. Missing API-auth credentials remain rejected. Both cases are covered by regression tests.

PR #14, opened 2026-07-21
fix(registry): serialize provider state changes

camjac251:
## Summary

- Serialize registry write transactions across processes with stale-owner recovery.
- Keep network discovery outside the lock, then reload and revalidate state before committing.
- Reject refresh results when credentials or discovery configuration changed in flight.

## Test plan

- [x] Focused registry and lock tests: 58 passed.
- [x] Isolated non-listener suite: 587 passed.
- [x] `pnpm typecheck`
- [x] `pnpm build`

### DIFF
--- a/src/registry/lock.ts
+++ b/src/registry/lock.ts
@@ -1,0 +1,335 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
+import { randomUUID } from 'node:crypto';
+import {
+  closeSync,
+  mkdirSync,
+  openSync,
+  readFileSync,
+  statSync,
+  unlinkSync,
+  writeFileSync,
+} from 'node:fs';
+import { dirname } from 'node:path';
+import { getProvidersPath } from '../paths.js';
+
+const DEFAULT_WAIT_MS = 30_000;
+const DEFAULT_RETRY_MS = 25;
+const INVALID_LOCK_STALE_MS = 10 * 60 * 1000;
+
+interface RegistryLockOwner {
+  pid: number;
+  startedAt: number;
+  token: string;
+}
+
+interface RegistryLockSnapshot {
+  raw: string;
+  device: number;
+  inode: number;
+  modifiedAt: number;
+}
+
+interface RegistryLockOptions {
+  lockPath?: string;
+  waitMs?: number;
+  retryMs?: number;
+  now?: () => number;
+  isAlive?: (pid: number) => boolean;
+}
+
+interface RegistryLockContext {
+  leases: ReadonlyMap<string, RegistryLockLease>;
+}
+
+interface RegistryLockLease {
+  active: boolean;
+}
+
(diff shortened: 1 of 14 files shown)


src/registry/lock.ts
```diff
@@ -0,0 +1,335 @@
...
+  const stats = statSync(lockPath);
+  const snapshot: RegistryLockSnapshot = {
+    raw,
+    device: stats.dev,
+    inode: stats.ino,
+    modifiedAt: stats.mtimeMs,
+  };
+  const owner = parseLockOwner(raw);
+  if (owner) return alive(owner.pid) ? null : snapshot;
+
```
bman654, line 95: **A live-but-wedged owner holds this lock forever — there's no age-based expiry.**

`startedAt` is parsed, validated (`parseLockOwner`), and written into every lock record, but it's never actually consulted for staleness. A well-formed lock is reaped *only* when its PID is dead. The `INVALID_LOCK_STALE_MS` grace period below applies solely to **malformed** locks.

This diverges from the established pattern elsewhere in the repo, which uses liveness **OR** age (`src/patcher.ts:239`):

```ts
!alive(existing.pid)
  || (typeof existing.startedAt === 'number' && now - existing.startedAt > PATCH_LOCK_STALE_MS)
```

CLAUDE.md documents that same "pid + staleness + ESRCH-liveness" shape for both `patch.lock` and `server-runtime.lock`, so this is the one place that can wedge.

**Failure scenario:** a clodex process hangs (stuck network call, SIGSTOP, debugger) or crashes in a way that lets its PID be reused. It owns the registry lock indefinitely. Every subsequent registry mutation — add provider, auth, refresh models, pricing — waits the full timeout and fails. I measured it: a lock held by a live PID caused a **30.45s wait then exit 1**, and it stays that way permanently rather than self-healing.

**Suggested fix:** treat a valid lock as stale when `now - owner.startedAt` exceeds a bounded max-hold, in addition to the dead-PID check:

```ts
if (owner) {
  const expired = now - owner.startedAt >= LOCK_MAX_HOLD_MS;
  return alive(owner.pid) && !expired ? null : snapshot;
}
```

The existing `raw`/`dev`/`ino`/`mtime` snapshot comparison in `removeStaleLock` already makes the reap safe against racing with a legitimate owner.

bman654 (changes requested):
#8 #13 and this PR are all changing the same files and it's going to be a string of merge conflicts.  I'll happily accept a single PR with all the changes combined along with the fixes requested in each PR.

src/registry/lock.ts
camjac251: Handled in `a37ade2` on #8. Valid lock records now expire after a 10-minute maximum hold even if the PID is still alive, while fresh live owners remain protected. Reclamation still rechecks the full raw, device, inode, and modification-time snapshot under the reaper guard before unlinking. Tests cover fresh live owners, dead owners, expired live owners, and replacement races.

PR #15, opened 2026-07-21
fix(auth): make credential cleanup crash-safe

camjac251:
## Summary

- persist cleanup intent before writing a stored credential
- fail closed when the provider registry or cleanup journal cannot be read and validated
- publish complete registry and journal payloads with file and parent-directory durability
- constrain cleanup entries to generated provider references and bound queued work
- reconcile from a command-scoped finalizer so cancellation and early errors still retry pending cleanup
- emit one consistent cleanup warning without repeating successful-path reconciliation
- reject malformed present registry fields while retaining unknown forward-compatible fields
- retain the exported custom-endpoint lifecycle under the same credential safety contract

## Recovery guarantees

- a credential saved before registry commit remains journaled and is removed during reconciliation
- a credential referenced by a committed provider is retained even if marker cancellation was interrupted
- malformed or unreadable registry state never authorizes credential deletion
- uncertain deletion and journal-persistence outcomes remain queued for an idempotent retry
- add, remove, and authorization commands retry queued cleanup even when they exit before mutation
- unrelated registry mutations never persist a lossy projection of malformed known fields

## Test plan

- [x] focused lifecycle, registry, and provider-command regressions
- [x] complete suite: 75 files and 798 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] secret and personal-identifier scans

## Related work

- #17 owns chunked credential publication, recovery, and deletion, including interruption before the base chunk record exists.
- #16 owns rejected access-token refresh and one-time request recovery.

### DIFF
--- a/docs/credential-helpers.md
+++ b/docs/credential-helpers.md
@@ -32,6 +32,15 @@ diagnostic in the error instead of continuing with tokens that cannot be
 durably stored. Set `CLODEX_CREDENTIAL_HELPER` to the absolute path of an
 external helper, then run the authorization command again.
 
+Provider creation, replacement, and removal use a durable cleanup journal in
+`providers.json`. A new unreferenced credential is journaled before it is
+written, provider changes are saved before superseded credentials are deleted,
+and uncertain deletion outcomes remain queued. Per-credential cross-process
+locks serialize writes, activation, removal, and reconciliation for the same
+reference without blocking unrelated credentials. The next `clodex providers`
+command retries queued deletions idempotently and never deletes a credential
+that is referenced by an active provider.
+
 ## Protocol
 
 The helper receives one of these invocations:
--- a/docs/credential-helpers.md
+++ b/docs/credential-helpers.md
@@ -32,14 +32,18 @@ diagnostic in the error instead of continuing with tokens that cannot be
 durably stored. Set `CLODEX_CREDENTIAL_HELPER` to the absolute path of an
 external helper, then run the authorization command again.
 
-Provider creation, replacement, and removal use a durable cleanup journal in
-`providers.json`. A new unreferenced credential is journaled before it is
-written, provider changes are saved before superseded credentials are deleted,
-and uncertain deletion outcomes remain queued. Per-credential cross-process
-locks serialize writes, activation, removal, and reconciliation for the same
-reference without blocking unrelated credentials. The next `clodex providers`
-command retries queued deletions idempotently and never deletes a credential
-that is referenced by an active provider.
+Provider creation, replacement, and removal use the durable cleanup journal at
+`~/.clodex/credential-cleanup.json` (under `CLODEX_HOME` when set). Keeping the
+journal separate from `providers.json` prevents registry writers that only know
+schema 1 provider fields from dropping pending cleanup. A new unreferenced
+credential is journaled before it is written, provider changes are saved before
+superseded credentials are deleted, and uncertain deletion outcomes remain
+queued. Per-credential cross-process locks serialize writes, activation,
+removal, and reconciliation for the same reference without blocking unrelated
+credentials. Reconciliation is best-effort: one busy credential does not block
+unrelated cleanup or turn an already-saved provider into a failed creation. The
+next `clodex providers` command retries queued deletions idempotently and never
+deletes a credential that is referenced by an active provider.
 
 ## Protocol
 
--- a/docs/credential-helpers.md
+++ b/docs/credential-helpers.md
@@ -39,11 +39,19 @@ schema 1 provider fields from dropping pending cleanup. A new unreferenced
 credential is journaled before it is written, provider changes are saved before
 superseded credentials are deleted, and uncertain deletion outcomes remain
 queued. Per-credential cross-process locks serialize writes, activation,
-removal, and reconciliation for the same reference without blocking unrelated
-credentials. Reconciliation is best-effort: one busy credential does not block
-unrelated cleanup or turn an already-saved provider into a failed creation. The
-next `clodex providers` command retries queued deletions idempotently and never
-deletes a credential that is referenced by an active provider.
+removal, and reconciliation for the same reference. Reconciliation is
+best-effort and sequential: a contended credential can delay later entries
+until its lock attempt times out, but the timeout does not turn an already-saved
+provider into a failed creation. The next `clodex providers` command retries
+queued deletions idempotently and never deletes a credential that is referenced
+by an active provider. If the registry cannot be read and validated, cleanup
+stays queued instead of treating the registry as empty.
+
+The cleanup journal accepts only credential accounts generated for Clodex
+provider and OAuth records, including replacement, custom-provider, and scoped
+credential instances. It rejects symbolic links, foreign ownership, broad
+permissions on POSIX, files over 1 MiB, and more than 1,024 queued entries
+before attempting any credential-store deletion.
 
 ## Protocol
 
--- a/src/providers-command.ts
+++ b/src/providers-command.ts
@@ -263,6 +263,9 @@ async function runTemplateAddFlow(): Promise<number> {
   }
 
   logConnected(template.name, result.modelCount ?? 0);
+  if (result.credentialCleanupPending) {
+    p.log.warn('A previous credential is queued for cleanup and will be retried by the next provider command.');
+  }
   return 0;
 }
 
@@ -481,8 +484,14 @@ export async function runProvidersCommand(args: string[]): Promise<number> {
     return 0;
   }
 
-  const cleanup = await reconcilePendingCredentialDeletes();
-  if (cleanup.pending.length > 0 || cleanup.persistenceError) {
+  let cleanupPending = false;
+  try {
+    const cleanup = await reconcilePendingCredentialDeletes();
+    cleanupPending = cleanup.pending.length > 0 || cleanup.persistenceError !== undefined;
+  } catch {
+    cleanupPending = true;
+  }
+  if (cleanupPending) {
     p.log.warn('Some credential cleanup is still pending and will be retried later.');
   }
 
--- a/src/providers-command.ts
+++ b/src/providers-command.ts
@@ -18,6 +18,7 @@ import {
   removeProviderFromRegistry,
   toggleProviderEnabled,
 } from './registry/crud.js';
+import { reconcilePendingCredentialDeletes } from './registry/credential-lifecycle.js';
 import { loadRegistry } from './registry/io.js';
 import { refreshAllProviderModels, refreshProviderModels } from './registry/refresh-models.js';
 import { resolveRefreshCredential } from './registry/refresh-credentials.js';
@@ -115,6 +116,9 @@ export async function runProvidersAuth(providerId: string, method?: ProviderAuth
   try {
     const result = await authenticateProvider(providerId, { method });
     p.log.success(`Signed in to ${result.registryProvider.name} — credential saved to the credential store.`);
+    if (result.credentialCleanupPending) {
+      p.log.warn('A previous credential is queued for cleanup and will be retried by the next provider command.');
+    }
     return 0;
   } catch (err) {
     if (err instanceof Error && err.message === 'Cancelled') {
@@ -321,6 +325,9 @@ export async function runProvidersRemove(id: string, interactive = false): Promi
   if (result.credentialDeleted) {
     p.log.info('Provider API key removed from the credential store.');
   }
+  if (result.credentialCleanupPending) {
+    p.log.warn('Provider credential cleanup is queued for retry.');
+  }
   return 0;
 }
 
@@ -472,6 +479,11 @@ export async function runProvidersCommand(args: string[]): Promise<number> {
   if (parsed.showHelp && parsed.subcommand !== 'auth') {
     console.log(providersHelpText());
     return 0;
+  }
+
+  const cleanup = await reconcilePendingCredentialDeletes();
+  if (cleanup.pending.length > 0 || cleanup.persistenceError) {
+    p.log.warn('Some credential cleanup is still pending and will be retried later.');
   }
 
   if (parsed.subcommand === 'list') return runProvidersList();
--- a/src/providers-command.ts
+++ b/src/providers-command.ts
@@ -43,6 +43,52 @@ export type ProvidersSubcommand = 'hub' | 'add' | 'list' | 'remove' | 'refresh-m
 const CREDENTIAL_CLEANUP_PENDING_MESSAGE =
   'Credential cleanup is pending and will be retried by the next provider command.';
 
+interface ProviderCommandCleanupState {
+  reconciled: boolean;
+  pending: boolean;
+}
+
+function reportCredentialCleanup(
+  pending: boolean,
+  state?: ProviderCommandCleanupState,
+  reconciled = false,
+): void {
+  if (state) {
+    state.reconciled ||= reconciled;
+    state.pending ||= pending;
+    return;
+  }
+  if (pending) {
+    p.log.warn(CREDENTIAL_CLEANUP_PENDING_MESSAGE);
+  }
+}
+
+async function reconcileCredentialCleanup(): Promise<boolean> {
+  try {
+    const cleanup = await reconcilePendingCredentialDeletes();
+    return cleanup.pending.length > 0 || cleanup.persistenceError !== undefined;
+  } catch {
+    return true;
+  }
+}
+
+async function runWithCredentialCleanup(
+  run: (state: ProviderCommandCleanupState) => Promise<number>,
+): Promise<number> {
+  const state: ProviderCommandCleanupState = {
+    reconciled: false,
+    pending: false,
+  };
+  try {
+    return await run(state);
+  } finally {
+    if (!state.reconciled) {
+      state.pending = await reconcileCredentialCleanup();
+    }
+    reportCredentialCleanup(state.pending);
+  }
+}
+
 export function parseProvidersArgs(args: string[]): {
   subcommand: ProvidersSubcommand;
   showHelp: boolean;
@@ -115,13 +161,15 @@ function providerLabel(name: string, modelCount: number, enabled: boolean): stri
   return `${fmtEnabledStar(enabled)} ${fmtProvider(name)} ${pc.dim(`(${modelCount} model${modelCount === 1 ? '' : 's'})`)}`;
 }
 
-export async function runProvidersAuth(providerId: string, method?: ProviderAuthMethod): Promise<number> {
+async function runProvidersAuthWithCleanupState(
+  providerId: string,
+  method?: ProviderAuthMethod,
+  cleanupState?: ProviderCommandCleanupState,
+): Promise<number> {
   try {
     const result = await authenticateProvider(providerId, { method });
     p.log.success(`Signed in to ${result.registryProvider.name} — credential saved to the credential store.`);
-    if (result.credentialCleanupPending) {
-      p.log.warn(CREDENTIAL_CLEANUP_PENDING_MESSAGE);
-    }
+    reportCredentialCleanup(result.credentialCleanupPending, cleanupState, true);
     return 0;
   } catch (err) {
     if (err instanceof Error && err.message === 'Cancelled') {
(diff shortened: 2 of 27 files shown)


src/registry/credential-lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+        persistenceError: errorMessage(error),
+      };
+    }
+  }
+
+  try {
+    return await withCredentialMutationLock(authRef, async () => {
+      try {
+        const clearedReferencedMarker = await withRegistryWriteLock(async () => {
+          if (!credentialIsReferenced(loadRegistry(), authRef)) return false;
```
bman654: **major:** This check fails open. `loadRegistry()` (io.ts) catches every read/parse error and returns an *empty* registry, so a transient `EIO`/`EACCES` or a corrupt `providers.json` makes this line report "unreferenced" for a credential that a committed provider still uses — reconcile then deletes it and clears the marker. Crash scenario: registry commit succeeds → crash before marker cancel → next command hits a transient registry read failure → live credential deleted.

"Empty registry" and "registry unreadable" can't share one result on the deletion path. Suggest a `loadRegistryStrict()` (or an out-param/thrown error) used here, with reconcile treating unreadable-registry as `persistenceError` (stay queued) rather than unreferenced.

src/registry/io.ts
```diff
@@ -40,9 +41,10 @@ export function ensureSecureAppHome(): void {
 function writeSecureFile(path: string, content: string): void {
   ensureSecureAppHome();
   mkdirSync(dirname(path), { recursive: true, mode: DIR_MODE });
-  const fd = openSync(path, 'w', FILE_MODE);
+  const fd = openSync(path, 'wx', FILE_MODE);
   try {
     writeSync(fd, content);
```
bman654: **major:** `writeSync`'s return value (bytes written) is unchecked. A short write — interrupted write, resource exhaustion — produces a truncated payload that the subsequent `fsyncSync` and `renameSync` then durably publish as `providers.json`. The commit caller cancels the new credential's journal marker on "success," and the next `loadRegistry()` parses the truncated file into an empty registry: the new credential is now unreferenced *and* unjournaled, and queued superseded deletes may proceed against a registry that looks empty (see the companion comment in credential-lifecycle.ts).

Loop until the full byte length is written (`writeSync` returns the count; standard retry loop) before fsync.

src/registry/credential-cleanup-journal.ts
```diff
@@ -0,0 +1,168 @@
...
+  const data = raw as Record<string, unknown>;
+  if (data.schemaVersion !== JOURNAL_SCHEMA_VERSION) {
+    throw new Error('Unsupported credential cleanup journal schema.');
+  }
+  if (!Array.isArray(data.pendingCredentialDeletes)) {
+    throw new Error('Credential cleanup journal is missing its pending list.');
+  }
+  const pending = data.pendingCredentialDeletes.filter(
+    (value): value is string =>
+      typeof value === 'string' && isStoredCredentialRef(value),
```
bman654: **medium (hardening):** this filter accepts any string with a `keyring:`/`helper:` prefix, so a poisoned journal entry (`keyring:<arbitrary-account>`) gets its account deleted from the store on the next provider command — the referenced-check only protects exact matches of live providers' refs. Granted, an attacker who can write `CLODEX_HOME` can also edit `providers.json`, so this isn't a new trust boundary — but constraining entries to the shapes clodex actually generates (`provider:<id>`, `provider:<id>:<uuid>`, `provider:<id>:replacement:<uuid>`, oauth refs) is cheap and shrinks the blast radius of any bug that writes garbage into the journal.

Two smaller items in the same area: the journal has no size/entry-count bound (a huge poisoned file means unbounded sequential store operations per command), and `ensureSecureAppHome` swallows chmod failures with no ownership/symlink validation — worth a bound and an lstat check while you're here.

src/registry/credential-lifecycle.ts
```diff
@@ -0,0 +1,167 @@
...
+        return {
+          deleted: false,
+          cleared: false,
+          persistenceError: errorMessage(error),
+        };
+      }
+
+      let deleted = false;
+      try {
+        deleted = await deleteProviderCredential(authRef);
```
bman654, line 96: **note (known gap — fine to defer to #17):** this delete can't clean a crash that happens mid-chunked-keyring-write. Chunk entries (`::chunk::N`) are written before the base entry that records the chunk count, and deletion discovers chunks only through the base entry — so a crash before the base entry is written leaves secret-bearing chunk entries that this call reports as successfully deleted (base missing ⇒ "success"), clearing the marker while the chunks persist forever. Same mechanism leaves stale tail chunks when a same-authRef re-auth shrinks the chunk count.

Since #17 explicitly targets chunked-credential crash-safety and depends on this PR, no change requested here — just flagging so the scenario is captured and #17 can test exactly this window.

tests/credential-lifecycle.test.ts
```diff
@@ -0,0 +1,255 @@
...
+}));
+
+import { deleteProviderCredential } from '../src/env.js';
+import * as cleanupJournal from '../src/registry/credential-cleanup-journal.js';
+import { reconcilePendingCredentialDeletes } from '../src/registry/credential-lifecycle.js';
+
+const TEST_HELPER_ID = 'a'.repeat(64);
+const helperRef = (account: string): string => `helper:v1:${TEST_HELPER_ID}:${account}`;
+
+describe('credential cleanup lifecycle', () => {
```
bman654, line 85: **tests (major ask):** the crash-safety suites simulate interruption points but assert against mocked registry/journal/lock state — no test performs a real operation against a temp `CLODEX_HOME` + fake credential helper, abandons it *between* steps (credential saved, registry not committed; committed, marker not canceled), and then invokes reconciliation from the **persisted** state to verify the credential is actually deleted or retained. That end-to-end persisted-state path is the property this PR exists to guarantee, and the one real-filesystem retry test seeds the journal by hand rather than producing the interrupted state through the operation under test.

Two more gaps worth covering: a repeat-reconcile test for the already-deleted-credential/stuck-marker case, and a same-authRef OAuth re-auth (the existing migration test switches keyring→helper, which sidesteps the interesting same-ref path).

tests/credential-cleanup-journal.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import {
+  getCredentialCleanupPath,
+  getProvidersPath,
+  resetLegacyMigrationForTests,
+} from '../src/paths.js';
+
+const TEST_HELPER_ID = 'a'.repeat(64);
+const helperRef = (account: string): string => `helper:v1:${TEST_HELPER_ID}:${account}`;
+
+describe('credential cleanup journal', () => {
```
bman654, line 30: **tests:** the registry durability suite verifies temp-fsync → rename → parent-dir-fsync ordering (including failure ordering) — but its spies only match registry temp files, so the journal's equivalent write path has no fsync/ordering/failure test. Also missing: corrupt-journal parses (malformed JSON, wrong `schemaVersion`, non-object, non-array pending list — the current parser test only feeds structurally valid JSON with invalid entries), and any test of the new `'wx'` exclusive-create behavior.

src/registry/custom-endpoint.ts
```diff
@@ -216,20 +222,52 @@ export async function addCustomEndpointProvider(input: AddCustomEndpointInput):
...
-  if (anonymous) return commitProvider();
-  return withCredentialMutationLock(authRef, async () => {
-    const saved = await saveProviderCredential(authRef, apiKey);
-    if (!saved) {
-      return {
-        added: false,
-        error: 'Could not save API key to the credential store.',
-        hint: 'Check credential-store access and try again.',
-      };
+  const result: AddCustomEndpointResult = anonymous
```
bman654, line 242: **minor:** `addCustomEndpointProvider` has no caller in `src/` (only `fetchAnthropicModels` is imported from this module), so this +64-line rework extends a vestigial path and its `credentialCleanupPending` is never surfaced to a user. CONTRIBUTING's "verify the code path is actually reachable" applies — either note the intentional consistency choice in the PR description, or leave this file for whichever PR revives the path.

docs/credential-helpers.md
```diff
@@ -31,6 +31,19 @@ diagnostic in the error instead of continuing with tokens that cannot be
...
 
+Provider creation, replacement, and removal use the durable cleanup journal at
+`~/.clodex/credential-cleanup.json` (under `CLODEX_HOME` when set). Keeping the
+journal separate from `providers.json` prevents registry writers that only know
+schema 1 provider fields from dropping pending cleanup. A new unreferenced
+credential is journaled before it is written, provider changes are saved before
+superseded credentials are deleted, and uncertain deletion outcomes remain
+queued. Per-credential cross-process locks serialize writes, activation,
+removal, and reconciliation for the same reference without blocking unrelated
+credentials. Reconciliation is best-effort: one busy credential does not block
```
bman654: **minor (docs):** "one busy credential does not block unrelated cleanup" overstates it — reconcile is a sequential loop and `withCredentialMutationLock` waits up to 30s per contended ref before the error is swallowed, so one busy credential delays every ref behind it by up to 30s. Best-effort, yes; non-blocking, no. (After #16 lands — it adds `waitMs` to `withCredentialMutationLock` — passing a short wait here would make this sentence true and keep the pre-dispatch reconcile from stalling interactive commands.)

src/providers-command.ts
```diff
@@ -474,6 +484,17 @@ export async function runProvidersCommand(args: string[]): Promise<number> {
...
 
+  let cleanupPending = false;
+  try {
+    const cleanup = await reconcilePendingCredentialDeletes();
+    cleanupPending = cleanup.pending.length > 0 || cleanup.persistenceError !== undefined;
+  } catch {
+    cleanupPending = true;
+  }
+  if (cleanupPending) {
+    p.log.warn('Some credential cleanup is still pending and will be retried later.');
```
bman654: **nit:** this is the third wording for the same condition ("…queued for cleanup and will be retried by the next provider command", "Provider credential cleanup is queued for retry", and this one) — unify. Also note add/remove/auth reconcile again internally after this pre-dispatch reconcile, so one command can print two differently-worded warnings for one stuck ref.

bman654 (changes requested):
Thanks — this is careful work, and the review panel (durability/concurrency, security, tests, architecture) verified a lot of it holds up: the separate-journal rationale is factually correct (`parseRegistry` whitelist-copies fields, so piggybacked cleanup state would be dropped by any writer not extended for it), the full lock-ordering graph is clean (credential → registry → journal, reentrant via AsyncLocalStorage, no inversion found at any call site), the wrong-credential deletion guards survived the crud.ts rework, no secret values ever land in the journal, the `'wx'` change is safe (single caller, always a unique tmp path), and the suite is green in my environment (72 files / 729 tests + typecheck). The journal-before-write / save-before-delete / cancel-after-commit protocol is sound at the operation boundaries we traced.

**Why request-changes** — two failure modes undercut the crash-safety guarantee itself:

1. **(major)** `loadRegistry()` fails open: any read/parse error yields an empty registry, so reconciliation treats "registry unreadable" as "credential unreferenced" and deletes a live credential. Inline comment on `credential-lifecycle.ts`.
2. **(major)** `writeSync`'s byte count is unchecked, so a short write can be fsynced and renamed into a truncated `providers.json` — which then feeds finding 1. Inline comment on `io.ts`.

Plus one security hardening ask (journal entries can name arbitrary store accounts), and test asks — the crash-safety tests validate mocked state rather than reconciling from genuinely persisted interrupted state.

**Known/deferred:** a crash mid-chunked-keyring-write leaves secret chunk entries the journal can't see (chunks are written before the base entry that indexes them). Since #17 ("make chunked credentials crash-safe") explicitly targets this and depends on this PR, treating it as out of scope here is fine — inline note with the details so it's captured.

**Stack sequencing (for the maintainer and you):** this PR **conflicts textually and semantically with #21** — both rewrite `provider-auth.ts` (this PR's `persistOAuthProvider` subsumes #21's warn→hard-fail change) and `crud.ts` (this journal's `isStoredCredentialRef` supersedes #21's guard). One of them needs to rebase after the other lands; as written, #21's provider-auth/crud hunks become redundant if this merges first. Composition with #16 was verified clean. Given five parallel PRs in one subsystem, an explicit declared merge order would help everyone.

Finally, commit-body honesty: the io.ts durability change (fsync + parent-dir sync + `'wx'` on every registry save), reconcile-on-every-`providers`-command, and the auth warn→fail behavior change are all user-visible and unmentioned — in a release-please repo the commit body is the changelog, so please enrich it.

src/registry/credential-lifecycle.ts
camjac251: Addressed in `85c019b`. Reconciliation now uses `loadRegistryStrict()`, which propagates read, JSON, schema, and provider-shape errors. Cleanup leaves the marker queued and reports a persistence error without calling credential deletion. Unit and persisted-state tests cover an unreadable or malformed registry.

src/registry/io.ts
camjac251: Addressed in `85c019b`. Secure registry writes now loop until the complete buffer is written, and zero progress fails before fsync or rename. The durability suite covers repeated short writes and verifies that a zero-progress write is never published.

src/registry/credential-cleanup-journal.ts
camjac251: Addressed in `85c019b`. Journal entries are restricted to generated provider, replacement, custom-provider, OAuth, and scoped credential shapes. Reads now enforce file identity, symlink, ownership, permission, 1 MiB, 1,024-entry, and per-reference bounds. Invalid entries fail the journal read instead of being filtered into cleanup work.

src/registry/credential-lifecycle.ts
camjac251: Kept in #17 scope as requested. The PR description now records interruption before the base chunk record exists as an explicit #17 responsibility. This branch does not claim chunk-level orphan cleanup coverage.

tests/credential-lifecycle.test.ts
camjac251: Addressed in `85c019b`. New persisted-state tests use a temporary application home and the real fake credential helper to cover interruption before registry commit, interruption after registry commit but before marker cancellation, and malformed registry recovery. The suite also covers repeated reconciliation after marker-clear failure and same-reference OAuth reauthorization.

tests/credential-cleanup-journal.test.ts
camjac251: Addressed in `85c019b`. A journal-specific durability suite verifies exclusive `wx` creation, temp-file fsync before rename, parent-directory fsync after rename, and both fsync failure paths. Parser coverage now includes malformed JSON, wrong schema, non-object roots, missing or non-array pending lists, invalid references, limits, symlinks, and broad permissions.

src/registry/custom-endpoint.ts
camjac251: The custom-endpoint lifecycle is intentionally retained so the exported path follows the same credential safety contract as the active provider flows. The PR description and `85c019b` commit body now disclose that it has no current CLI caller; no reachability claim is being made.

docs/credential-helpers.md
camjac251: Addressed in `85c019b`. The documentation now describes reconciliation as sequential and notes that a contended credential can delay later entries until the lock attempt times out. It no longer claims unrelated cleanup is non-blocking.

src/providers-command.ts
camjac251: Addressed in `85c019b`. All three paths use one cleanup-warning message, and add, remove, and active auth mutations skip the pre-dispatch reconcile because they reconcile internally. A command-level remove regression verifies that pending cleanup emits exactly one warning.

src/providers-command.ts
```diff
@@ -474,6 +487,27 @@ export async function runProvidersCommand(args: string[]): Promise<number> {
     return 0;
   }
 
+  const reconcilesDuringMutation =
```
bman654, line 562: Suppressing pre-dispatch reconciliation for `add`/`remove`/active-`auth` assumes each mutation reconciles internally — but that only holds when the flow reaches its common result handling. Early returns skip it entirely: `addProviderFromTemplate()` returns before its reconcile block on package-probe/empty-key/existing-provider/model-discovery failures; `runProvidersRemove()` returns on the provider-not-found precheck; a cancelled or failed OAuth authorization returns through the catch before `persistOAuthProvider()`'s post-commit reconcile; cancelling the add picker never enters a mutation flow. On all of those, queued cleanup is neither retried nor warned about — contrary to the "next provider command retries" contract in the docs and warning text. A single reconcile attempt in a common `finally`/result wrapper (with centralized warning emission, so no double-reconcile on the success path) would cover the gaps.

src/registry/io.ts
```diff
@@ -117,6 +140,29 @@ function parseRegistry(raw: unknown): ProviderRegistry {
   return registry;
 }
 
+function parseRegistryStrict(raw: unknown): ProviderRegistry {
```
bman654, line 178: `parseRegistryStrict()` rejects a provider only when `parseProvider()` returns null — but `parseProvider()` silently *normalizes* malformed known fields on the way through: invalid `subscriptionFilter`/`authType` values are omitted, non-string `refreshedAt` dropped, malformed `modelsCache` metadata/entries filtered. A registry with those defects passes strict validation, and then any unrelated mutation (e.g. `toggleProviderEnabled()` strict-loads and saves the whole registry) persists the lossy projection. That's the remaining instance of the "reject partial data before mutations" class this commit targets: malformed *present* known fields should fail strict validation rather than be sanitized. (Unknown forward-compat fields can keep the existing policy.)

bman654 (changes requested):
Re-reviewed at 1df7f4c against our previous review. The crash-safety core is now solid: both original majors are verified fixed, and the persisted-state test suite is the real thing — temp app home, executable fake helper, real journal/lock/save paths, actual interrupted states on disk. Verified locally: typecheck + full suite green (74 files / 755 tests).

Verified addressed:
- Reconcile now uses `loadRegistryStrict()` under the decision lock; read/JSON/schema failures leave the marker queued and never reach credential deletion (unit + persisted-state coverage). ENOENT is correctly distinguished from corruption, so first-run still degrades to an empty registry.
- `writeSecureFile()` loops to completion, fails on zero progress before fsync/rename; durability tests drive actual repeated short writes.
- Journal reads are bounded and identity-checked (regular-file/symlink/ownership/permissions/1 MiB/1,024 entries/per-ref bytes), entry shapes are restricted, and any invalid content fails the whole read with cleanup left queued.
- The chunk-crash window is cleanly deferred to #17 with the scoped credential-instance account shape already admitted here.
- Docs no longer claim non-blocking reconcile; 1df7f4c extending strict reads to every mutation/save path is a welcome hardening beyond what we asked.

Two mediums remain from the revision itself (inline comments): the pre-dispatch reconcile suppression opens early-return gaps where queued cleanup is neither retried nor warned about, and `parseRegistryStrict()` still silently normalizes malformed *known* fields — the residual form of the data-loss class 1df7f4c targets.

Heads-up for #17: its current draft is based on pre-1df7f4c main and rewrites add-template/CRUD/custom-endpoint/OAuth paths from the old fail-open base. When rebasing it over this PR, accepting those versions wholesale would silently regress the strict-mutation boundary added here — worth an explicit pass during that rebase.

Rebase note: #21 is landing ahead of this, so a rebase is needed (five conflicting files); our read is this PR's lifecycle implementation should win those hunks — its `persistOAuthProvider` hard-fail and journal `isStoredCredentialRef` supersede #21's equivalents. Composes cleanly with #16 (one semantic note lives in the #16 review).

src/providers-command.ts
camjac251: Addressed in `19d0001`. Add, remove, and active authorization now run inside a command-scoped cleanup finalizer. Mutation helpers report when they already reconciled, so successful paths do not repeat the work, while picker cancellation, pre-mutation failures, missing providers, and authorization cancellation or failure still retry queued cleanup and emit at most one warning.

src/registry/io.ts
camjac251: Addressed in `19d0001`. Strict registry loading now rejects malformed present `subscriptionFilter`, `authType`, `refreshedAt`, and `modelsCache` fields before any mutation can persist a normalized projection. Regressions verify the original file is unchanged on failure, while unknown provider and cached-model fields remain accepted for forward compatibility.

bman654 (approved):
Approving. Both outstanding mediums verified fixed against the head:

- The cleanup finalizer now covers all early-return paths — cancelled OAuth, provider-not-found, and failed-add all reconcile + warn, while successful paths don't double-reconcile.
- `parseRegistryStrict` now rejects malformed known fields (`authType`, `subscriptionFilter`, `refreshedAt`, `modelsCache`) before any write, so an unrelated mutation can no longer persist a lossy projection.

Sweep of the new commits surfaced only one non-blocking minor (hub-mode mutations bypass `runWithCredentialCleanup`), tracked on our side. The journal durability + persisted-state crash tests are solid.

PR #16, opened 2026-07-21
fix(auth): recover once from rejected access tokens

camjac251:
## Summary

- cache only access-token metadata and revalidate the shared store within 30 seconds
- resolve a replacement after an upstream authentication rejection and retry at most once before output starts
- share the replacement decision across native and translated dispatch without restructuring route-specific error handling
- preserve valid opaque JSON for non-refresh references while rejecting unknown JSON on refresh references
- strip credential references, account identifiers, and provider data from public model listings
- enforce the 30-second refresh-request timeout
- let serialized credential mutations wait through the 150-second refresh-holder budget

## Recovery guarantees

- an unchanged replacement token is never retried
- a second rejection surfaces without another refresh attempt
- a response that has started is never replayed
- concurrent refresh callers join in-flight work but re-resolve if the joined token is the caller's rejected token
- repeated compare-and-set generation changes stop at the bounded exhaustion error
- a valid refresh holder cannot make a waiting persistence operation time out at 30 seconds

## Test plan

- [x] focused authentication, route, lock, and transport regressions
- [x] complete suite: 71 files and 776 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] secret and personal-identifier scans

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -152,7 +152,18 @@ function oauthProviderIdFromAccount(account: string): string | null {
 }
 
 const oauthRefreshInflight = new Map<string, Promise<string | null>>();
-const oauthCredentialCache = new Map<string, StoredOAuthCredential>();
+interface CachedOAuthCredential {
+  access: string;
+  expires: number;
+  accessRejected?: true;
+  checkedAt: number;
+}
+
+// Another process can replace the shared credential without invalidating this
+// process. Keep only access-token metadata and bound that stale view to 30 seconds;
+// rejection and expiration bypass it immediately.
+const OAUTH_CREDENTIAL_CACHE_MAX_AGE_MS = 30_000;
+const oauthCredentialCache = new Map<string, CachedOAuthCredential>();
 const rejectedEnvCredentialFingerprints = new Map<string, string>();
 const OAUTH_REFRESH_LOCK_WAIT_MS = 150_000;
 const OAUTH_STATE_KEY_SEPARATOR = '\0';
@@ -453,6 +464,32 @@ function clearOAuthCredentialCache(authRef: string): void {
   }
 }
 
+function cacheOAuthCredential(
+  stateKey: string,
+  credential: StoredOAuthCredential,
+): void {
+  oauthCredentialCache.set(stateKey, {
+    access: credential.access,
+    expires: credential.expires,
+    ...(credential.accessRejected === true ? { accessRejected: true as const } : {}),
+    checkedAt: Date.now(),
+  });
+}
+
+function cachedOAuthCredentialIsUsable(
+  credential: CachedOAuthCredential | undefined,
+  providerId: string,
+  rejectedAccessToken?: string,
+): boolean {
+  if (!credential) return false;
+  const age = Date.now() - credential.checkedAt;
+  return age >= 0
+    && age < OAUTH_CREDENTIAL_CACHE_MAX_AGE_MS
+    && credential.access !== rejectedAccessToken
+    && credential.accessRejected !== true
+    && !oauthCredentialShouldRefresh(credential, providerId);
+}
+
 async function readOAuthProviderSecret(
   ref: StoredCredentialRef,
   providerId: string,
@@ -469,12 +506,7 @@ async function readOAuthProviderSecret(
   }
 
   const cached = oauthCredentialCache.get(stateKey);
-  if (
-    cached
-    && cached.access !== rejectedAccessToken
-    && cached.accessRejected !== true
-    && !oauthCredentialShouldRefresh(cached, providerId)
-  ) {
+  if (cached && cachedOAuthCredentialIsUsable(cached, providerId, rejectedAccessToken)) {
     return cached.access;
   }
   if (cached?.access === rejectedAccessToken) oauthCredentialCache.delete(stateKey);
@@ -483,9 +515,7 @@ async function readOAuthProviderSecret(
     const latestCached = oauthCredentialCache.get(stateKey);
     if (
       latestCached
-      && latestCached.access !== rejectedAccessToken
-      && latestCached.accessRejected !== true
-      && !oauthCredentialShouldRefresh(latestCached, providerId)
+      && cachedOAuthCredentialIsUsable(latestCached, providerId, rejectedAccessToken)
     ) {
       return latestCached.access;
     }
@@ -499,7 +529,7 @@ async function readOAuthProviderSecret(
         const decoded = decodeProviderSecret(raw);
         return decoded === rejectedAccessToken ? null : decoded;
       }
-      oauthCredentialCache.set(stateKey, cred);
+      cacheOAuthCredential(stateKey, cred);
 
       const forceRefresh = cred.access === rejectedAccessToken || cred.accessRejected === true;
       if (!forceRefresh && !oauthCredentialShouldRefresh(cred, providerId)) {
@@ -590,7 +620,7 @@ export async function saveProviderCredential(
       const oauth = parseStoredOAuthCredential(key);
       const oauthProviderId = oauthProviderIdFromAccount(parsed.account);
       if (oauth && oauthProviderId) {
-        oauthCredentialCache.set(oauthCredentialStateKey(oauthProviderId, cacheKey), oauth);
+        cacheOAuthCredential(oauthCredentialStateKey(oauthProviderId, cacheKey), oauth);
       }
       return true;
     }
--- a/src/env.ts
+++ b/src/env.ts
@@ -1,6 +1,6 @@
 // src/env.ts
 import { CONFLICTING_ENV_VARS } from './constants.js';
-import { randomUUID } from 'node:crypto';
+import { createHash, randomUUID } from 'node:crypto';
 import {
   deleteCredentialHelperAccount,
   readCredentialHelperAccount,
@@ -11,6 +11,7 @@ import { resolveContextWindow } from './context-window.js';
 import {
   oauthCredentialToKeychainJson,
   parseStoredOAuthCredential,
+  type StoredOAuthCredential,
 } from './oauth/types.js';
 import { refreshStoredOAuthCredential, oauthCredentialShouldRefresh } from './oauth/refresh.js';
 import { withCredentialMutationLock } from './registry/lock.js';
@@ -151,12 +152,19 @@ function oauthProviderIdFromAccount(account: string): string | null {
 }
 
 const oauthRefreshInflight = new Map<string, Promise<string | null>>();
+const oauthCredentialCache = new Map<string, StoredOAuthCredential>();
+const rejectedEnvCredentialFingerprints = new Map<string, string>();
+const OAUTH_REFRESH_LOCK_WAIT_MS = 150_000;
 
 export type ParsedAuthRef =
   | { kind: 'keyring'; account: string }
   | { kind: 'helper'; helperId: string; account: string }
   | { kind: 'env'; varName: string }
   | { kind: 'none' };
+
+export interface ResolveCredentialOptions {
+  rejectedAccessToken?: string;
+}
 
 /** Parse registry credential references. */
 export function parseAuthRef(authRef: string): ParsedAuthRef | null {
@@ -183,6 +191,37 @@ function readEnvCredential(varName: string): string | null {
   const raw = process.env[varName];
   if (!raw?.trim()) return null;
   return raw.trim().split(/\r?\n/)[0]?.trim() || null;
+}
+
+function credentialFingerprint(value: string): string {
+  return createHash('sha256').update(value).digest('hex');
+}
+
+function usableEnvCredential(
+  source: string,
+  value: string | null,
+  rejectedAccessToken?: string,
+): string | null {
+  if (!value) {
+    rejectedEnvCredentialFingerprints.delete(source);
+    return null;
+  }
+
+  const fingerprint = credentialFingerprint(value);
+  if (
+    rejectedAccessToken !== undefined
+    && fingerprint === credentialFingerprint(rejectedAccessToken)
+  ) {
+    rejectedEnvCredentialFingerprints.set(source, fingerprint);
+    return null;
+  }
+
+  const rejectedFingerprint = rejectedEnvCredentialFingerprints.get(source);
+  if (rejectedFingerprint === fingerprint) return null;
+  if (rejectedFingerprint !== undefined) {
+    rejectedEnvCredentialFingerprints.delete(source);
+  }
+  return value;
 }
 
 function readKeyringAccountFromService(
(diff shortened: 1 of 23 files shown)


src/env.ts
```diff
@@ -368,68 +418,159 @@ export async function resolveProviderOAuthProviderData(
...
-  if (existing) return existing;
+  const stateKey = oauthCredentialStateKey(providerId, authRef);
+  const existing = oauthRefreshInflight.get(stateKey);
+  if (existing) {
+    const resolved = await existing;
+    if (resolved !== rejectedAccessToken) return resolved;
+    return readOAuthProviderSecret(ref, providerId, diag, rejectedAccessToken);
+  }
+
+  const cached = oauthCredentialCache.get(stateKey);
```
bman654, line 508: **major:** this cache-hit shortcut breaks cross-process freshness. Store reads now happen only on cache miss, expiry, or rejection — and `saveProviderCredential`/`deleteProviderCredential` invalidate only *this process's* cache. So after `clodex providers auth` in another terminal replaces the stored credential (e.g. switching ChatGPT accounts), a long-running `clodex server` keeps resolving — and billing — the superseded account's token until it expires or 401s, potentially its full remaining lifetime. `main` re-read the store each call, and the PR summary's "resolve current stored credentials before dispatch" is exactly what this line skips.

Options, cheapest first: a short TTL on cache hits (even 30–60s bounds the staleness window while keeping the keychain off the hot path); a cheap store-fingerprint/mtime check before trusting the hit; or cross-process invalidation. Whichever you pick, please also add a comment here stating the staleness invariant ("healed only by expiry or 401") — it's non-obvious and the next reader will assume the store is authoritative.

src/env.ts
```diff
@@ -151,13 +152,21 @@ function oauthProviderIdFromAccount(account: string): string | null {
 }
 
 const oauthRefreshInflight = new Map<string, Promise<string | null>>();
+const oauthCredentialCache = new Map<string, StoredOAuthCredential>();
```
bman654: **low (security):** this cache retains the full `StoredOAuthCredential` — including the long-lived **refresh** token — in process memory for the server's lifetime, with no TTL or shutdown eviction. Previously the full credential was only transient (access tokens were already long-lived in routes/models, but refresh tokens weren't). A crash dump or process-memory read now recovers the refresh token. Caching only what resolution needs (`access`, `expires`, `accessRejected`) and re-reading the store when a refresh is actually required would keep the hot path fast without parking refresh tokens in memory.

src/oauth/types.ts
```diff
@@ -6,6 +6,7 @@ export interface StoredOAuthCredential {
   refresh: string;
   /** Epoch millis when the access token expires. */
   expires: number;
+  accessRejected?: true;
```
bman654, line 9: **low / doc ask:** persisting this flag to the shared store is the right call for the short-lived-wrapper model (an in-process signal would never survive to the process that next refreshes), and the trigger is conservative — but the marker outlives transient conditions: if inference transiently 401s and the IdP returns the *same* access token on refresh, every process now force-refreshes and refuses that still-valid token until the IdP mints a different one or the user re-auths. Worth (a) one sentence in docs/credential-helpers.md stating the escape hatch ("cleared when a refresh yields a different token; otherwise reauthenticate"), and (b) considering whether `clodex providers auth` should explicitly clear it on success — it does today via the fresh credential write, so mainly: document it.

src/proxy.ts
```diff
@@ -605,18 +619,39 @@ export function startProxyCatalog(
...
+            sdkAttempt === 0 &&
+            !res.headersSent &&
+            route.refreshToken
+          ) {
+            const replacement = await route.refreshToken(apiKey).catch(() => null);
+            if (replacement && replacement !== apiKey) {
+              apiKey = replacement;
+              route.apiKey = replacement;
+              sdkAttempt += 1;
+              plog(() => 'sdk oauth credential replaced after 401; retrying once');
```
bman654: **suggestion (non-blocking):** this 401-replacement gate (`oauth && status===401 && attempt===0 && !headersSent` → resolve with rejected token → `replacement !== apiKey`) is now hand-inlined here and twice in `server/router.ts`, plus `fetchWithOAuthRetry`. Inlining the *loops* is right — the per-site error rendering genuinely differs and restructuring dispatch would violate the prime directive — but the decision predicate is pure and will drift (this copy already differs by also mutating `route.apiKey`; router's relies on `resolveModelApiKey` mutating `model.apiKey`). A tiny shared `resolveRetryReplacement(status, attempt, headersSent, refresh, apiKey)` called inside each site's own catch would kill the drift vector without touching control flow. Fine as a follow-up.

tests/proxy.test.ts
```diff
@@ -910,3 +916,126 @@ describe('anthropic passthrough debug logging', () => {
...
+      const [, init] = vi.mocked(fetch).mock.calls[0]!;
+      expect((init?.headers as Record<string, string>).Authorization).toBe(
+        'Bearer fresh-oauth-token',
+      );
+    } finally {
+      handle.close();
+    }
+  });
+
+  it('rebuilds the translated SDK route and retries once after an OAuth 401', async () => {
```
bman654, line 1097: **tests (major ask):** this and the two server-router siblings cover only the happy cell: 401 → changed token → success. The *safety* cells — retry happens at most once (second 401 surfaces, no loop), no retry after output has started, unchanged-token replacement refused — are pinned only on the isolated `fetchWithOAuthRetry` unit tests, so a regression in these hand-inlined loops (e.g. dropping the `sdkAttempt` guard or the `headersSent` check) would pass the suite. One test per path returning 401 twice, and one that 401s after the first SSE chunk was written, would pin the contract where it actually lives. There's also no integration test driving a 401 through the native passthrough wiring in either the endpoint proxy or the server router.

tests/fixtures/credential-helper.mjs
```diff
@@ -25,6 +25,45 @@ try {
...
+    if (firstGet) {
+      const releasePath = `${storePath}.release-stale-get`;
+      const deadline = Date.now() + 5_000;
+      while (!existsSync(releasePath) && Date.now() < deadline) {
+        await new Promise(resolve => setTimeout(resolve, 5));
+      }
+      if (!existsSync(releasePath)) process.exit(1);
+    }
+  }
+  if (mode === 'interleave-readback') {
```
bman654, line 47: **tests:** this `interleave-readback` mode is exactly the tool for the untested branches of the refresh state machine — the `currentRaw !== raw` compare-and-set re-read, the three-generation exhaustion error, and the inflight-join recursion when the joined result equals the caller's rejected token — but no test activates it. Those branches are the most likely to hide bugs and currently ship unexercised; two or three tests using this mode would close the gap the fixture was clearly built for.

tests/oauth.test.ts
```diff
@@ -52,6 +68,50 @@ describe('oauth refresh http', () => {
...
+      new URLSearchParams({ grant_type: 'refresh_token' }),
+      {
+        contentType: 'form',
+        errorPrefix: 'token refresh failed',
+      },
+    );
+
+    expect(fetchMock).toHaveBeenCalledWith(
+      'https://auth/token',
+      expect.objectContaining({ signal: expect.any(AbortSignal) }),
```
bman654: **tests (minor):** this asserts a signal *exists*, not that it aborts — the 30s bound is never proven (fake timers + a hung fetch would do it deterministically). Separately, this file stubs global `fetch` via `vi.stubGlobal` but cleans up with `vi.restoreAllMocks()`, which does not undo stubbed globals — use `vi.unstubAllGlobals()` so a failure here can't leak the stub into later files.

src/env.ts
```diff
@@ -368,68 +418,159 @@ export async function resolveProviderOAuthProviderData(
   return parseStoredOAuthCredential(raw)?.providerData;
 }
 
-function decodeProviderSecret(raw: string | null): string | null {
+function decodeProviderSecret(raw: string | null, allowOpaqueJson = false): string | null {
```
bman654, line 432: **minor (undeclared ride-along):** the `allowOpaqueJson` semantics change means unknown-JSON secrets on OAuth accounts that previously resolved as trimmed raw text now resolve to `null` — a behavior change for anyone storing opaque JSON credentials, and notably the strand that visibly churned inside this PR's own commits (commit 2's "preserve opaque structured credentials" corrects commit 1's regression here). It's defensible, but it's neither in the commit bodies nor the PR description; in a release-please repo the commit body is the changelog. Please declare it (same for the `/models` stripping of pre-existing `oauthAccountId`/`providerData` — welcome hardening, also undeclared).

bman654 (changes requested):
Thanks — the review panel (cache/retry correctness, security, tests, architecture) verified the core design holds: the 3-generation refresh state machine is sound (no unbounded recursion, never returns a known-rejected token, forced-refresh failures don't fall back to the rejected token), retry-once is correctly bounded and `headersSent`-guarded on every dispatch path, the WebSocket partitioning already fingerprints authorization so a stale socket can't be reused with the old token, the `accessRejected` schema evolution is compatible in both directions (old parsers ignore it; it self-heals when refresh yields a different token), no token material lands in logs, and `/models` stripping actually fixes a pre-existing metadata exposure. Suite is green in my environment: 70 files / 732 tests + typecheck, listener suites included. Composition with #15 was checked — clean textually and semantically.

**Why request-changes — one real regression:**

1. **(major)** Cross-process cache staleness. The cache-hit shortcut serves any unexpired token without re-reading the store, and invalidation only happens in the process that mutated the credential. A long-running `clodex server` keeps dispatching (and billing) as account A for up to the token's remaining lifetime after `clodex providers auth` in another terminal switched to account B. `main` re-read the store on every resolution, and this PR's own summary says "resolve current stored credentials before dispatch" — the shortcut breaks that contract. Inline comment on `src/env.ts` with options.

**Secondary asks (inline):** pin the retry safety cells (at-most-once, no-retry-mid-stream, unchanged-token) on the three real dispatch paths rather than only the isolated helper; exercise the compare-and-set/generation-exhaustion branches — your own fixture's `interleave-readback` mode was built for exactly this and no test activates it; actually bound-test the 30s refresh timeout; declare the two ride-alongs (`decodeProviderSecret` opaque-JSON semantics and the `/models` field stripping) in the commit body; and document the `accessRejected` escape hatch. None of it needs a redesign.

src/env.ts
camjac251: Addressed in `491467d`. The cache now holds only access-token metadata and expires after 30 seconds; clock regression, expiration, rejection, or the caller's rejected token bypasses it immediately. The nearby comment states the cross-process staleness bound, and a regression replaces the helper credential externally and verifies the new value is read at the bound.

src/env.ts
camjac251: Addressed in `491467d`. `CachedOAuthCredential` contains only `access`, `expires`, the rejected marker, and `checkedAt`. Refresh tokens, account IDs, and provider data are no longer retained in the long-lived cache; the complete record is reread only when refresh work needs it.

src/oauth/types.ts
camjac251: Addressed in `491467d`. The credential-helper documentation now states that a different refreshed token clears the rejected marker and that reauthentication is required when the identity provider continues returning the same rejected token.

src/proxy.ts
camjac251: Implemented in `491467d`. `resolveOAuthRetryReplacement()` centralizes the pure enablement, 401, attempt, output-started, resolver, refresh-failure, and unchanged-token checks. Each native or translated dispatch loop keeps its existing control flow and route-specific error rendering.

tests/proxy.test.ts
camjac251: Addressed in `491467d`. Route-level tests now cover a second 401 for the endpoint path and both server-facing translated paths, an unchanged replacement token, and a rejection after the first streamed event. A real-listener test also drives native passthrough through 401, replacement resolution, and successful retry.

tests/fixtures/credential-helper.mjs
camjac251: Addressed in `491467d`. The fixture's interleaved readback now returns the original generation on the first read and waits for the replacement marker on later reads. Tests use it for compare-and-set replacement adoption, three-generation exhaustion, and in-flight join recursion when the joined result equals the caller's rejected token.

tests/oauth.test.ts
camjac251: Addressed in `491467d`. Refresh requests now use an explicit 30-second abort timer that is unreferenced and always cleared. A hung-fetch test verifies the signal remains live at 29,999 ms, aborts at 30,000 ms with `TimeoutError`, and global fetch stubs are removed during cleanup.

src/env.ts
camjac251: Declared in `491467d` and the updated PR description. Valid opaque JSON remains available only for non-OAuth references, while unknown JSON on OAuth references fails closed. Public model listings omit credential references, OAuth account IDs, and provider data; the behavior is now also documented in the credential-helper guide.

src/env.ts
```diff
@@ -368,68 +429,178 @@ export async function resolveProviderOAuthProviderData(
...
+      }
       return refreshed.access;
-    } catch (err) {
-      diag?.(err instanceof Error ? err.message : String(err));
-      if (cred.access && cred.expires > Date.now()) return cred.access;
-      throw err;
     }
+    throw new Error('OAuth credential changed repeatedly while refresh was in progress');
+  }, {
+    waitMs: OAUTH_REFRESH_LOCK_WAIT_MS,
```
bman654, line 579: Refresh callers get `OAUTH_REFRESH_LOCK_WAIT_MS = 150_000` to wait out a holder that can legitimately spend up to ~30s per HTTP refresh across the three-generation loop — but the paths that must *supersede* a rejected credential don't get the same budget. `clodex providers auth` persistence (and #15's `persistOAuthProvider`) acquires the same per-credential mutation lock with the 30-second default wait. In the exact documented recovery scenario — `accessRejected` persisted, a busy server refreshing in a loop — the user can complete device authorization and then have the save time out behind the refresh holder it's meant to replace. Failure is safe (no corruption) but the escape hatch is unreliable while the server is active. Credential-mutation paths that supersede a rejected token should wait at least as long as the bounded refresh holder (or the design should otherwise guarantee reauth can acquire the lock).

bman654 (changes requested):
Re-reviewed at 9ef635d against our previous review. The headline concern is resolved: the credential cache is now metadata-only (`access`/`expires`/rejected/`checkedAt`) with a hard 30-second TTL from the last actual store read — no sliding expiration, and the regression test genuinely replaces the credential externally (bypassing in-process invalidation) and observes the new value at the boundary. Cross-process staleness is now bounded exactly as documented. Verified locally: typecheck + full suite green (70 files / 743 tests).

Also verified addressed: refresh tokens/account ids/provider data no longer retained long-lived; `resolveOAuthRetryReplacement()` unifies the retry gates across all three dispatch paths plus native passthrough, with route-level second-401 / unchanged-token / output-started tests on the real paths; the `interleave-readback` fixture now exercises compare-and-set adoption, three-generation exhaustion, and joined-result recursion; the 30s refresh abort has an explicit unreffed timer, always cleared, with the hung-fetch boundary test (live at 29,999 ms, TimeoutError at 30,000 ms); 9ef635d correctly extends that same deadline through response-body consumption without introducing any size cap; the opaque-JSON and `/models` ride-alongs are declared and documented.

One medium remains — a lock-wait asymmetry that undermines the documented `accessRejected` escape hatch (inline comment) — plus one small test-cleanup leftover:

- `tests/oauth.test.ts` — the new refresh-HTTP block correctly calls `vi.unstubAllGlobals()`, but the sibling `describe('oauth refresh')` still cleans up with only `vi.restoreAllMocks()` after `vi.stubGlobal('fetch', …)`; `restoreAllMocks()` doesn't undo `stubGlobal`, so that block still leaks its stub. One file-level `unstubAllGlobals()` would cover both.

(Also: the PR description's test count is one stale — head runs 743.)

Rebase note: #21 is landing ahead of this, so a rebase is needed (proxy/router files). Composes cleanly with #15 at current heads (merge-tree clean; #15's two-arg `withCredentialMutationLock` callers remain compatible with the optional `options` added here). There's also a `src/proxy.ts` conflict with #10 — in that resolution #10's route-unavailable rejection must stay ahead of this PR's token-count/refresh handling.

src/env.ts
camjac251: Addressed in `a93fdd9`. Credential-mutation locks now default to a bounded 150-second wait, matching the legitimate refresh-holder budget while preserving explicit caller overrides. The regression keeps a contender pending through 149,999 ms and confirms it acquires the same credential lock after release. The remaining fetch-stub cleanup is also restored in `afterEach`.

bman654 (approved):
Approving. Both outstanding items verified fixed:

- Mutation-lock default is now 150s (matching the refresh-holder budget) with explicit overrides still honored, so the `accessRejected` escape hatch no longer times out behind the refresh loop.
- The global fetch-stub leak is resolved (`unstubAllGlobals` in both `oauth.test.ts` describes).

Sweep of `8af2e48..a93fdd9` found no new defects.

PR #17, opened 2026-07-21
fix(auth): make chunked credentials crash-safe

camjac251:
## Summary

- journal chunk generations and deletion state before publishing credential markers
- use stable config-scoped credential accounts for retryable provider ownership
- serialize provider, credential, and registry mutations across processes
- encrypt filesystem recovery intent with a random per-account key kept in the OS credential store
- allow direct reauthorization after sentinel checks prove the main and chunk namespaces are empty
- preserve opaque, malformed, older, and newer journal data without modifying credentials or chunks
- retire known unverifiable deletion state only after a complete empty-inventory proof
- regenerate unsupported managed-state key material only after proving the credential namespace empty
- link sentinel account and value identities before recognizing removable bookkeeping
- keep credential coordination tests inside temporary native-home sandboxes

## Unreleased branch note

Anyone who ran a commit from this PR before `97c766f` should remove only the unreleased branch's old `clodex-journal:*` entries using the `__relay_chunk_journal__:v1:` prefix before moving to the current branch. Released versions did not write these journals. The current code intentionally treats unknown journal formats as opaque and does not alter their credential state.

## Test plan

- [x] focused keyring regression suite: 101 tests
- [x] complete suite on current `main`: 80 files, 1011 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] staged secret scan
- [x] focused review of recovery, encryption, deletion, and test isolation

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,17 +214,33 @@ clodex --version    # version
 ## Configuration
 
 - Config home: `~/.clodex` (override with `CLODEX_HOME`). On first run, config is migrated automatically from a legacy `~/.relay-ai` directory if present; the legacy directory is never modified.
-- The config-home filesystem must support hard links because registry locks are
-  published atomically. Keep `CLODEX_HOME` on a local filesystem rather than
-  FAT, exFAT, or a network mount that rejects hard links. An abrupt process kill
-  during lock publication can leave a `*.lock.*.tmp` file; it does
-  not block later lock acquisition and can be removed when no Clodex process is
+- The config-home filesystem and the native account home must support hard
+  links because registry and credential locks are published atomically. Keep
+  `CLODEX_HOME` and `~/.clodex/credential-locks` on local filesystems rather
+  than FAT, exFAT, or a network mount that rejects hard links. An abrupt process
+  kill during lock publication can leave a `*.lock.*.tmp` file; it does not
+  block later lock acquisition and can be removed when no Clodex process is
   running. A canonical `providers.json.lock` whose recorded PID is no longer
-  running is reclaimed automatically on the next lock acquisition. If it remains
-  while that PID is active, stop every Clodex process and verify the recorded PID
-  before removing the lock manually. Never remove the canonical lock while a
-  Clodex process is active.
-- Credentials live in the OS credential store (Keychain / Windows Credential Manager / Secret Service) under the `clodex` service. Set `CLODEX_CREDENTIAL_HELPER` to an absolute executable path to use an external secure store instead; see [credential helpers](docs/credential-helpers.md).
+  running is reclaimed automatically on the next lock acquisition. If it
+  remains while that PID is active, stop every Clodex process and verify the
+  recorded PID before removing the lock manually. Never remove the canonical
+  lock while a Clodex process is active.
+- Credentials live in the OS credential store (Keychain / Windows Credential
+  Manager / Secret Service). The `clodex` service holds the main value or
+  published marker; `clodex-chunks` holds current long-credential chunks;
+  `clodex-journal` holds crash-recovery metadata and a deletion marker; and
+  `clodex-deleted` holds a redundant non-secret deletion guard. Use Clodex
+  provider removal instead of deleting these entries individually. Non-secret
+  per-account managed-state markers live under the native OS account home at
+  `~/.clodex/keyring-state`; before each journal write they record the exact
+  non-secret intent so a retry can replay and verify it. They also prevent a
+  temporarily unavailable keyring journal from being mistaken for an absent
+  one. Credential mutation locks live beside that state at
+  `~/.clodex/credential-locks`. Both paths are independent of `CLODEX_HOME` and
+  runtime or temporary-directory environment overrides. This keeps concurrent
+  processes serialized when they use different config homes. Set
+  `CLODEX_CREDENTIAL_HELPER` to an absolute executable path to use an external
+  secure store instead; see [credential helpers](docs/credential-helpers.md).
 - Proxied routes forward configured provider headers for API-key and OAuth authentication. Anonymous routes preserve non-credential headers while removing authorization, API-key, cookie, token, secret, and credential-bearing header names before dispatch.
 - `CLODEX_CLAUDE_PATH` overrides Claude Code binary discovery.
 - **Outbound proxy:** when `HTTP_PROXY`/`HTTPS_PROXY` (and optionally `NO_PROXY`) are set in clodex's environment, all clodex-originated network calls honor them — OAuth sign-in and token refresh, model-list and models.dev refreshes, upstream OpenAI API calls, and the ChatGPT/Codex OAuth WebSocket transport (tunneled via HTTP CONNECT).
--- a/README.md
+++ b/README.md
@@ -229,13 +229,19 @@ clodex --version    # version
   Manager / Secret Service). The `clodex` service holds the main value or
   published marker; `clodex-chunks` holds current long-credential chunks;
   `clodex-journal` holds crash-recovery metadata and a deletion marker; and
-  `clodex-deleted` holds a redundant non-secret deletion guard. Use Clodex
-  provider removal instead of deleting these entries individually. Non-secret
-  per-account managed-state markers live under the native OS account home at
-  `~/.clodex/keyring-state`; before each journal write they record the exact
-  non-secret intent so a retry can replay and verify it. They also prevent a
-  temporarily unavailable keyring journal from being mistaken for an absent
-  one. Credential mutation locks live beside that state at
+  `clodex-deleted` holds a redundant non-secret deletion guard. A
+  `clodex-state-key` entry protects each account's recovery metadata. Use
+  Clodex provider removal instead of deleting these entries individually.
+  Authenticated encrypted per-account managed-state markers live under the
+  native OS account home at `~/.clodex/keyring-state`; before each journal
+  write they record the exact recovery intent so a retry can replay and verify
+  it. The encryption key remains in the OS credential store, so the filesystem
+  marker alone cannot be used to test credential guesses. The marker also
+  prevents a temporarily unavailable keyring journal from being mistaken for
+  an absent one. If the OS credential store was completely reset, Clodex
+  permits direct reauthorization only after sentinel checks prove the main and
+  chunk namespaces are empty. Hidden, locked, or partially restored state
+  remains fail-closed. Credential mutation locks live beside that state at
   `~/.clodex/credential-locks`. Both paths are independent of `CLODEX_HOME` and
   runtime or temporary-directory environment overrides. This keeps concurrent
   processes serialized when they use different config homes. Set
--- a/docs/credential-helpers.md
+++ b/docs/credential-helpers.md
@@ -59,6 +59,68 @@ credential instances. It rejects symbolic links, foreign ownership, broad
 permissions on POSIX, files over 1 MiB, and more than 1,024 queued entries
 before attempting any credential-store deletion.
 
+## OS keyring layout and compatibility
+
+The default OS credential-store backend uses four service namespaces:
+
+- `clodex` stores a short credential directly or publishes the marker for a
+  long credential;
+- `clodex-chunks` stores the chunks for current long credentials;
+- `clodex-journal` records crash recovery, the active chunk generation, and a
+  deletion marker;
+- `clodex-deleted` stores a redundant non-secret deletion guard.
+
+Clodex also keeps a non-secret per-account managed-state marker under the
+native OS account home at `~/.clodex/keyring-state`. Before each
+cleanup-journal write, the marker records the exact non-secret journal intent.
+A retry republishes and verifies that intent before continuing, then marks it
+managed. If the OS keyring temporarily reports a managed journal as absent,
+the marker makes reads, writes, and deletes fail closed instead of replacing
+unknown chunk inventory. Malformed local intent also remains fail-closed.
+Credential mutation locks live beside that state under
+`~/.clodex/credential-locks`. Neither path depends on `CLODEX_HOME`,
+`XDG_RUNTIME_DIR`, or temporary-directory environment variables because the OS
+keyring service and account namespaces are shared across those process-local
+settings. The native account-home filesystem must support hard links so lock
+publication remains atomic.
+
+New provider credentials use a stable, versioned account instance owned by the
+provider slot and selected credential backend. A retry derives the same
+candidate, so an ambiguous result cannot make its reference unreachable.
+Provisioning resumes well-formed candidate state, while unavailable or
+malformed recovery metadata remains fail-closed. Refresh paths replace the
+registry's current account only when its prior keyring state can be confirmed.
+Reauthorization provisions the selected backend first, then updates the
+registry after read-back verification. If the keyring hides both the main value
+and its metadata, replacement and deletion stop without publishing new state
+or reporting success.
+
+The active-generation journal is live metadata, not stale debris. Clodex keeps
+one generation after a successful long-credential write so a later release can
+retire the chunks through a current provider-removal operation if an older
+release removes or replaces only the main marker. Use the Clodex
+provider-removal path instead of deleting one of these entries manually.
+
+Long chunked credentials are not readable by older releases that do not
+understand their marker format. If a downgrade removes the main marker while
+leaving chunks behind, passive resolution preserves the recorded inventory
+because a missing keyring value cannot be distinguished from a collapsed read
+error on every platform. Remove the provider with a current Clodex release
+before reauthorizing to retire the orphaned generation. A published marker that
+does not match its recovery journal fails closed and leaves every recorded
+generation intact.
+
+Clodex does not implicitly import credentials from the legacy `relay-ai`
+service. Existing legacy entries remain untouched. Reauthorize or explicitly
+save the provider credential to publish it under the `clodex` service, verify
+the new credential, and remove the old entry separately if it is no longer
+needed. Provider removal deletes only the Clodex credential. Redundant
+non-secret deletion guards keep ambiguous deleted state from becoming readable;
+an explicit later credential save clears those guards. Unknown JSON-shaped
+values in non-OAuth keyring and helper accounts remain opaque. Historical
+`wellknown` token and OAuth access envelopes retain their existing decoding
+behavior, while structured OAuth validation applies only to OAuth accounts.
+
 ## Protocol
 
 The helper receives one of these invocations:
--- a/docs/credential-helpers.md
+++ b/docs/credential-helpers.md
@@ -61,22 +61,36 @@ before attempting any credential-store deletion.
 
 ## OS keyring layout and compatibility
 
-The default OS credential-store backend uses four service namespaces:
+The default OS credential-store backend uses five service namespaces:
 
 - `clodex` stores a short credential directly or publishes the marker for a
   long credential;
 - `clodex-chunks` stores the chunks for current long credentials;
 - `clodex-journal` records crash recovery, the active chunk generation, and a
   deletion marker;
-- `clodex-deleted` stores a redundant non-secret deletion guard.
+- `clodex-deleted` stores a redundant non-secret deletion guard;
+- `clodex-state-key` stores a random per-account key that protects the
+  filesystem recovery marker.
 
-Clodex also keeps a non-secret per-account managed-state marker under the
-native OS account home at `~/.clodex/keyring-state`. Before each
-cleanup-journal write, the marker records the exact non-secret journal intent.
-A retry republishes and verifies that intent before continuing, then marks it
-managed. If the OS keyring temporarily reports a managed journal as absent,
-the marker makes reads, writes, and deletes fail closed instead of replacing
-unknown chunk inventory. Malformed local intent also remains fail-closed.
+Clodex also keeps an authenticated encrypted per-account managed-state marker
+under the native OS account home at `~/.clodex/keyring-state`. Before each
+cleanup-journal write, the marker records the exact journal intent. A retry
+decrypts, republishes, and verifies that intent before continuing, then marks
+it managed. The encryption key remains only in the OS credential store, so a
+copied filesystem marker does not expose credentials or an offline credential
+confirmation value. If the OS keyring temporarily reports a managed journal as
+absent, the marker makes reads, writes, and deletes fail closed instead of
+replacing unknown chunk inventory. Malformed or unauthenticated local intent
+also remains fail-closed.
+
+If the filesystem marker outlives a complete OS credential-store reset,
+Clodex allows direct reauthorization only after sentinel-backed service
+enumeration proves that the main credential and chunk namespaces are empty.
+The recovery first records durable deleted state, then follows the normal
+journal-before-publication path for the replacement. A hidden, locked,
+incomplete, or partially restored namespace cannot take this path and remains
+fail-closed.
+
 Credential mutation locks live beside that state under
 `~/.clodex/credential-locks`. Neither path depends on `CLODEX_HOME`,
 `XDG_RUNTIME_DIR`, or temporary-directory environment variables because the OS
--- a/src/credential-helper.ts
+++ b/src/credential-helper.ts
@@ -1,10 +1,14 @@
 import { spawn } from 'node:child_process';
 import { createHash } from 'node:crypto';
 import { accessSync, constants, statSync } from 'node:fs';
-import { isAbsolute, normalize } from 'node:path';
+import { isAbsolute, normalize, resolve } from 'node:path';
+import { getAppHome } from './paths.js';
 
 export const CREDENTIAL_HELPER_ENV = 'CLODEX_CREDENTIAL_HELPER';
 export const CREDENTIAL_HELPER_SERVICE = 'clodex';
+
+const CREDENTIAL_ACCOUNT_INSTANCE_SEPARATOR = '::credential::';
+const CREDENTIAL_ACCOUNT_INSTANCE_PATTERN = /^v1:[0-9a-f]{32}$/;
 
 const HELPER_TIMEOUT_MS = 10_000;
 const HELPER_MAX_OUTPUT_BYTES = 1024 * 1024;
@@ -57,6 +61,30 @@ export function credentialAuthRef(account: string): string {
   return helper
     ? `helper:v1:${helper.id}:${account}`
     : `keyring:${account}`;
+}
+
+export function credentialInstanceAuthRef(account: string): string {
+  const configScope = createHash('sha256')
+    .update('clodex-credential-account\0')
+    .update(normalize(resolve(getAppHome())))
+    .digest('hex')
+    .slice(0, 32);
+  return credentialAuthRef(
+    `${account}${CREDENTIAL_ACCOUNT_INSTANCE_SEPARATOR}v1:${configScope}`,
+  );
+}
+
+export function isCredentialAccountInstance(account: string): boolean {
+  const separatorIndex = account.lastIndexOf(CREDENTIAL_ACCOUNT_INSTANCE_SEPARATOR);
+  if (separatorIndex <= 0) return false;
+  return CREDENTIAL_ACCOUNT_INSTANCE_PATTERN.test(
+    account.slice(separatorIndex + CREDENTIAL_ACCOUNT_INSTANCE_SEPARATOR.length),
+  );
+}
+
+export function credentialAccountBase(account: string): string {
+  if (!isCredentialAccountInstance(account)) return account;
+  return account.slice(0, account.lastIndexOf(CREDENTIAL_ACCOUNT_INSTANCE_SEPARATOR));
 }
 
 async function runCredentialHelper(
--- a/src/env.ts
+++ b/src/env.ts
@@ -1736,6 +1736,30 @@ function recoverEmptyOpaqueManagedKeyringState(
   return true;
 }
 
+function retireUnreadableManagedStateForDeletion(
+  keyring: KeyringApi,
+  account: string,
+  diag?: (msg: string) => void,
+): boolean {
+  try {
+    readKeyringManagedState(keyring, account);
+    return true;
+  } catch (err) {
+    diag?.(classifyKeyringError(err));
+    if (!(err instanceof KeyringManagedStateKeyUnavailableError)) {
+      return false;
+    }
+  }
+  try {
+    removeKeyringManagedState(account);
+    diag?.('retired unreadable managed credential metadata for explicit deletion');
+    return true;
+  } catch (err) {
+    diag?.(classifyKeyringError(err));
+    return false;
+  }
+}
+
 function writeKeyringAccountLocked(
   keyring: KeyringApi,
   account: string,
@@ -1969,6 +1993,7 @@ function deleteJournalLessManagedKeyringAccount(
   keyring: KeyringApi,
   account: string,
   diag?: (msg: string) => void,
+  retireUnreadableManagedState = false,
 ): boolean {
   const sentinels: KeyringEnumerationSentinel[] = [];
   let sentinelsRemoved = false;
@@ -2028,6 +2053,10 @@ function deleteJournalLessManagedKeyringAccount(
       }
     }
     sentinelsRemoved = true;
+    if (retireUnreadableManagedState) {
+      removeKeyringManagedState(account);
+      diag?.('retired unreadable managed credential metadata for explicit deletion');
+    }
     writeKeyringJournal(keyring, account, {
       mode: 'deleted',
       generations: [],
@@ -2066,11 +2095,11 @@ async function deleteKeyringAccount(
           managedState = readKeyringManagedState(keyring, account);
         } catch (err) {
           diag?.(classifyKeyringError(err));
-          if (
-            err instanceof KeyringManagedStateKeyUnavailableError &&
-            recoverEmptyOpaqueManagedKeyringState(keyring, account, diag)
-          ) {
-            return true;
+          if (err instanceof KeyringManagedStateKeyUnavailableError) {
+            if (recoverEmptyOpaqueManagedKeyringState(keyring, account, diag)) {
+              return true;
+            }
+            return deleteJournalLessManagedKeyringAccount(keyring, account, diag, true);
           }
           return false;
         }
@@ -2082,12 +2111,17 @@ async function deleteKeyringAccount(
             pendingJournalRaw = encodeKeyringJournal(managedState.journal);
             diag?.('restored keyring cleanup journal from managed state');
           }
+        } else if (readKeyringDeletionGuard(keyring, account) !== null) {
+          return reconcileKeyringJournal(keyring, account, diag);
         }
       }
       let pendingJournal: KeyringChunkJournal | null = null;
       try {
         if (pendingJournalRaw !== null) {
           pendingJournal = parseKeyringChunkJournal(pendingJournalRaw);
+          if (!retireUnreadableManagedStateForDeletion(keyring, account, diag)) {
+            return false;
+          }
           if (pendingJournal.mode === 'deleted') {
             return reconcileKeyringJournal(keyring, account, diag);
           }
--- a/src/env.ts
+++ b/src/env.ts
@@ -1,6 +1,8 @@
 // src/env.ts
 import { CONFLICTING_ENV_VARS } from './constants.js';
 import { createHash, randomUUID } from 'node:crypto';
+import { homedir } from 'node:os';
+import { join } from 'node:path';
 import {
   deleteCredentialHelperAccount,
   readCredentialHelperAccount,
@@ -14,7 +16,10 @@ import {
   type StoredOAuthCredential,
 } from './oauth/types.js';
 import { refreshStoredOAuthCredential, oauthCredentialShouldRefresh } from './oauth/refresh.js';
-import { withCredentialMutationLock } from './registry/lock.js';
+import {
+  withCredentialMutationLock,
+  withRegistryWriteLock,
+} from './registry/lock.js';
 import type { ConflictInfo } from './types.js';
 import { removeAnthropicProxyBypass } from './wrapper-env.js';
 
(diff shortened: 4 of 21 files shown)


bman654 (changes requested):
Reviewed the full change (a four-lens pass: security, crash-safety/correctness, test-quality, architecture) at 878015c.

First: this is careful, well-built work, and most of the hard parts are verified sound. Journal-before-publish ordering is correct (the `{preparing}` managed-state file is fsync'd + dir-synced before the keyring journal, which is written before any chunk; the active pointer is the last publication step and generations are UUID-namespaced so a new generation never overwrites the old); reconcile-before-write means the write/delete-generation bounds are never exceeded and the active set is never mistakenly cleaned; a `>128`-chunk payload fails loud *before* any chunk is written (no silent truncation); temp → fsync → rename → dir-sync is atomic; distinct `CLODEX_HOME`s are isolated while a genuinely-shared account is serialized across homes; and the happy-path legacy read + in-place adoption works. Security is clean — no API key or OAuth token is written to the filesystem (only PIDs, tokens, UUIDs, and SHA-256 digests, all at `0600`/`0700`), chunks are tombstone-overwritten before deletion, and the cross-wiring the PR advertises is genuinely enforced by config-scoped, provider-namespaced account ids. The tests are thorough and mostly behavioral (interrupt→recover, corruption→fail-closed, legacy backward-compat, real cross-process serialization).

One blocker before merge, plus a few optional nice-to-haves.

---

## Blocker — a `{managed}` state file whose keyring journal has vanished permanently bricks the account (read, re-provision, and delete all fail-closed, with no self-heal or in-product remediation)

Once a credential is stored, its state lives in two places: the keyring `clodex-journal` entry, and an on-disk marker at `~/.clodex/keyring-state/<sha256(account)>.managed` (= `v1:managed`). If the OS keyring entries are cleared **without** removing `~/.clodex/keyring-state`, the `.managed` file survives but the journal entry is gone. That combination is not exotic:
- the standard "delete the keychain item and re-login" troubleshooting step,
- a fresh/relocked login keychain after an OS upgrade or keychain reset,
- (strongest) restoring `~/.clodex` from a dotfiles/backup sync onto a machine whose OS keychain is empty.

In that state, `reconcileKeyringJournal` reaches the `rawJournal === null && deletionGuard === null && managedState !== null` branch (~`src/env.ts:742`) and returns `false` **unconditionally, without clearing the stale marker** — so it never self-heals. That `false` then gates all three public entry points:
- **read** returns `null` — `readKeyringAccount` at `if (finalJournalRaw === null && readKeyringManagedState(account)?.mode === 'managed') return null` (~`src/env.ts:1290`),
- **re-provision** returns `false` — `writeKeyringAccountLocked`'s `if (!reconciled) return false` (~`src/env.ts:1401`), so re-login **silently fails**,
- **delete** returns `false` — `deleteKeyringAccount`'s matching reconcile gate (~`src/env.ts:1578`).

Net effect: the account is bricked. The credential reads as absent, re-authenticating doesn't fix it, and there is no in-product way to clear it — the only escape is manually deleting an opaque `sha256(account).managed` file. The diagnostic ("managed keyring cleanup metadata is temporarily unavailable") is also misleading, since for a genuinely-absent (vs briefly-locked) entry the condition is permanent.

Suggested direction: when the journal is genuinely absent and there is no deletion guard, treat the stale `{managed}` marker as recoverable — clear it (`removeKeyringManagedState`) and return `true` so a subsequent provision can re-establish the credential from scratch — or otherwise distinguish "keyring briefly locked" from "entry gone" so the fail-closed path can't become permanent. I found this by static trace through the three gates; you can confirm it live by storing a credential, `security delete-generic-password`-ing the `clodex-journal` service item (leaving `~/.clodex` intact), then retrying resolve and re-login.

---

## Non-blocking — nice to have if the fix is simple, otherwise fine to skip

1. **Best-effort legacy adoption.** On the first read of a pre-upgrade credential, `readKeyringAccount` reads the value and then performs an adoption `writeKeyringJournal(...)`. If that write throws — the native `~/.clodex` not being writable (disk full / read-only mount), or a keyring write rejection — the enclosing `catch` returns `null` (~`src/env.ts:1324`), **discarding the value that was just read successfully** and forcing an unnecessary re-login. It's a narrow, one-time-per-credential window and low-likelihood, but since the value is already in hand, wrapping the adoption write so it returns `value` even when the journal write fails removes the sharp edge (adoption simply retries on the next read). Pre-change, that read was pure and always returned the value.

2. **Test the `>128`-chunk guard.** The overflow guard (~`src/env.ts:1511`) is correct — it throws before writing any chunk — but nothing exercises it. A test that a `>128`-chunk payload resolves `false`, emits the guard diagnostic, and leaves no orphaned chunks (plus a read-side test that a seeded `>128`-count marker resolves `null`) would lock the behavior in.

3. **End-to-end concurrent-writer test.** The "serialize per-account mutations, no cross-wiring" guarantee is demonstrated at the lock primitive (the cross-process `registry-lock` test) but only inferred for the public API. Two concurrent `saveProviderCredential` calls on the same account, asserting the resolved value is exactly one of the two and the losing generation's chunks are fully cleaned, would demonstrate the headline guarantee directly.

---

Happy to approve once the blocker's recovery path is closed; the three items above are genuinely optional. Nice work on the core — the crash-safety machinery holds up under scrutiny.

camjac251:
The managed-state recovery path has been updated.

- New managed-state files retain the exact non-secret cleanup journal, so a missing keyring copy can be reconstructed without guessing from repeated empty reads.
- Temporarily hidden journals remain fail-closed for reads and provisioning.
- Existing journal-less markers can be cleared through explicit deletion only after guarded main-entry removal and verified enumeration and cleanup of both chunk namespaces.
- The optional legacy-adoption, oversized-credential, and concurrent-writer cases are also covered.

Fresh verification completed with 81 focused keyring tests, 969 tests across the full suite, type checking, a production build, diff checks, and changed-file secret scans.

bman654 (changes requested):
Re-reviewed at `9614ec1` with a four-lens pass (security / correctness / test-quality / architecture). Thanks for the fast turnaround on this — the recovery work is a real improvement.

**Headline: the round-1 blocker is downgraded, not closed, and the fix introduced one new issue worth addressing before merge.**

Verified good: the branch merges cleanly over `main` (now 2.1.0) with no conflicts, and the suite is green (81/81 in the keyring file, 969 across the suite).

---

## 1. C1 — improved, but plain re-login still fails · MAJOR (was BLOCKER)

Credit where it's due: embedding the journal in the marker genuinely fixes the *bricked* part. The journal is restored from the marker at `src/env.ts:809-822`, and `delete` now succeeds against an empty keychain — so recovery no longer requires manually `rm`-ing an opaque hash file. That was the worst part of round 1 and it's gone.

But the headline symptom is still there. Traced through all three entry points for the real scenario (`~/.clodex` restored from backup onto an empty keychain, marker in the new `v1:managed:<journal>` format):

- `read` → `null` — correct.
- `delete` → `true`, and re-provision after that works.
- **`provision` / `save` → `false`, permanently.**

The restored journal describes a *published* generation, so reconcile never reaches the chunk handling — it stops at the `activeValue === null && journal.unpublished !== true` gate (`src/env.ts:996-1000` for the chunked branch, `src/env.ts:1103-1107` for the short branch) and returns `false`, gating provision at `src/env.ts:1508`.

So a user who restores their config directory hits a stable, cryptic failure — `published credential state cannot be confirmed while cleanup metadata is active` — on OAuth re-login, and has to *guess* that the fix is "remove the provider, then re-add it". `tests/keyring-credentials.test.ts:1267-1289` pins exactly this two-step contract, so I take it as deliberate rather than an oversight.

Two ways to close it, either is fine:

- **Preferred:** let provision recover directly. The enumeration sentinel you already built is exactly the tool for this — it can prove the namespace is genuinely *empty* rather than *hidden*, which is the distinction the gate is (correctly) refusing to guess at.
- **Floor:** if the two-step contract is intentional, make it discoverable — the diag should name the recovery action, and `docs/credential-helpers.md` should document it. A correct-but-unguessable recovery path is still a support burden.

## 2. NEW: the marker now persists an unsalted digest of the live credential · MAJOR

This one is a regression introduced by this round's fix, so it wasn't in round 1.

The managed-state marker used to be the literal `v1:managed\n`. It now carries the journal inline, and the journal carries credential digests — `shortDigest` is `sha256(rawValue)` (`src/env.ts:1426`), and chunked generations carry `digest = sha256(full credential)`. `base64url` is encoding, not encryption, so `~/.clodex/keyring-state/<hash>.managed` is now a plainly decodable file containing a hash of the live secret. Confirmed by decoding a marker the code would write:

```
v1:managed:eyJtb2RlIjoic2hvcnQi...
  → {"mode":"short","generations":[],"shortDigest":"<sha256 of the credential>"}
```

A failed rotation leaves *both* the old and new digests on disk.

Round 1 explicitly certified "no plaintext secret to disk". That's still literally true, but the posture changed, and the docs now overclaim: `docs/credential-helpers.md:71,73,75,118` and `README.md:232,236` all describe this marker as **non-secret**. As written it's a confirmation/correlation oracle, and it's brute-forceable for a low-entropy key pasted into a custom endpoint. It also partially steps around the macOS keychain ACL boundary — the whole point of putting this material in the keychain.

Not calling it a blocker (first-party OpenAI credentials are high-entropy), but please do one of:

- **Minimum:** correct the "non-secret" wording in both docs so the claim matches reality.
- **Better:** make the on-disk digest useless on its own — e.g. HMAC-SHA256 keyed by a random per-account key kept *in the keyring*. That keeps your verification semantics intact while restoring the property that the disk file alone reveals nothing. It fits the subsystem's existing split cleanly.

## 3. Smaller items

- **MINOR — new wire constant uses the legacy fork prefix.** `KEYRING_JOURNAL_PREFIX = '__relay_chunk_journal__:v1:'` is a *new* constant (`main` only has `__relay_chunked__:`), and it's now baked into every marker payload on disk. Renaming to `__clodex_*` is free today and frozen forever after release.
- **MINOR — doc drift.** `docs/credential-helpers.md:77-79` still describes the pre-delta "deletes fail closed" behavior that this delta deliberately changed.
- **MINOR — C2 residual.** The adopt-failure fix is correct: the legacy value is computed before `adoptKeyringJournal` and returned regardless (`src/env.ts:1404-1430`). One wrinkle — a failed adopt leaves a `preparing` marker behind, so during a *sustained* journal-write outage only the first read succeeds and later ones return `null`. Self-healing and not a regression, but it undercuts what C2 was after.
- **MINOR — journal-less (`v1:managed\n`) state.** All three `persistKeyringManagedState` call sites (`:746`, `:751`, `:848`) pass a non-null journal, and this subsystem doesn't exist on `main` — so no released version ever wrote that format and it's only reachable on a machine that ran an earlier commit of this branch. Flagging only because one sub-case is a genuine permanent brick: journal-less marker plus one truly-missing chunk makes the new escape hatch return `false` forever (`src/env.ts:1705-1716`). Low priority given reachability.
- **NIT** — the `finally` in `deleteJournalLessManagedKeyringAccount` discards `deletePassword()`'s boolean; `probeProviderCredentialStore` reports the store healthy while an account is stuck in the state described in §1.

## On the sentinel mechanism — keep it

I considered asking you to replace it with something smaller, and decided against it. Writing throwaway entries to prove `findCredentials` is telling the truth is consistent with the read-back-verification philosophy the rest of the file already uses, and it's leak-safe: a stranded sentinel can't be returned as a credential, counted as a chunk, or used as an authRef, and the next delete reaps it. No change requested here.

## Round-1 test items — both addressed

- **T1 — done.** The `>128`-chunk guard now has behavioral coverage on both the write side (oversized → `false`, no partial state) and the read side (forged 129-chunk marker → `null`).
- **T2 — partially.** There's a genuine concurrent provision∥provision test through the real file lock, with contention actually verified. No delete∥write or replace∥replace pairing yet; optional.

One coverage gap worth closing alongside §1: the new-format recovery test only exercises a *short* credential, while real OAuth JSON is chunked — and no test drives `saveProviderCredential` (the replace path real re-login takes) from the recovered state, which is why the §1 gap survived a green suite.

---

Requesting changes for §1 and §2. Everything else is optional. The core design — journal-before-publish ordering, generation namespacing, reconcile-before-write, atomic marker writes, tombstone-before-delete, cross-config-home isolation — re-verified sound after the delta, and I confirmed the delta adds no hot-path cost on a healthy account.

camjac251:
The two major items are now addressed.

Direct provision and replacement can recover in one call when managed
filesystem state outlives a completely reset credential store. The recovery is
allowed only after sentinel-backed enumeration proves the main and chunk
namespaces are empty; retained main or chunk state, hidden inventory, and
incomplete sentinel cleanup remain fail-closed. The regression coverage
includes both a short credential and the chunked replacement path used by
reauthorization.

Managed recovery intent now uses AES-256-GCM with a random per-account key held
under the `clodex-state-key` service. The account, format version, and marker
mode are bound as authenticated data. The filesystem marker no longer exposes
the credential digest, and changed ciphertext, malformed key material, or a
missing key alongside retained credential state all fail closed.

The smaller items were folded in as well:

- the new journal prefix is `__clodex_chunk_journal__:v1:`
- metadata adoption no longer leaves a `preparing` marker during a sustained
  journal-write outage
- explicit journal-less deletion can retire a credential with a confirmed
  missing chunk after complete enumeration
- sentinel cleanup checks and reports a false deletion result
- documentation now describes encrypted recovery state and direct
  empty-store recovery

The branch is rebased onto `main` at `e272754`. Verification is green:

- focused keyring suite: 87 tests
- complete suite: 80 files, 990 tests
- type checking
- production build
- diff and secret checks
- focused security review of encryption, recovery, and key lifecycle

bman654 (changes requested):
Both round-2 items are genuinely resolved — I verified by producing and attacking the artifacts, not by reading the code. But this round's pass found a blocker that predates the round, plus two new ones introduced by the round-3 machinery.

First, a correction we owe you.

## Our round-2 crypto advice was the weaker option

We asked for "HMAC keyed by a random per-account key" as the better fix, with a doc correction as the minimum. That was wrong: HMAC provides integrity, not confidentiality, so `sha256(credential)` would still have been sitting on disk in the clear and the offline credential-confirmation oracle — the actual complaint — would have survived. And omitting the digest isn't available either: the marker exists so a crashed process can replay a journal write, and the restored journal's `shortDigest`/`digest` is what validates the credential afterwards, so a digest-less marker would have to fail closed and defeat its own purpose. **AES-GCM was the right call.** Don't undo it.

## Round-2 items: both fixed

**The digest is no longer recoverable from disk.** I provisioned through the real path and inspected the bytes: `v2:managed:8-miT04FyzNKlHfCoy6t08Alq…`, 239 bytes, decoding to 12-byte nonce ‖ 16-byte tag ‖ 142-byte ciphertext. `marker contains digest? false`, `marker contains secret? false`, and the round-2 attack (naive base64url decode) recovers nothing. The crypto itself audits clean: `randomBytes(32)` key, per-account and distinct across accounts; a fresh `randomBytes(12)` nonce per write (25 writes → 25 distinct nonces and ciphertexts); a full 16-byte tag actually applied via `setAuthTag` before `final()`; and the AAD binding is real — I confirmed with two positive attacks, rewriting the marker mode prefix and transplanting a marker plus its key to a second account, both of which fail closed. Docs are corrected; the surviving "non-secret" phrases refer to the `clodex-deleted` guard, whose value is literally `v1:deleted`, so they're accurate.

**C1 is fixed.** After a full credential-store reset with `~/.clodex` markers surviving, `provisionProviderCredential` and `saveProviderCredential` both recover in **one call** — verified for a short credential, for a 4000-char chunked credential, and for the `replace` intent that the real re-login path actually takes. All four fail-closed cases still refuse (retained main entry, retained chunks, hidden inventory, incomplete sentinel cleanup), and the risk I was most worried about is handled correctly: a locked or erroring keychain cannot be mistaken for an empty one, because both sentinels are written *and read back* before the emptiness read, and any throw lands in the catch with `empty` still false. I confirmed with every keyring call throwing — `provision -> false`, state file intact, keychain unchanged, and the original secret still resolving after "unlock".

Round-2 minors (c) and (d) are also addressed, and the round-2 test that pinned the old two-step contract was correctly renamed and inverted rather than deleted.

## BLOCKER — an `unverifiable` tombstone is a permanent brick, and it deletes the credential on the way in

`replaceMalformedKeyringJournalWithTombstone` writes `{mode:'delete', generations:[], unverifiable:true}` on any `InvalidKeyringJournalError`. `reconcileKeyringJournal` then hits `if (journal.unverifiable === true) return false;` at `src/env.ts:1189` — which sits **before** the `blockLegacy → 'deleted'` conversion at `:1191` — and `unverifiable` is only ever set, never cleared (written at 902/942/991/1591/2047, read at 925/954/978/1189, no clearing site). So reconcile returns false forever and every operation on the account fails permanently. Worse: the delete branch has already run `deleteKeyringEntry(KEYRING_SERVICE, account)` at `:1018` on the way in, so the live credential is destroyed en route to the brick.

Reproduced twice independently. Provision `'secret'`, set the journal entry to `'garbage'`:

```
replace   -> false
delete    -> false
delete    -> false
provision -> false
journal: __clodex_chunk_journal__:v1:{"mode":"delete","generations":[],"blockLegacy":true,"unverifiable":true}
entries remaining: ["clodex-deleted:<A>", "clodex-journal:<A>", "clodex-state-key:<A>"]   # main secret gone
```

Deleting the managed-state file does not help. The only escape is hand-deleting `clodex-journal:<account>` from the OS keychain. Two things make the end state worse than it first looks: re-authenticating doesn't recover, because `credentialInstanceAuthRef` derives the account suffix from `sha256` of the app home rather than randomly, so a fresh login lands on the same bricked account; and after the first attempt there is no `diag?.()` call on the `:1189` path at all, so the user sees only the bare "check access and try again".

I want to be precise about how likely this is, because it changes what I'm actually asking for. I went looking for ways a *normal* user reaches it and mostly came up empty:

- **Not** reachable by concurrency — `withKeyringAccountLock` wraps the whole read-reconcile-compare-write sequence and fails closed rather than proceeding lockless, and the lock root is the native passwd home, so it serializes even across differing `CLODEX_HOME`.
- **Not** reachable by truncation — `generations` is hard-bounded (`KEYRING_MAX_DELETE_GENERATIONS = 6`), worst-case journal is 1033 chars against the 1200-char guard, and the oversize guard throws a plain `Error` rather than `InvalidKeyringJournalError`, so it can't tombstone.
- **Not** reachable by interrupted writes — I audited all 15 journal write sites against the parser rules; every shape a single build can emit round-trips through its own parser, so a crash leaves a *stale but valid* journal.
- **Not** reachable from 2.1.1 today — v2.1.1 writes no journal at all, cold upgrade from both plain and legacy-chunked state adopts cleanly with byte-identical recovery, and every mixed 2.1.1/branch interleaving I tested fails safe with the credential preserved.

So today the realistic exposure is you and anyone else who ran an intermediate commit of this branch. **The reason I still think this blocks is forward compatibility.** `parseKeyringChunkJournal` tolerates only purely *additive* unknown fields. Everything else — a new `mode`, an existing flag newly permitted on another mode, a new required field — is treated as corruption, and the credential is destroyed:

| journal mutation | result |
|---|---|
| unknown extra field (short **and** chunked journals) | tolerated, credential intact |
| unknown `mode` value | credential **destroyed**, account bricked |
| `unverifiable:true` on `mode:'short'` | credential **destroyed** |
| `shortDigest` omitted from a `short` journal | credential **destroyed** |
| `blockLegacy:true` on `mode:'write'` | credential **destroyed**, chunk generations swept in |
| earlier in-branch prefix `__relay_chunk_journal__:v1:` | credential **destroyed** |
| **future prefix `__clodex_chunk_journal__:v2:`** | credential **destroyed** |

That last row is the one I'd most like changed. `src/env.ts:577` treats any non-matching prefix as `InvalidKeyringJournalError` — so `v1:` looks like a version gate but isn't one. A future `v2:` journal doesn't read as "written by a version I don't understand", it reads as corruption. **A version prefix whose only failure mode is destroying credentials is worse than no prefix at all.** And the shape demonstrably does change without a version bump: within this branch's own five commits the prefix was renamed and two modes plus four fields were added, all still labelled `v1`.

Right now that trap is disarmed, because nothing released writes a journal. It arms permanently the first time the journal shape changes again — and at that point it's in users' keychains and can't be fixed by a later release.

**In fairness: this is not something you introduced this round.** `unverifiable` and `replaceMalformedKeyringJournalWithTombstone` have been in the branch since its first commit (`ec2bc347`), unchanged through all five — our rounds 1 and 2 chased the recovery gates and never probed the tombstone path, so we missed it twice. It surfaced this round from a fresh review pass on the Opus 5 release, which went at the corruption paths harder than we had in previous rounds.

**And to be fair to your design: this is deliberate, not an oversight.** `tests/keyring-credentials.test.ts:1998` is named *"fails closed with a durable tombstone after a malformed journal"* and explicitly asserts the main entry is gone afterwards. So this isn't me telling you there's a bug you didn't notice — it's me disagreeing with the policy. Destroying a readable, self-contained credential because its *metadata* is unparseable is disproportionate, particularly since `unverifiable` conflates two different things: "the journal was malformed" (which says nothing about whether the published value is good) and its original meaning at `:942`/`:2034`, "the published value could not be parsed".

**Suggested fix, two parts:**

1. Make the tombstone retirable — after the generation cleanup loop succeeds, let an `unverifiable` delete journal finish by proving emptiness with the machinery already in the file (`verifyEmptyKeyringCredentialInventory`) and convert to `{mode:'deleted'}` when it passes, instead of the unconditional `return false` at `:1189`.
2. Give the journal a real forward-compatibility story: an unrecognized wire prefix, an unknown `mode`, or a shape this build doesn't understand should mean *"not mine — leave the credential alone"*, not *"corrupt — destroy it"*. Treating an unknown prefix the way legacy unmanaged entries are treated would do it. That's what keeps the next journal change from being a credential-destroying event for anyone running an older build.

A recovery path for the genuinely-corrupt case would also help — even a documented `clodex providers repair` that clears the journal, guard, state key and marker for one account, so the remedy isn't "hand-delete four keychain items you'd have to read the source to know about".

## MAJOR — malformed `clodex-state-key` material is never regenerated and never routed to recovery

A *missing* state key self-heals: it throws `KeyringManagedStateKeyUnavailableError`, `writeKeyringAccountLocked` routes that into `recoverEmptyOpaqueManagedKeyringState`, and a fresh key is minted. A *malformed* one does neither. `parseKeyringManagedStateKey` throws a plain `Error`, which the catch at `src/env.ts:1381-1384` collapses into `'keyring managed-state marker is invalid'` — not the `Unavailable` class — so the recovery gate never fires; and `ensureKeyringManagedStateKey` re-parses the same bad value and throws rather than replacing it.

Result: reads still work, but **every write, delete, and re-provision fails forever**. And because `credentialInstanceAuthRef` derives the account instance deterministically from `sha256(getAppHome())`, re-authenticating lands on the same wedged account. Observed with a `v1:CORRUPT` key: `re-provision -> false`, `resolve -> null`; and after deleting the marker file as a workaround, `resolve -> "sk-…"` but `save -> false` permanently.

This is also a forward-compat trap — the `v1:` prefix exists to allow rotation, but a future `v2:` key format would hard-brick every older build instead of degrading the way a missing key does.

**Suggested fix:** have `parseKeyringManagedStateKey`/`readKeyringManagedStateKey` throw `KeyringManagedStateKeyUnavailableError` for unparseable material so it takes the same path as a missing key, and have `ensureKeyringManagedStateKey` overwrite unparseable material with a freshly generated key. The latter is safe by construction — a key that cannot be parsed can decrypt nothing, so replacing it destroys no recoverable state.

Related, smaller: when the key is corrupt but the marker is intact, the surfaced diagnostic says `marker is invalid`, pointing the user at the wrong artifact.

## MAJOR — one transient sentinel-removal failure permanently blocks the new recovery path

`verifyEmptyKeyringCredentialInventory` treats any `::chunk::`-prefixed entry that isn't one of the two sentinels it just wrote as a live credential chunk. But `removeKeyringEnumerationSentinel` → `deleteKeyringEntry` writes a `__clodex_delete__:<uuid>` tombstone *over* the sentinel before calling `deletePassword()` — so if that throws, returns false, or the process is killed in the `finally`, an orphan tombstone is left in the `::chunk::` namespace. From then on the emptiness check can never return true and recovery is dead, on an otherwise **healthy** keychain:

```
provision #1 -> false
orphans left: ["clodex-chunks:<A>::chunk::5ca7d158-…::0", "clodex:<A>::chunk::5453b4ca-…::0"]
provision #2 -> false      # keychain fully healthy from here on
delete       -> false
provision #3 -> false
--- after manually rm ~/.clodex/keyring-state/<sha256>.managed ---
provision #4 -> true
```

That's the same "escape only by manually rm'ing an opaque hash file" that round 2 asked to eliminate, relocated into the recovery path itself.

**Suggested fix:** don't count entries whose value carries a reserved clodex prefix (`KEYRING_ENUMERATION_SENTINEL_PREFIX` / `KEYRING_DELETE_TOMBSTONE_PREFIX`) as credential chunks — try to delete them, and fail closed only if the delete fails. A reserved-prefix value is clodex bookkeeping by construction, and a delete tombstone already means "logically removed".

## MAJOR — the concurrency coverage doesn't hold

Two related problems, both mutation-verified:

- **T2 is still not done.** There is no delete∥write pairing. I removed the `withCredentialMutationLock` wrapper from `deleteProviderCredential` only, and the **entire suite stayed green — 80 files / 990 tests**. Since delete and the write paths share that lock, a lost lock there is exactly the "write publishes marker → concurrent delete removes chunks" credential-destruction scenario this PR exists to prevent.
- **`serializes concurrent public writes` (`tests/keyring-credentials.test.ts:349`) doesn't verify serialization.** It asserts both calls resolved true and the end state is self-consistent, never that they contended — and it structurally can't, because after the single `await import('@napi-rs/keyring')` the whole critical section is synchronous, so an in-process `Promise.all` cannot interleave it. Making `withCredentialMutationLock` a pass-through leaves all 87 focused tests green; the 4 failures that do appear under that mutation are all in *other* files. So the round-2 "verified contention" credit belongs to `credential-helper.test.ts:201`, not this suite.

**Suggested fix:** one gated delete∥write test in the style of `credential-helper.test.ts:201` — start a write, wait for its in-critical-section sentinel, launch `deleteProviderCredential`, assert no overlap, release, then assert the end state is one of the two legal outcomes and never a torn marker/chunk set. And either gate the fake keyring so the second writer demonstrably queues, or rename `serializes concurrent public writes` to what it actually checks.

## MAJOR — test isolation regression against the developer's real home

On `main`, `getCredentialMutationLockPath` returned a path inside `CLODEX_HOME`, fully contained by the suite's sandbox. This PR moves it to `userInfo().homedir + '/.clodex/credential-locks'`, and `os.userInfo().homedir` reads the passwd database and ignores `$HOME` — so it escapes both `CLODEX_HOME` **and** the `HOME` sandbox floor.

`tests/keyring-credentials.test.ts` correctly mocks `node:os` `userInfo`, so the new suite is clean. Four other suites don't, and `pnpm test` writes ~40 distinct lock digests into the real `~/.clodex/credential-locks` from 7 worker pids: `tests/oauth-credential-store.test.ts` (41 artifacts), `tests/credential-helper.test.ts` (11), `tests/credential-lifecycle-persistence.test.ts` (5), `tests/registry-lock.test.ts` (3). One of them hashes to a **production** authRef (`keyring:provider:openai`) — so running the suite while `clodex server` refreshes that credential contends on the identical lock file.

**Suggested fix:** hoist the `vi.mock('node:os', … userInfo)` into a shared helper (or `vitest.setup.ts`) used by every credential-touching suite, and extend `tests/test-sandbox-floor.test.ts` to assert `getCredentialLockRoot()`/`getCredentialStateRoot()` land inside the vitest temp sandbox — otherwise the floor silently stops covering the most dangerous state in the repo.

## Minors

- `deleteKeyringAccount` has no `recoverEmptyOpaqueManagedKeyringState` branch, so after a keychain reset `clodex providers remove` returns false forever and `reconcileCredentialDelete` keeps the entry queued in `credential-cleanup.json` — every later `clodex providers` command reports pending cleanup that can never complete. Deleting an already-empty namespace is unambiguously safe, so the write path's branch applies here too.
- When `verifyEmptyKeyringCredentialInventory` concludes "not empty" it emits no diagnostic, so a retained-chunk refusal surfaces only as `keyring managed-state encryption key is unavailable` — which is not the reason it failed and suggests no next step.
- `persistKeyringDeletionTombstone` `setPassword`s the tombstone over a live value *before* verifying it can delete it, so if a backend ever reported a present entry as absent from both `getPassword` and `findCredentials`, the secret would be destroyed while the operation reports failure. I could not demonstrate a backend that behaves that way, so this is theoretical — but the destructive ordering is real regardless. Requiring a successful pre-read (or skipping the tombstone when the entry reads absent) removes it.
- The AAD binding has no test — deleting both `setAAD` calls leaves 87/87 green. One test that rewrites a `v2:preparing:` marker's prefix to `v2:managed:` and asserts the read returns null would pin it.
- `expect(keyring.getCount).toBe(6)` (`:411`) pins an exact call count where the requirement is an upper bound; `toBeLessThanOrEqual(6)` still catches the O(n)-scan regression it guards.
- Testers on this branch should know: the `__relay_chunk_journal__:v1:` → `__clodex_chunk_journal__:v1:` rename landed in `97c766f` only, and the old prefix hard-fails `parseKeyringChunkJournal` — which trips the BLOCKER above, with credential loss. Anyone who ran `ec2bc34`..`c557101` should delete `clodex-journal:*` from their keychain before upgrading. No end users are affected since this is unreleased, but it's worth a line in the PR description.

Counts confirmed: 87 focused, 990 across 80 files, typecheck clean. The fake keyring is faithful to `@napi-rs/keyring`'s sync surface and models throw-on-read/enumerate/write, delete-returns-false, and the partial-visibility case — this is not a fake-only suite; essentially all the new `src/env.ts` code executes for real.

camjac251:
The latest review items are addressed in `fc84a45`, with current `main` integrated in `4769221`.

Unknown, malformed, older, and future journal formats are preserved as opaque state. Known unverifiable deletion state retires only after a complete sentinel-backed empty-inventory proof, and malformed or future managed-state keys regenerate only when the namespace is proven empty.

Sentinel cleanup now requires linked account and value identities and rechecks the exact stored value before deletion, so credential chunks that resemble bookkeeping entries remain untouched. Ambiguous deletion stays pending until absence is confirmed. Public write/delete contention, delete-side empty-state recovery, AAD binding, diagnostics, and native-home sandboxing are covered.

The PR description also includes the requested warning for unreleased entries using the earlier journal prefix. Current verification is green: 101 focused keyring tests, 1,011 tests across 80 files, type checking, production build, secret scan, and both GitHub checks. The current head is mergeable against `main`.

bman654 (changes requested):
Round 4 at `4769221`. Merges clean, typecheck clean, build clean, **1011/1011 tests green on macOS**. The round-3 blocker and three of the four MAJORs are genuinely fixed — I verified each by running the same probe against `97c766f` and against this head, so the comparisons below are measured, not inferred. **One MAJOR is unchanged**, and it is the same forward-compatibility trap you just closed for journals — details at the end.

## Fixed and verified

**BLOCKER — `unverifiable` tombstone: resolved.** The flag is now retirable (`reconcileKeyringJournal` proves an empty inventory, rewrites a plain `delete`, and falls through to the `blockLegacy` conversion), and the two destructive `replaceMalformedKeyringJournalWithTombstone` call sites are gone.

The forward-compat half is resolved too. I drove five journal shapes — future `v2:` prefix, unknown `mode`, unknown extra field, missing required field, and outright garbage — against a healthy credential:

| | `97c766f` | this head |
|---|---|---|
| secret survives read + failed replace | yes | yes |
| unknown journal preserved verbatim | **no — clobbered with a tombstone** (4 of 5 shapes) | **yes, all 5** |
| account recoverable afterward | no | yes, see below |

Every shape now fails closed with a diagnostic and leaves both the credential and the journal untouched. Deleting the single `clodex-journal:<account>` entry restores the account completely (I confirmed a subsequent provision and read both succeed). That is the behavior I asked for.

**MAJOR — orphan sentinel: resolved.** Forcing every sentinel removal to fail during empty-store recovery, then letting the keychain go healthy again:

- `97c766f`: second attempt `false`, read `null` — permanently blocked, exactly as reported.
- this head: second attempt `true`, read returns the secret, diagnostic `recovered managed credential metadata after verifying an empty keyring`.

**MAJOR — T2 delete∥write: resolved.** Removing `withCredentialMutationLock` from `deleteProviderCredential` now reddens a test. In round 3 that mutation left all 990 green.

**MAJOR — test isolation: resolved, and I checked the real home.** The `node:os` `userInfo` mock in `vitest.setup.ts` keeps the native-home lock root inside the per-file sandbox. Running 147 credential tests left `~/.clodex/credential-locks` with its mtime unchanged and never created `~/.clodex/keyring-state`.

**Minors:** AAD binding is now covered — deleting both `setAAD` calls reddens a test (round 3: 87/87 stayed green). `deleteKeyringAccount` has the recovery branch. `deleteKeyringEntry` reads before it tombstones and verifies after deleting, so the destructive-ordering note is addressed.

## Still open

**MAJOR — a malformed or future-format `clodex-state-key` is an unrecoverable dead end.** This one is byte-identical between `97c766f` and this head. With credential entries still present and the state key unreadable:

- `resolveProviderCredential` → `null`, `deleteProviderCredential` → `false`, `provisionProviderCredential` → `false`. All three in-product paths fail, so the user cannot even delete the credential and start over.
- The credential itself is preserved — nothing is destroyed, which is why this is a MAJOR and not a blocker.
- Removing the `clodex-state-key:<account>` keychain entry **is not enough**; it still fails. Recovery also requires deleting `~/.clodex/keyring-state/<sha256(account)>.managed`. Once both are gone, the original secret reads back fine and provisioning works.
- The only diagnostic is `keyring managed-state encryption key is unavailable`, which names neither the keychain entry nor the file.

The reason I am still holding on this: it is the same trap you just fixed for journals. `parseKeyringManagedStateKey` accepts only `v1:<32-byte base64url>` and treats everything else as unavailable, so a future clodex that writes a `v2:` state key bricks the account for any older binary. I verified this specifically — same 32 bytes of key material, only the version tag changed to `v2:` — and got the identical dead end (`read null`, `del false`, `prov false`, credential preserved).

Two ways out, either is fine by me:

1. Apply the journal rule to the state key: an unrecognized key format means "not mine" — leave the credential alone but let `deleteProviderCredential` succeed, so re-auth is a working escape hatch.
2. Keep failing closed, but make the diagnostic name both artifacts to remove. A dead end the user can actually get out of is a very different thing from one they cannot.

Given how much of this is now landed and verified, I would happily approve on either.

**Minor — the sandbox-floor guard no longer detects what it was built to detect.** Because `userInfo` is mocked suite-wide, any path derived from the native home now resolves inside the sandbox and passes `expectInsideVitestSandbox` by construction. The mock genuinely fixes the leak — I confirmed that above — but the assertion can no longer fail if state is relocated to the real home in future. Capturing the unmocked `homedir` once and asserting the path is outside it would restore the tripwire without giving up the sandbox.

camjac251:
The remaining managed state-key recovery item is addressed in `adcd570`.

- Reads and provisioning still fail closed and preserve the credential, chunks, opaque key, and encrypted managed-state file when current-format key material is malformed or the key format is from a future version.
- Explicit deletion can now retire unreadable managed metadata after a known cleanup journal has been parsed.
- If the cleanup journal is absent, deletion uses sentinel-backed enumeration to verify the complete credential and chunk inventory before removing anything.
- Malformed, unknown, and future cleanup journals are still preserved verbatim because journal parsing happens before managed-state retirement.
- A retained deletion guard lets an interrupted journal-less deletion reconstruct its journal and finish on retry.
- The sandbox-floor test now compares coordination paths with the unmocked native home, restoring the real-home regression tripwire.

Fresh verification on the pushed commit:

- type checking passed
- production build passed
- 1,015 tests passed across 80 files
- diff checks passed
- repository secret scan found no leaks

bman654 (approved):
Round 5 at `adcd570` — **approving.** The last MAJOR is fixed, and I verified it the same way as the previous rounds rather than reading the summary: 1015/1015 on macOS, typecheck clean, build clean, CI green.

**MAJOR — unreadable `clodex-state-key` now has a real escape hatch.** Probed directly against this head:

| | round 4 (`4769221`) | this head |
|---|---|---|
| `resolveProviderCredential` | `null` (credential preserved) | `null` (credential preserved) |
| `deleteProviderCredential` | **`false`** | **`true`** |
| re-provision + read back | `false` / `null` | `true` / returns the new secret |

Same result for a future-format key (`v2:` with identical key material), which was the specific forward-compat case I was worried about. I also ran it against a **4-chunk credential**: delete succeeds, re-provision and read work, and **no orphaned `::chunk::` entries are left behind**. The diagnostic `retired unreadable managed credential metadata for explicit deletion` says exactly what happened. Reads and provisioning still fail closed with the credential, chunks and encrypted state file intact, which is the right split — destructive recovery only on an explicit delete.

**No regression in the journal forward-compat work.** Re-ran all four unknown-journal shapes (future `v2:` prefix, unknown `mode`, missing field, garbage) on this head: the secret and the journal are still preserved verbatim in every case, and deletion still refuses. Your ordering note is the reason it holds — journal parsing runs before managed-state retirement, so an unknown journal is never the thing that authorizes a retirement. Worth calling out that this leaves the unknown-journal case still needing manual removal of the one `clodex-journal:<account>` entry; that is the behavior I asked for in round 3 and I am explicitly not reopening it — refusing to destroy state a newer version may own is the correct call.

**Minor — sandbox-floor tripwire restored.** Comparing against `vi.importActual('node:os')` is the right fix, and asserting `realUserHome !== userInfo().homedir` pins that the mock is actually in effect. I confirmed it bites: deleting the `vi.mock('node:os')` block from `vitest.setup.ts` reddens the test.

**The tests hold the fixes.** Four mutations, each reddens at least one test: replacing the new `deleteJournalLessManagedKeyringAccount(..., true)` call with `return false`; removing `withCredentialMutationLock` from `deleteProviderCredential`; deleting both `setAAD` calls; and removing the `node:os` mock. I also re-confirmed the suite leaves the real home alone — a full 1015-test run left `~/.clodex/credential-locks` mtime unchanged and never created `~/.clodex/keyring-state`.

Merged with #44 on an integration branch: zero conflicts, **1023/1023**, typecheck and build clean.

This was a long one — five rounds, and every round the asks got harder and narrower. Thanks for staying with it, and in particular for pushing back where you were right: the AES-GCM call in round 3 was yours and it was correct, and the retirable-tombstone design you landed is better than the `providers repair` command I floated. Nice work.

PR #21, opened 2026-07-22
fix(auth): enforce anonymous credential boundaries

camjac251:
## Summary

- represent explicitly anonymous providers without sentinel credentials throughout catalog resolution and routing
- enforce one credential-header classifier across SDK forwarding, native forwarding, and diagnostic redaction
- remove authorization, API-key, cookie, token, secret, and credential-bearing headers from anonymous dispatch while preserving ordinary metadata
- intentionally retain configured provider headers on authenticated API-key and OAuth proxy routes
- fail credential activation before registry updates when secure persistence fails
- clarify automatic dead-PID lock reclamation and the narrow manual-removal case

## Boundary guarantees

- process-level credential defaults cannot attach to anonymous upstream requests
- configured credential-bearing headers cannot cross an anonymous dispatch boundary
- URL-plus-init request headers are filtered for each supported provider family
- single-route messages, token counting, and alternate chat forwarding preserve the anonymous boundary on the wire
- diagnostic redaction and outbound filtering cannot drift to different header-name rules

## Test plan

- [x] focused header, provider, proxy, and route suite (104 tests across 6 files)
- [x] full Node 24 container suite (717 tests across 70 files)
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] `git diff --cached --check`
- [x] staged secret and personal-identifier scans

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,10 @@ clodex --version    # version
   FAT, exFAT, or a network mount that rejects hard links. An abrupt process kill
   during lock publication can leave a `providers.json.lock.*.tmp` file; it does
   not block later lock acquisition and can be removed when no Clodex process is
-  running.
+  running. If `providers.json.lock` remains after a crash, stop every Clodex
+  process and verify that the recorded PID is not a running Clodex process before
+  removing the lock. Never remove the canonical lock while a Clodex process is
+  active.
 - Credentials live in the OS credential store (Keychain / Windows Credential Manager / Secret Service) under the `clodex` service. Set `CLODEX_CREDENTIAL_HELPER` to an absolute executable path to use an external secure store instead; see [credential helpers](docs/credential-helpers.md).
 - `CLODEX_CLAUDE_PATH` overrides Claude Code binary discovery.
 - **Outbound proxy:** when `HTTP_PROXY`/`HTTPS_PROXY` (and optionally `NO_PROXY`) are set in clodex's environment, all clodex-originated network calls honor them — OAuth sign-in and token refresh, model-list and models.dev refreshes, upstream OpenAI API calls, and the ChatGPT/Codex OAuth WebSocket transport (tunneled via HTTP CONNECT).
--- a/README.md
+++ b/README.md
@@ -219,11 +219,13 @@ clodex --version    # version
   FAT, exFAT, or a network mount that rejects hard links. An abrupt process kill
   during lock publication can leave a `providers.json.lock.*.tmp` file; it does
   not block later lock acquisition and can be removed when no Clodex process is
-  running. If `providers.json.lock` remains after a crash, stop every Clodex
-  process and verify that the recorded PID is not a running Clodex process before
-  removing the lock. Never remove the canonical lock while a Clodex process is
-  active.
+  running. A canonical `providers.json.lock` whose recorded PID is no longer
+  running is reclaimed automatically on the next lock acquisition. If it remains
+  while that PID is active, stop every Clodex process and verify the recorded PID
+  before removing the lock manually. Never remove the canonical lock while a
+  Clodex process is active.
 - Credentials live in the OS credential store (Keychain / Windows Credential Manager / Secret Service) under the `clodex` service. Set `CLODEX_CREDENTIAL_HELPER` to an absolute executable path to use an external secure store instead; see [credential helpers](docs/credential-helpers.md).
+- Proxied routes forward configured provider headers for API-key and OAuth authentication. Anonymous routes preserve non-credential headers while removing authorization, API-key, cookie, token, secret, and credential-bearing header names before dispatch.
 - `CLODEX_CLAUDE_PATH` overrides Claude Code binary discovery.
 - **Outbound proxy:** when `HTTP_PROXY`/`HTTPS_PROXY` (and optionally `NO_PROXY`) are set in clodex's environment, all clodex-originated network calls honor them — OAuth sign-in and token refresh, model-list and models.dev refreshes, upstream OpenAI API calls, and the ChatGPT/Codex OAuth WebSocket transport (tunneled via HTTP CONNECT).
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1264,6 +1264,9 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
   const usesAnthropicProxy = selectedModel.modelFormat === 'anthropic' &&
     (isOAuthAnthropic || anonymousProvider);
 
+  // Static provider headers remain part of the proxied endpoint contract,
+  // including OAuth routes. Anonymous dispatch filters credential-bearing
+  // names at the final boundary while retaining non-credential metadata.
   if (usesAnthropicProxy) {
     // The passthrough proxy owns upstream authentication for OAuth and strips it
     // entirely for explicitly anonymous Anthropic endpoints.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1249,7 +1249,8 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
   }
 
   const launchApiKey = await resolveLocalProviderApiKey(activeProvider);
-  if (!launchApiKey?.trim()) {
+  const anonymousProvider = activeProvider.authType === 'none';
+  if (!anonymousProvider && !launchApiKey?.trim()) {
     p.log.error(
       `No credential found for ${activeProvider.name}. Add a key or sign in with clodex providers.`,
     );
@@ -1260,9 +1261,12 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
   let childEnv: NodeJS.ProcessEnv;
 
   const isOAuthAnthropic = selectedModel.modelFormat === 'anthropic' && activeProvider.authType === 'oauth';
+  const usesAnthropicProxy = selectedModel.modelFormat === 'anthropic' &&
+    (isOAuthAnthropic || anonymousProvider);
 
-  if (isOAuthAnthropic) {
-    // Anthropic OAuth passthrough — proxy injects compatibility metadata and Bearer auth.
+  if (usesAnthropicProxy) {
+    // The passthrough proxy owns upstream authentication for OAuth and strips it
+    // entirely for explicitly anonymous Anthropic endpoints.
     try {
       proxyHandle = await startProxy(
         selectedModel.baseUrl ?? 'https://api.anthropic.com',
@@ -1271,16 +1275,17 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
         selectedModel.contextWindow,
         {
           providerId: activeProvider.id,
-          authType: 'oauth',
+          authType: activeProvider.authType,
           oauthAccountId: activeProvider.oauthAccountId,
           providerData: activeProvider.providerData,
           modelFormat: 'anthropic',
+          headers: activeProvider.headers,
         },
-        launchApiKey,
+        launchApiKey ?? '',
       );
-      if (!isAgentStdoutMode()) p.log.info(`OAuth proxy started on port ${proxyHandle.port}`);
+      if (!isAgentStdoutMode()) p.log.info(`Anthropic proxy started on port ${proxyHandle.port}`);
     } catch (err) {
-      p.log.error(`Failed to start OAuth proxy: ${err instanceof Error ? err.message : String(err)}`);
+      p.log.error(`Failed to start Anthropic proxy: ${err instanceof Error ? err.message : String(err)}`);
       return 1;
     }
     childEnv = buildChildEnv(
@@ -1294,7 +1299,7 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
     childEnv = buildChildEnv(
       selectedModel.baseUrl!,
       selectedModel.id,
-      launchApiKey,
+      launchApiKey ?? '',
       undefined,
       selectedModel.contextWindow,
     );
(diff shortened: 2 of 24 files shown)


src/provider-factory.ts
```diff
@@ -23,14 +23,29 @@ const RESPONSES_ONLY_PREFIXES = [
...
 };
 
 const factoryCache = new Map<string, Promise<SdkProviderFactory>>();
 
+const fetchWithoutCredentialHeaders: typeof fetch = (input, init) => {
+  const headers = new Headers(
+    init?.headers ?? (input instanceof Request ? input.headers : undefined),
+  );
+  headers.delete('authorization');
+  headers.delete('x-api-key');
```
bman654: **major (security):** This denylist covers only `authorization`/`x-api-key`, but the wrapper's job is to enforce a *no-credential* boundary, and configured provider headers are arbitrary user-persisted values (`src/registry/custom-endpoint.ts`) that get merged into anonymous SDK requests below (and into native Anthropic requests via `anthropicUpstreamHeaders`). A stale `Cookie: session=…`, `Proxy-Authorization:`, `X-Auth-Token:`, or similar on an anonymous provider still reaches the upstream.

The codebase already classifies the broader set of credential-bearing header names for trace redaction (`src/trace-log.ts` — cookie/token/secret/credential patterns). Suggest extracting that classification into a shared case-insensitive predicate and using it both here and in the `authType === 'none'` filter in `src/upstream-forward.ts`, while preserving genuinely non-secret headers (content negotiation, plan selection, etc.).

src/upstream-forward.ts
```diff
@@ -7,18 +7,30 @@ export function anthropicUpstreamHeaders(
...
   extraHeaders?: Record<string, string>,
 ): Record<string, string> {
   const key = sanitizeCredential(apiKey) ?? apiKey.trim();
-  const isOAuth = authType === 'oauth';
+  const resolvedAuthType = authType ?? 'api';
+  const isOAuth = resolvedAuthType === 'oauth';
+  const forwardedExtraHeaders = resolvedAuthType === 'none'
+    ? Object.fromEntries(
+        Object.entries(extraHeaders ?? {}).filter(
+          ([name]) => !['authorization', 'x-api-key'].includes(name.toLowerCase()),
```
bman654: Same two-name denylist as `fetchWithoutCredentialHeaders` in `src/provider-factory.ts` — see the comment there: for `authType === 'none'` this should filter the broader credential-bearing header set (cookie/proxy-authorization/token/secret variants), ideally via one shared predicate so the two sites can't drift.

tests/provider-factory.test.ts
```diff
@@ -363,17 +385,62 @@ describe('createLanguageModel', () => {
...
 
     expect(createOpenAICompatible).toHaveBeenCalledWith({
       name: 'kilo',
       baseURL: 'https://api.kilo.ai/api/gateway',
+      fetch: expect.any(Function),
     });
     vi.doUnmock('@ai-sdk/openai-compatible');
   });
 
+  it('strips generated credential headers for anonymous Anthropic providers', async () => {
```
bman654, line 494: **major (tests):** This is the only test that actually invokes the captured fetch wrapper, and it passes credentials via a `Request` object with no `init` — so the `init?.headers` branch of `fetchWithoutCredentialHeaders` is never exercised. That's the call form the AI SDKs actually use (`fetch(url, { headers })`); a mutant that strips only `Request.headers` while forwarding credential-bearing `init.headers` passes every test in this PR.

Two smaller points in the same area:

- The OpenAI and openai-compatible cases (lines ~262 and ~396) only assert `fetch: expect.any(Function)` — they'd pass with any function, including one that strips nothing. Invoking each captured wrapper with URL-plus-`init` headers and asserting the credential headers are gone (and ordinary headers survive) would make these behavioral.
- `vi.stubGlobal('fetch', transport)` at line 424 is only undone after all assertions — a failure mid-test leaks the stub into subsequent tests. Wrap in `try/finally` (the analogous test in `tests/proxy.test.ts` does this correctly).

tests/proxy.test.ts
```diff
@@ -107,6 +107,51 @@ describe('SDK anonymous route handling', () => {
     expect(res.status).toBe(502);
     expect(res.body).not.toContain('Missing API key');
   });
+
+  it('forwards anonymous Anthropic routes without authentication headers', async () => {
```
bman654, line 115: **tests:** Good wire-level test — but it covers only `startProxyCatalog` + `/v1/messages`. Two changed anonymous branches remain untested:

- **Single-route `startProxy`** (`src/proxy.ts` bottom) — the path `clodex claude` actually launches anonymous Anthropic providers through, and the code that newly forwards `authType`/`headers` into the route. A propagation regression there would pass this suite.
- **The anonymous `count_tokens` gate** (`src/proxy.ts` ~379-407) — its 401 short-circuit got its own `routeAuthType !== 'none'` exemption, separately from the messages path.

A variant of this test through `startProxy`, plus one `count_tokens` request, would pin both. Supplying a configured `Authorization`/`X-Custom` header pair on the route here would also verify the extra-headers filtering end-to-end rather than only in the `anthropicUpstreamHeaders` unit test.

src/server/router.ts
```diff
@@ -481,6 +482,7 @@ async function handleOpenAIChatCompletions(
       requestPreview: getLatestMessagePreview(body.messages, body.system),
     });
     await relayAnthropicMessages(res, completionsUrl, forwardBody, apiKey, Boolean(body.stream), {
+      authType: model.authType ?? 'api',
```
bman654, line 485: **tests:** This authType plumbing has no anonymous wire test — the new listener test covers only `/anthropic/v1/messages`. An `authType: 'none'` model with a `completionsUrl` posted to `/openai/v1/chat/completions`, asserting the upstream listener receives neither `authorization` nor `x-api-key`, would cover this branch.

src/cli.ts
```diff
@@ -1271,16 +1275,17 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
         selectedModel.contextWindow,
         {
           providerId: activeProvider.id,
-          authType: 'oauth',
+          authType: activeProvider.authType,
           oauthAccountId: activeProvider.oauthAccountId,
           providerData: activeProvider.providerData,
           modelFormat: 'anthropic',
+          headers: activeProvider.headers,
```
bman654, line 1285: **minor (undeclared behavior change):** This line changes behavior for **OAuth** launches too, not just anonymous ones — previously the single-route `startProxy` path dropped configured provider headers entirely; now they're forwarded (and `src/proxy.ts` threads `sdk?.headers` into the route). For anonymous providers it's required; for OAuth providers it's a new side effect that looks like a fix but isn't mentioned in the commit body or PR description. If intentional, please say so in the commit body; if not, gate it on `authType === 'none'`.

README.md
```diff
@@ -219,7 +219,10 @@ clodex --version    # version
   FAT, exFAT, or a network mount that rejects hard links. An abrupt process kill
   during lock publication can leave a `providers.json.lock.*.tmp` file; it does
   not block later lock acquisition and can be removed when no Clodex process is
-  running.
+  running. If `providers.json.lock` remains after a crash, stop every Clodex
```
bman654: **minor (docs, scope):** Two things: (1) this paragraph is unrelated to the PR's stated purpose — fine to keep, but worth a mention in the commit body; (2) it implies manual cleanup is required, but the lock implementation auto-reclaims locks whose recorded PID is dead (`src/registry/lock.ts` — stale detection + `removeStaleLock`), so a crash-left `providers.json.lock` clears itself on the next acquisition. Suggest adding a clause noting dead-PID locks are reclaimed automatically, keeping the manual steps only for the pathological live-PID-collision case.

bman654 (changes requested):
Thanks for the follow-up — this is a well-shaped continuation of #8. Four-dimension review (security, regression sweep, test quality, architecture/conventions); full suite runs green on this branch in my environment (69 files / 701 tests, including the listener-backed `proxy.test.ts` and `server-router.test.ts` suites you couldn't run locally), and `pnpm typecheck` passes.

**What's solid**

- The sentinel removal is the right shape for this codebase: it completes the `authType` union #8 already threaded through the route/spec types, and mirrors the existing per-site `authType === 'oauth'` branching instead of inventing a new abstraction. Crucially the guards fail **closed** — a site that misses the `!== 'none'` exemption drops a route or 401s; it never leaks a credential.
- A full regression sweep of the old `'anonymous'` sentinel's consumers came back clean: every `resolveLocalProviderApiKey` caller handles `''`, `authType` is populated at every production route/spec construction site, template/custom-endpoint blank-credential providers are normalized to `authType: 'none'` at persist/materialize time, and `env:` refs are untouched.
- The warn→hard-fail change on OAuth credential persistence is coherent with #8's upfront store probe and the sibling refresh-persistence path (which already threw), and it closes a real broken-state generator (registry pointing at a credential that never saved). The throw happens before the registry upsert and the lock releases correctly on throw.
- The `docs/credential-helpers.md` rewording is accurate against actual behavior, and no existing test was weakened — the replaced graceful-degradation test's successor asserts strictly more.

**Why request-changes**

1. **(major)** The anonymous credential boundary strips only `Authorization`/`x-api-key`, but configured provider headers are arbitrary persisted values — a stale `Cookie:`, `Proxy-Authorization:`, or `X-Auth-Token:` on an anonymous provider still reaches the upstream. Inline comment with details on `src/provider-factory.ts`.
2. **(major, tests)** The fetch-wrapper tests never exercise the `init.headers` path — the call form the SDKs actually use — so a wrapper that only stripped `Request.headers` would pass every test. A few other changed branches (single-route `startProxy`, anonymous `count_tokens`, anonymous `/openai/v1/chat/completions`) have no wire-level coverage. Inline comments.
3. **(minor)** One undeclared behavior change (OAuth launches now forward configured provider headers that were previously dropped) that looks intentional but deserves a sentence in the commit body, plus a README nuance. Inline comments.

None of this is structural — the design can stay exactly as is.

src/provider-factory.ts
camjac251: Addressed in `16952a8`. `isCredentialBearingHeader()` is now the shared case-insensitive classifier used by anonymous SDK forwarding, native forwarding, and diagnostic redaction. It removes authorization, API-key, cookie, token, secret, and credential segments while retaining ordinary content and provider-metadata headers.

src/upstream-forward.ts
camjac251: Addressed in `16952a8`. Native anonymous forwarding now uses the same shared classifier as SDK forwarding and trace redaction, so the two outbound boundaries cannot drift to different credential-header rules.

tests/provider-factory.test.ts
camjac251: Addressed in `16952a8`. The tests invoke each captured provider fetch wrapper with the URL-plus-`init.headers` form and verify the broad credential set is removed while content and custom headers survive. The helper restores the global transport in `finally`, and the behavior is exercised for all three provider factories.

tests/proxy.test.ts
camjac251: Addressed in `16952a8`. A real single-route listener test covers both messages and token counting with anonymous authentication, configured credential-bearing headers, and a preserved custom header. Both upstream calls arrive without authorization, API-key, cookie, or token headers.

src/server/router.ts
camjac251: Addressed in `16952a8`. A real listener test sends an anonymous request through the alternate chat-completions endpoint and verifies the upstream receives neither authorization nor API-key headers.

src/cli.ts
camjac251: The authenticated-header behavior is intentional and is declared in `16952a8` and the updated PR description. Configured provider headers remain available on API-key and OAuth proxy routes; only explicitly anonymous dispatch applies the credential-bearing filter. Tests pin both sides of that boundary.

README.md
camjac251: Addressed in `16952a8`. The lock guidance now states that a canonical lock with a dead recorded PID is reclaimed automatically on the next acquisition. Manual removal is limited to the pathological case where the lock remains while the recorded PID is active, and the commit body declares this documentation correction.

bman654 (approved):
Re-reviewed at 49dfcfc against our previous review. Every finding is addressed — this is now a clean, well-tested boundary. Verified locally: `pnpm typecheck` and the full suite green (70 files / 718 tests), including the new listener-backed anonymous wire tests.

What we verified:

- `isCredentialBearingHeader()` in `src/credential-headers.ts` is genuinely one shared, segment-based classifier used at all three boundaries (anonymous SDK fetch wrapper, native forwarding, trace redaction). Segment matching is correct: strips `Cookie`/`Proxy-Authorization`/`X-Auth-Token`/`X-Client-Secret`, leaves `X-Goog-Api-Client` and ordinary content headers alone.
- The fetch-wrapper tests now exercise the `init.headers` form through all three provider factories with transport restored in `finally`.
- Real single-route listener coverage for anonymous `/v1/messages` + `count_tokens`, and the `/openai/v1/chat/completions` route — upstream observed receiving no credential headers while `X-Plan`/`X-Custom` survive.
- The authenticated/anonymous boundary keys off `authType`, not empty-key coincidence, and is pinned by tests on both sides.
- 49dfcfc closes a real gap (some routes previously dropped configured headers wholesale, which could make stripping tests pass vacuously) and is declared in the PR description.
- README lock-reclaim wording now matches the code.

Non-blocking nits (no action needed to merge): the PR description's test count is stale by one (head runs 718); `tests/credential-headers.test.ts` could optionally pin the illustrative negatives `X-Goog-Api-Client` / `X-Tokenizer`.

Heads-up for your other branches: once this merges, #15 and #16 will need rebases (both conflict with it pairwise — GitHub's mergeable flag only tests against main). For the #15 overlap our read is that #15's lifecycle implementation should win the conflicting hunks (its `persistOAuthProvider` hard-fail and journal `isStoredCredentialRef` supersede the equivalents here).

PR #23, opened 2026-07-23
fix(server): wait for listener readiness

camjac251:
## Summary

- wait until bound TCP listeners accept connections before returning server handles
- normalize wildcard and IPv6 listener addresses when forming local URLs
- remove temporary listeners on startup failure and accept both common connection-failure codes in the transport regression

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run --maxWorkers=1 --no-file-parallelism` (70 files, 699 tests)
- [x] `pnpm build`
- [x] `pnpm pack --dry-run`
- [x] staged secret scan

### DIFF
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -1,7 +1,7 @@
 import * as http from 'node:http';
 import * as https from 'node:https';
 import * as net from 'node:net';
-import type { Socket } from 'node:net';
+import type { AddressInfo, Socket } from 'node:net';
 import { randomUUID } from 'node:crypto';
 import { URL } from 'node:url';
 import { createBrotliDecompress, createGunzip, createInflate } from 'node:zlib';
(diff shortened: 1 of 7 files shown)


bman654 (approved):
Reviewed at 5bce7e5, including as a production startup change rather than test infra. The readiness design holds up: successful-listen callback first, then a condition-based TCP connect probe against the *actual* bound address (5 ms retry spacing, 50 ms per-connect cap, 1 s deadline), so normal startup pays one loopback handshake and no fixed sleep. Bind failures surface immediately with the original error (`EADDRINUSE` doesn't wait out the readiness deadline), the bound server is closed on readiness timeout, wildcard binds probe via loopback with IPv6 hosts bracketed, and `startServer()` now returns a usable URL built from the bound address. Critically, `server-runtime.json` registration still happens only after the awaited successful start — a bind or readiness failure can't publish a discovery record. Verified locally: typecheck green, full suite 70 files / 699 tests (an initial cross-worktree flake on our side was environmental, clean on isolated rerun). `fix(server)` is the right type.

Heads-up: #10 has a one-line import conflict with this in `src/http-proxy/server.ts` — mechanical to resolve when it rebases.

PR #24, opened 2026-07-23
test(proxy): wait for complete usage logs

camjac251:
## Summary

- wait for completion plus both streamed usage records before asserting the request log
- remove the ordering assumption between independent asynchronous lifecycle paths

## Verification

- `pnpm typecheck`
- `pnpm build`
- focused proxy test composed with #23: 1 passed
- full suite composed with #23: 70 files passed, 699 tests passed

## Note

On unstacked `main`, the focused test can hit the independent listener startup race addressed by #23 before reaching these assertions.

### DIFF
--- a/tests/http-proxy-server.test.ts
+++ b/tests/http-proxy-server.test.ts
@@ -178,14 +178,18 @@ describe('selective HTTP proxy', () => {
       expect(receivedPath).toBe('/v1/messages?beta=true');
       expect(receivedAuth).toBe('Bearer subscription-oauth-token');
       expect(receivedBody.equals(body)).toBe(true);
-      // The proxy flushes response_usage/response_completed log lines as it drains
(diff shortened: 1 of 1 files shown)


bman654 (approved):
Reviewed at b3f3c4e. The wait is genuinely condition-based, not a sleep in disguise: it polls only until the three records the assertions require (`response_completed`, both `response_usage` shapes) all exist, exits immediately when they do, and on the 5 s deadline the existing detailed assertions fail with the missing record rather than silently passing. The old wait-for-`response_completed`-only approach encoded an ordering assumption between two independent async write paths — this fixes the actual race. The log path is unique per test so cross-talk can't satisfy the predicate. Verified locally: full suite green (69 files / 693 tests) plus three isolated reruns of the touched file, all 11/11. `test(proxy)` is the right non-release type.

PR #25, opened 2026-07-23
test: remove live provider probe

camjac251:
## Summary

- remove a live authenticated provider probe from default test discovery
- keep ordinary test runs isolated from local credential state and external services

## Validation

- type checking passes
- the full suite reaches 649 passing tests
- the remaining 43 listener-readiness failures reproduce on the base branch and are addressed by #23


bman654 (approved):
Reviewed at d3224bf. Confirmed the deleted file was exactly what it looked like: it loaded the user's real provider registry, resolved a stored OAuth credential, and made an authenticated network call to a live provider endpoint — with zero assertions, logging and passing on every failure mode. That's a live-network probe with no regression signal, in direct violation of the repo's test rules, and deletion (rather than skip-guarding) is the right call since a skip would preserve the non-signal. No remaining references to the file or its helper anywhere in the repo; offline OAuth/model-refresh coverage elsewhere is untouched; test count changed by exactly the expected one. Verified locally: full suite green (68 files / 692 tests) with no network egress. `test:` is the right non-release type.

Thanks for catching and removing this — it takes a live authenticated call out of every CI/test run.

PR #26, opened 2026-07-23
feat(logging): correlate response lifecycles

camjac251:
## Summary

- carry the validated session identifier onto response lifecycle records
- classify abnormal response endings as `downstream_client`, `local_shutdown`, or `upstream_failure`
- mark local shutdown before tracked sockets are destroyed so records keep the correct source
- cover pass-through usage, translated completion, upstream failures, downstream disconnects, local shutdown, and invalid identifiers

## Privacy

Only canonical UUID-shaped session identifiers are retained. Invalid values are omitted, and no request or response content is added to lifecycle records.

## Validation

- type checking passes
- focused lifecycle logging tests pass
- complete suite: 79 files and 878 tests
- production build passes
- diff hygiene and secret scanning pass

### DIFF
--- a/src/trace-log.ts
+++ b/src/trace-log.ts
@@ -231,6 +231,7 @@ export type InferenceResponsePhase =
 export interface InferenceResponseLifecycleLogEntry {
   event: InferenceResponseLifecycleEvent;
   requestId: string;
+  claudeSessionId?: string;
   modelId: string;
   provider: string;
   route: 'passthrough' | 'translated';
@@ -254,6 +255,7 @@ export interface InferenceResponseLifecycleLogEntry {
   lastPartType?: string;
   errorType?: string;
   errorSignature?: string;
+  disconnectSource?: 'downstream_client';
 }
 
 export type ProxyLifecycleEvent =
@@ -409,6 +411,7 @@ export function writeInferenceResponseLifecycleLog(
   path: string,
   entry: InferenceResponseLifecycleLogEntry,
 ): void {
+  const claudeSessionId = safeClaudeSessionId(entry.claudeSessionId);
   const statusCode = nonNegativeInteger(entry.statusCode);
   const durationMs = nonNegativeInteger(entry.durationMs);
   const timeToFirstByteMs = nonNegativeInteger(entry.timeToFirstByteMs);
@@ -428,6 +431,7 @@ export function writeInferenceResponseLifecycleLog(
     timestamp: new Date().toISOString(),
     event: entry.event,
     requestId: compactLogValue(entry.requestId, 100),
+    ...(claudeSessionId ? { claudeSessionId } : {}),
     modelId: compactLogValue(entry.modelId),
     provider: compactLogValue(entry.provider, 200),
     route: entry.route,
@@ -451,6 +455,9 @@ export function writeInferenceResponseLifecycleLog(
     ...(entry.lastPartType ? { lastPartType: compactLogValue(entry.lastPartType, 100) } : {}),
     ...(entry.errorType ? { errorType: compactLogValue(entry.errorType, 200) } : {}),
     ...(entry.errorSignature ? { errorSignature: compactLogValue(entry.errorSignature, 100) } : {}),
+    ...(entry.event === 'response_client_disconnected' && entry.disconnectSource
+      ? { disconnectSource: entry.disconnectSource }
+      : {}),
   }));
 }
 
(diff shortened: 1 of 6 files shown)


src/trace-log.ts
```diff
@@ -253,6 +254,7 @@ export interface InferenceResponseLifecycleLogEntry {
   lastPartType?: string;
   errorType?: string;
   errorSignature?: string;
+  disconnectSource?: 'downstream_client';
```
bman654: As written, `disconnectSource` carries no information: it has exactly one possible value, is only assigned when emitting `response_client_disconnected`, and is only serialized for that event — and upstream failures were already distinguishable because they emit a differently-shaped event (`response_failed` with `errorType`; the `failed`/`writableFinished` guards predate this PR). Meanwhile the one ambiguity that *does* exist stays masked: a local teardown (e.g. proxy shutdown destroying tracked sockets mid-stream) closes `res` with `writableFinished=false, failed=false` and gets labeled `downstream_client` even though the client never disconnected.

So three good options, in rough preference order: (a) promote this to a top-level `terminationSource` on all terminal events covering the three real scenarios (`downstream_client` / `local_shutdown` / `upstream_failure`), so consumers don't need to know the event taxonomy to answer "who ended this response?"; (b) keep it on the disconnect event but actually distinguish `downstream_client` from `local_shutdown`; or (c) drop the field. As-is it only adds bytes.

bman654 (changes requested):
Reviewed at a0959cc. The session-correlation feature is welcome and the execution is appropriately surgical: `claudeSessionId` extraction is just hoisted (no new per-request work), the field is threaded through existing lifecycle writers without touching adapter/streaming control flow, and the added tests are behavioral across translated, passthrough, disconnect, and invalid-id cases. Verified locally: typecheck + full suite green (69 files / 693 tests). Logging the raw session id is fine here — these logs live in the same security context as Claude Code's own session transcripts (which are literally named by this UUID), and the raw value is what makes cross-correlation with `~/.claude/projects/<proj>/<session-id>.jsonl` work.

The one change requested is `disconnectSource` (inline): as written it doesn't disambiguate anything — but there's a real taxonomy nearby that would make it earn its place, or it can simply be dropped.

No textual conflicts with #21 or #10 at current heads.

src/trace-log.ts
camjac251: Addressed in `e983d4c`. The one-valued field is now `terminationSource` on abnormal response endings, with distinct values for upstream failures, downstream disconnects, and local proxy shutdown. Shutdown is marked before tracked sockets are destroyed, and regressions cover each termination path.

bman654 (commented):
Content-wise this is good to go — `terminationSource` is now top-level on the terminal events with three distinct values, and local shutdown is flagged before the socket destroy (no longer conflated with `downstream_client`). That fully resolves the ask from last round.

One logistics item before it can merge: it now **conflicts with #16**, which just landed on `main` — both touch `src/proxy.ts` `startProxyCatalog` (your lifecycle `try/catch` vs #16's session-id extraction). Could you rebase over `main` and push? It should be a small resolution confined to that one region. Once it's clean I'll approve.

bman654 (approved):
Approving at 434c575. The rebase over #16 is clean and preserves both features.

The one conflict region was `startProxyCatalog` in `src/proxy.ts`. Resolution is correct: the `claudeSessionId` extraction (header + `extractClaudeSessionId(anthropicBody, …)`) is hoisted once above `createTranslationLifecycle`, so the single value now feeds both #16's `sdkTranslateRequest` path and this PR's lifecycle correlation — no duplicate extraction, neither feature lost.

Re-confirmed the content that was already good from last round is intact: `terminationSource` is top-level on the terminal events with three distinct values (`downstream_client` / `local_shutdown` / `upstream_failure`), local shutdown is flagged (`shuttingDown = true` at the top of `close()`) before the socket destroy so it's no longer conflated with `downstream_client`, and `claudeSessionId` is threaded through the passthrough + translated lifecycle writers and sanitized via the existing `safeClaudeSessionId`.

Verified locally on a branch merging this over current `main` together with #10: `pnpm typecheck` clean, full suite green (79 files / 889 tests).

Land-order note (no change needed here): if this lands alongside #10, there's one mechanical conflict in `src/http-proxy/server.ts` at the translated-route `forwardToAdapter` call — #10 switches the body arg to `adapterBody`, this PR adds `claudeSessionId` + trailing `() => shuttingDown`. Whoever lands second combines all three; I verified the combined form typechecks and the suite stays green.

PR #29, opened 2026-07-23
fix(transport): retry pre-frame websocket failures

camjac251:
## Summary

- retry a WebSocket request once when transport fails before the first response frame
- replay the complete original payload on a fresh socket while preserving request isolation
- share the existing retry budget and refuse replay after any upstream frame
- record bounded retry outcomes without retaining raw transport error text

## Root cause

The custom streaming transport returns its downstream response before the
outbound socket has necessarily opened. A connection reset at that point is
therefore delivered inside the response stream, after the ordinary request
retry boundary has already completed.

The recovery now lives at the socket boundary where the process can prove that
no upstream frame has been received. Incremental continuations are replayed
with their full original context and without a stale response identifier.

## Safety

- one retry maximum across transport and missing-continuation recovery
- no replay after any response frame or partial model output
- cancellation is checked before and after replacement socket creation
- parallel auxiliary requests remain isolated from reusable connection heads
- retry diagnostics fingerprint private error text instead of logging it

## Validation

- `pnpm exec vitest run tests/responses-websocket.test.ts` (39 tests)
- `pnpm test` (70 files, 733 tests)
- `pnpm typecheck`
- `pnpm build`
- `git diff --check origin/main...HEAD`
- public diff secret and personal-data scans

### DIFF
--- a/src/oauth/responses-websocket.ts
+++ b/src/oauth/responses-websocket.ts
@@ -82,6 +82,7 @@ interface RequestContext {
   responseId?: string;
   pendingEvents: unknown[];
   emittedModelData: boolean;
+  transportRetryPending: boolean;
   outputByIndex: Map<number, OutputAccumulator>;
   outputIndexByItemId: Map<string, number>;
   reasoningPartsByItemId: Map<string, Map<number, ReasoningPartState>>;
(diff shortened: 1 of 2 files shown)


bman654 (approved):
Approving. Surgical, additive change that honors the do-not-restructure directive and correctly reuses the existing single retry budget (`ctx.retried`).

No blocking findings — security is clean (error text hashed/bounded, replay stays within the original auth partition), and adversarial verification cleared the shared-budget concern (the untested cross-path is unreachable: a reset sets `continued=false`, so a transport failure can't chain onto a `previous_response_not_found` retry).

A few diagnostic/maintainability minors are non-blocking and tracked on our side: the retry choreography now exists in two copies with divergent cancellation guards; `transportRetryPending` isn't cleared on a post-dispatch cancel; `frameCount` isn't reset for the `exhausted` diagnostic branch.

PR #30, opened 2026-07-23
feat(logging): record proxy process lifecycle

camjac251:
@/tmp/clodex-pr-proxy-process-lifecycle.md


bman654 (approved):
Approving. Clean, pattern-consistent wiring of the existing `writeProxyLifecycleLog` helper.

Adversarial verification cleared both first-pass concerns: the hoisted log-path lookup still exits 1 via the top-level `.catch`, and the missing `proxy_process_exit` hook only drops a trace line on an unclean exit (no wrong behavior/leak/race).

One non-blocking test-hardening nit (tracked on our side): the lifecycle test doesn't assert `parentPid`, `host` on stopping/stopped, or the `reason` field, so those fields could regress undetected.

PR #31, opened 2026-07-23
fix(logging): attribute proxy transport failures

camjac251:
## Summary

- distinguish adapter request, early close, response error, response abort, and response close failures
- retain bounded transport error codes on response lifecycle records
- carry the recognized transport failure code into translated-response lifecycle records
- add deterministic transport-failure coverage through injectable request handling

## Privacy

Diagnostics retain only bounded error types, bounded error codes, and fixed failure-source values. Request and response content is not added to lifecycle records.

## Dependency

This draft is stacked on #26. Until that branch lands, the final commit is the topic-specific review scope.

## Validation

- type checking passes
- focused transport and lifecycle logging tests pass
- complete suite: 79 files and 881 tests
- production build passes
- diff hygiene and secret scanning pass

### DIFF
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -271,6 +271,7 @@ function forwardRawAnthropicRequest(
     provider: string;
     progressIntervalMs: number;
   },
+  isLocalShutdown: () => boolean = () => false,
 ): Promise<void> {
   return new Promise(resolve => {
     const startedAt = Date.now();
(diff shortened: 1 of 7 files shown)


camjac251:
Replaced by #43, which keeps the diagnostics change on current `main` and drops the lifecycle commits that already landed through #26.

Closing this draft so review stays on the conflict-free branch.

PR #43, opened 2026-07-24
fix(logging): attribute proxy transport failures

camjac251:
## Summary

- distinguish client disconnects, adapter request failures, and adapter response aborts in lifecycle diagnostics
- preserve bounded transport error codes without copying unbounded platform error text
- carry failure source and transport code through request traces and translated stream logs
- keep aborted upstream responses attributed to the hop where the failure occurred

## Test plan

- [x] focused diagnostics suite: 3 files and 58 tests
- [x] complete suite: 79 files and 892 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] changed-file secret scans

### DIFF
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -23,6 +23,7 @@ import {
   writeInferenceResponseLifecycleLog,
   writeInferenceResponseErrorLog,
   writeWebSocketDiagnosticRequestLog,
+  type InferenceFailureSource,
   type InferenceResponsePhase,
 } from '../trace-log.js';
 
@@ -161,6 +162,8 @@ export interface HttpProxyOptions {
   anthropicRejectUnauthorized?: boolean;
   /** Test hook for observing relay-route isolation without calling an AI provider. */
   adapterHandle?: ProxyHandle;
+  /** Test hook for exercising adapter request transport failures. */
+  adapterRequest?: typeof http.request;
   /** Test hook; production emits a progress record every 30 seconds. */
   responseProgressIntervalMs?: number;
 }
@@ -451,6 +454,7 @@ function forwardToAdapter(
   res: http.ServerResponse,
   rawBody: Buffer,
   adapter: ProxyHandle,
+  adapterRequest: typeof http.request = http.request,
   lifecycle?: {
     logPath: string;
     requestId: string;
@@ -546,7 +550,10 @@ function forwardToAdapter(
       resolve();
     });
 
-    const failAdapterRequest = (err: Error) => {
+    const failAdapterRequest = (
+      err: Error,
+      failureSource: InferenceFailureSource,
+    ) => {
       if (clientDisconnected) {
         resolve();
         return;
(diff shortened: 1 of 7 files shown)


bman654 (approved):
Reviewed at `88213d6`. This is a clean, well-scoped diagnostics change — approving.

Verified against the code rather than the description:

- **Attribution is genuinely first-failure-wins.** Both `failAdapterRequest` (`src/http-proxy/server.ts:561-562`) and `failAdapterResponse` (`:626-627`) return early on the existing `failed` guard, so the five new `failureSource` values can't double-log. On an abort, `aborted` fires before `close`, so the entry is tagged `adapter_response_aborted` rather than the vaguer close; on a socket error, `error` wins over both. That's the informative ordering, and it's the behavior the PR claims.
- **The client-disconnect hop really does win.** Both handlers check `clientDisconnected` first, so a downstream disconnect suppresses the upstream-failure record instead of racing it — which is the "attributed to the hop where the failure occurred" property.
- **`transportCode` is genuinely bounded.** `boundedTransportCode` (`src/upstream-error.ts:57-63`) preserves exactly one literal and maps everything else to `undefined`, so no unbounded platform text reaches the log. Confirmed the code is real and not speculative — `src/oauth/responses-websocket.ts:846` emits `websocket_transport_error` precisely when there's no HTTP status.
- **`errorCode` can't leak.** It's a Node errno on the adapter paths, `undefined` for the synthetic errors (so the field is omitted), and it goes through `compactLogValue(..., 100)`.
- No conflict with the 403→429 work that landed in #33 — this rebases onto it cleanly, and `upstream-error.ts` keeps doing only retry-after.
- Suite passes locally with an isolated `CLODEX_HOME` (79 files / 892 tests).

Nice catch on the second commit, too — reshaping the `sk-` fixture keeps repo secret scans useful, and the replacement still has ≥8 chars of `[A-Za-z0-9_-]`, so it exercises the same regex branch at `src/trace-log.ts:606`. Not a weakened test.

Non-blocking, take or leave:

- **Three of the five new sources are untested.** `adapter_response_error`, `adapter_response_aborted`, and `adapter_response_close` have no coverage — only the two request-side sources do. The response-side ordering (abort-beats-close) is the subtlest part of the change and the most likely to regress silently under a future refactor.
- **`boundedTransportCode` has no direct negative test.** The bounding is the load-bearing safety claim of the PR, and there's no test asserting that an unexpected or long `code` value is dropped. A two-line unit test on `sdkUpstreamErrorDetails` would pin it.
- **Nit:** `transportCode` is set unconditionally (`src/upstream-error.ts:92`), so the key is always present and sometimes `undefined`, while `retryAfterSeconds` two lines above uses a conditional spread. Harmless — `JSON.stringify` drops it — but the neighbouring style is inconsistent.
- **Nit:** `errorCode` now carries two vocabularies — Node errno on `response_failed`, the WebSocket sentinel on `translation_failed`. Distinguishable by event type, so it's fine, just worth being deliberate about if more codes get added later.

None of these block. Approved.

PR #44, opened 2026-07-24
fix(wrapper): retry transient listener checks

camjac251:
## Summary

- probe all ordered server candidates concurrently before retrying
- preserve proxy-first and newest-first selection when several candidates answer
- retry only timed-out probes under one shared 500 ms deadline
- treat connection refusal and other definitive errors as immediate failures
- keep single-listener startup readiness on the retry-all default
- use portable loopback reservations for refusal-path regression tests

## Test plan

- [x] deterministic readiness and wrapper suite: 2 files, 18 tests
- [x] complete suite: 79 files, 900 tests
- [x] type checking
- [x] production build
- [x] diff hygiene
- [x] staged secret scan
- [x] focused review of retry, deadline, and fallback behavior


bman654 (changes requested):
Reviewed at `92d6383`. The refactor is nice — collapsing the wrapper's ad-hoc probe and `listenTcpServer`'s retry loop into one `waitForTcpListener` is a genuine dedup, and the extracted helper is behaviourally identical for the `listenTcpServer` caller (same 1000ms default, same 50ms probe cap, same 5ms retry).

But I'd like one change before this lands, because of where the cost falls.

## The 500ms budget is per candidate, and it lands on the hot path · MAJOR

`docs/background-agents.md` and the PR description both say discovery stays "bounded to 500 ms". It's bounded to 500ms **per candidate**, inside the `for` loop at `src/claude-wrapper.ts:77-88` — so N advertised-but-unreachable servers cost 500ms × N before claude launches.

More importantly, the old `portIsOpen` returned essentially instantly for an unreachable port (loopback `ECONNREFUSED` is immediate); the 100ms was only an upper bound for a blackholed port. The new helper retries that immediate refusal every 5ms for the full budget — roughly 100 connect attempts. I measured both on this machine:

```
OLD  single closed-port probe:  2ms      three closed candidates:    2ms
NEW  single closed-port probe:  501ms    three closed candidates: 1500ms
```

That matters here specifically because `CLAUDE.md` calls out this binary as latency-sensitive — *"Keep the wrapper tiny and its imports minimal — it runs for every spawned agent process."* Adding up to 0.5–1.5s per spawn is a lot for a process that runs on every subagent launch.

And the triggering state isn't exotic — it's the one the code you removed was explicitly written for. `readLiveServerRuntimeStates` filters dead pids, so the expensive case is *live pid, closed port*: a stale `server-runtime.json` record whose pid has been recycled by an unrelated process (the old comment: *"the state file can outlive a SIGKILLed listener"*), or a server process that's up with a dead listener. Neither gets pruned, since pruning keys off pid liveness. In that state every single agent spawn pays the full budget, indefinitely. It's also reachable in the ordinary two-server setup: proxy is preferred over endpoint, so a stale proxy record makes every launch burn 500ms before falling through to the live endpoint server.

Suggested fix, which keeps both properties:

1. **Fast pass first** — one cheap probe across all candidates (~2ms total, i.e. today's behavior). If one answers, done.
2. **Then a bounded retry pass**, only if none answered, against a *single overall* deadline shared across candidates rather than a per-candidate one.

That preserves the transient-refusal recovery you're after while keeping the common and stale-record cases as fast as they are today.

## Smaller

- **The retry's motivation isn't stated, and it's worth pinning down.** A server registers itself only *after* `listenTcpServer` has confirmed reachability, so "registered but not yet listening" shouldn't be a real state. I assume you're targeting a full accept backlog under a burst of simultaneous agent launches returning `ECONNREFUSED` on loopback — which is plausible and worth fixing. Could you confirm the symptom you actually hit? It affects whether the retry belongs in discovery at all, versus being scoped to the burst case.
- **`CLAUDE.md` is now stale.** Its wrapper section still describes a "pid + <100ms TCP liveness check". `docs/background-agents.md` got updated; this one didn't.
- **Test gap:** the new test covers the positive case (listener appears late → `true`), but nothing covers the negative — that an unreachable port returns `false` *within* the budget. Given the finding above, a bound assertion is the test that would actually protect the invariant.
- **Nit:** the new test rebinds a port it just released, and `delayedServer.once('error', reject)` stays attached after the assertion settles. Low flake risk, but re-binding a just-freed port is inherently racy.

Happy to re-review quickly once the budget is shared across candidates — everything else here looks good.

camjac251:
The branch now uses one concurrent fast-pass across ordered candidates, followed
by retry rounds under a single shared 500 ms deadline. Selection still follows
the caller's proxy-first and newest-first order when more than one candidate
answers.

The readiness tests use injected probes, time, and delay to pin the exact shared
budget without process-startup timing. The wrapper integration tests now check
the selected environment and unavailable result without wall-clock assertions.
The registration documentation was also corrected: servers register after
listener readiness, so the retry covers transient refusal from an
already-advertised process.

The branch is rebased onto `main` at `e272754`. Verification is green:

- focused readiness and wrapper suite: 17 tests
- complete suite: 79 files, 899 tests
- type checking
- production build
- diff and secret checks

bman654 (changes requested):
The round-1 finding is genuinely fixed — I re-measured rather than taking the description on faith, and the numbers are convincing. One thing blocks merge: the new tests can't run on macOS.

## The fix works

Built each revision and timed `node dist/claude-wrapper.js --check` against a seeded `server-runtime.json` carrying a live pid, 7 runs per cell, median:

| scenario | main (e272754) | round 1 (409132c) | **this PR (52edb7a)** |
|---|---|---|---|
| no `server-runtime.json` | 39 ms | 45 ms | **38 ms** |
| one live + reachable server | 41 ms | 43 ms | **43 ms** |
| 1 record, live pid + closed port | 41 ms | 540 ms | **543 ms** |
| 3 records, live pids + closed ports | 43 ms | **1545 ms** | **541 ms** |
| dead proxy + reachable endpoint | 47 ms | 542 ms | **44 ms** |
| newer proxy dead, older proxy reachable | 50 ms | 565 ms | **42 ms** |

The N-scaling is gone and the reachable-fallback cases no longer pay the budget at all. Selection identity holds too — I checked the actual env handed to a fake claude, not just the exit code, and proxy-first / newest-first picks are byte-identical to main. `findIndex(Boolean)` genuinely preserves caller priority rather than first-to-answer, since `Promise.all` fills the results array in candidate order.

Also verified: no socket/fd leak across the full budget (`lsof` 22 → 22 after ~1215 probe creations, including a run where every probe exits via the timeout path), the loop arithmetic is bounded even against a packet-dropping peer (505 ms for a 500 ms budget, flat at 501 ms for three candidates), and `listenTcpServer` semantics are unchanged — same 1000 ms budget, same 50 ms probe cap, same close-on-failure and error text.

Your doc correction is also right: registration does happen strictly after `listenTcpServer` confirms reachability, in both modes.

## Blocking: the new tests fail on macOS

`openLoopbackServer('127.0.0.2')` (`tests/claude-wrapper.test.ts:76-93`) throws `EADDRNOTAVAIL` on macOS. Linux routes all of `127.0.0.0/8` to `lo`, but macOS `lo0` carries only `127.0.0.1` unless someone has run `sudo ifconfig lo0 alias 127.0.0.2`. So:

```
FAIL tests/claude-wrapper.test.ts > uses a live endpoint when the preferred proxy record is stale
FAIL tests/claude-wrapper.test.ts > reports unavailable when every candidate refuses connections
Error: listen EADDRNOTAVAIL: address not available 127.0.0.2

Test Files  1 failed | 78 passed (79)
     Tests  2 failed | 897 passed (899)
```

CI is `ubuntu-latest` (`.github/workflows/ci.yml:26`) so this stays green there, but `pnpm test` exits 1 for any macOS contributor — and the sharper problem is that `uses a live endpoint when the preferred proxy record is stale` is precisely the end-to-end regression guard for the round-1 fix. On macOS it dies in setup instead of exercising the fix.

**Suggested fix**, verified on macOS here: get a guaranteed-refusing port by binding `127.0.0.1:0`, reading `address().port`, and closing the server — keep the number. That yields `ECONNREFUSED` in 0.4–1.1 ms and behaves identically to what you want. (An IPv6-only `::1` bind probed over IPv4 also works, 2–3.5 ms, if you prefer holding a listener open.)

One thing that may have motivated `127.0.0.2` — that binding it *reserves* the port so a parallel test can't take it — doesn't hold: binding one address never reserves the port on a different address in the same family. I confirmed by binding `192.168.68.104:56108` and then successfully binding `127.0.0.1:56108`. So the bind-and-close approach has identical collision exposure with none of the portability cost.

## Worth reconsidering: is the retry earning its keep?

The all-unreachable case now costs ~500 ms per wrapper invocation where main cost ~0 ms, and CLAUDE.md notes the wrapper runs for every spawned agent process. Against that, the benefit looks close to zero:

- Graceful shutdown unregisters before `close()` in both modes, and a SIGKILLed server leaves a dead pid that `readLiveServerRuntimeStates` filters out — so the wrapper sees zero candidates and pays nothing.
- The only state that produces the 500 ms is a recycled pid.
- As your own doc change now says, registration happens *after* a successful loopback TCP connect. So a registered server has provably accepted a connection on the exact address/port the wrapper probes, and on loopback a live listening socket doesn't transiently emit `ECONNREFUSED`.

That leaves accept-backlog overflow (SYN drop) as the only real transient — which needs ~511 simultaneous connections to a Node server. What symptom did you actually hit? If it was backlog overflow, the targeted fix is better than a blanket retry: `probeTcpListener` currently collapses `error` and `timeout` into the same `false` (`src/listener-ready.ts:30-32`), so surfacing the reason and retrying **only on timeout** would keep the protection while treating `ECONNREFUSED` as definitive and restoring main's latency. Dropping the wrapper retry entirely (keeping it in `listenTcpServer`) is the other clean option.

## Smaller notes

- The unit tests do protect the shared-deadline behavior — I reverted to a per-candidate budget and `shares one exact deadline across all unreachable candidates` goes red on `elapsedMs` (24 vs 12). Good guard. But consider loosening `expect(probeTimeouts).toEqual([12, 7, 2])` and the six-entry `probes` array: they freeze the literal value of `LISTENER_READY_RETRY_MS` and the exact `Math.min` expression, so a pure tuning change (5 → 4) turns three tests red with no behavior change. The `elapsedMs` assertions alone catch the real regression.
- No test asserts the wrapper stays *fast*, which is what would have caught round 1. Once the macOS issue is fixed, wrapping the `--check` call in `process.hrtime.bigint()` and asserting `>= 400 && < 1200` would pin it — measured separation is 575–616 ms for this PR vs 1593–2631 ms for a per-candidate regression, so it isn't flaky.
- `src/claude-wrapper.ts:73` ends mid-sentence with a dangling "The" left over from the old comment.

camjac251:
The latest review items are addressed in `ccca229`.

Wrapper discovery now distinguishes timed-out probes from definitive connection failures. Only timed-out candidates retry under the shared deadline; refused or unreachable candidates drop out immediately. Listener startup retains its retry-all behavior.

The refusal tests use IPv6-only reservations instead of `127.0.0.2`, and the three-candidate path is bounded below 400 ms.

Current verification is green: 18 focused tests, 900 tests across 79 files, type checking, production build, secret scan, and both GitHub checks. The current head is mergeable.

bman654 (approved):
Round 3 — **approving.** Both round-2 items are fixed, and I verified each one rather than taking the summary at face value.

**macOS blocker: gone.** Full suite runs **900/900 green on macOS** (previously 2 failures every run, one of which was the regression guard itself). `127.0.0.2` no longer appears anywhere in `src/`, `tests/`, or the docs.

Your IPv6-only reservation is a better choice than the `127.0.0.1:0`-then-close recipe I suggested — measured on macOS, an IPv4 connect to a `::1`-bound `ipv6Only` port refuses in **1.85 ms**, deterministically. (One thing worth knowing: it does not actually reserve the port against an IPv4 bind — I bound `127.0.0.1` on the same port successfully while the v6 listener was up. That is no worse than the alternative, just not a property to rely on.) The only environments this cannot serve are ones with no `::1` at all, e.g. a container started with IPv6 disabled; GitHub's runners and macOS both have it, so this is a note, not an ask.

**Retry-only-on-timeout: implemented as discussed, and it removes the cost entirely.** `probeTcpListener` now distinguishes `ready` / `timeout` / `unreachable`, the wrapper passes `retryFailure: result => result === 'timeout'`, and `waitForTcpListener` keeps retry-all for listener startup. Measured, 3 stale records with no live server, 5 runs each:

| | latency |
|---|---|
| `main` (single-shot probe, no retry) | ~48 ms |
| this PR at `52edb7a` (round 2) | ~546 ms |
| this PR at `ccca229` (round 3) | **~48 ms** |

So the retry now costs nothing in the case that actually happens — a refused loopback probe drops the candidate immediately — while a genuinely hung probe still gets the shared 500 ms budget. That is exactly the right shape.

**Selection identity is unchanged.** I compared the env actually handed to a fake `claude` binary, `main` vs this branch, across both-live / proxy-dead-endpoint-live / none-live / no-records / two-proxy-newest-wins: **byte-identical in all five**, including `NODE_EXTRA_CA_CERTS` and the newest-startedAt tiebreak.

**The tests hold the fixes.** Four mutations, each reddens at least one test: dropping the `retryFailure` option (1 failure), mapping every socket error to `timeout` (1), ignoring `retryFailure` in the pending-candidate filter (2), and selecting the last ready candidate instead of the first (2). The latency guard at `claude-wrapper.test.ts:356` and the docs correction in `docs/background-agents.md` are both there and accurate. Typecheck clean.

Two leftovers, neither worth holding the PR for:

- `expect(elapsedMs).toBe(10)` in the retry test still pins `LISTENER_READY_RETRY_MS` — retuning 5→4 reddens it. That is down from 3 tests to 1, so this is much better than round 2; take it or leave it.
- Worst case for a single genuinely blackholed candidate is now 500 ms vs `main`'s 100 ms. Bounded and unreachable in practice on loopback — just noting it is a real (and deliberate) trade.

Nice work on this one — the round-2 measurement request turned out to matter, and the shared-deadline plus timeout-only policy is a better design than what either of us started with.

PR #45, opened 2026-07-24
build(deps): drop node-gyp-build workaround via tweakcc 4.3.2

bman654:
## Summary

`clodex patch` loads `tweakcc`, which pulls in `node-lief` to unpack and repack the native Claude Code
binary. clodex has been carrying a direct `node-gyp-build` dependency that no clodex source imports,
purely to work around **node-lief 1.3.1**: that release demoted `node-gyp-build` to a devDependency while
its runtime loader still `require`d it, so fresh end-user installs got a node-lief that threw
`Cannot find module 'node-gyp-build'`. tweakcc's lazy loader swallowed that into a null, and clodex
surfaced it as the misleading `Failed to extract JavaScript from native installation`.

**node-lief 1.3.2 removes the `node-gyp-build` require entirely** — `lib/index.js` now requires the
platform prebuild by path, and the package declares no runtime `dependencies` at all. The workaround is
no longer needed, so this drops it.

- `tweakcc` 4.3.0 → 4.3.2
- transitive `node-lief` 1.3.0 → 1.3.2
- removes the direct `node-gyp-build` dependency
- updates the patcher + toolchain notes in `CLAUDE.md` and the header comment in `src/tweakcc.d.ts`

## Freshness guard

Both new versions are younger than the repo's 10-day `minimumReleaseAge` supply-chain guard
(tweakcc 4.3.2 published 2026-07-20, node-lief 1.3.2 published 2026-07-23), so the lockfile was refreshed
with a deliberate one-shot `pnpm install --config.minimumReleaseAge=0`, plus
`pnpm update node-lief --config.minimumReleaseAge=0` — an ordinary install will not move a *transitive*
pin that is already in the lockfile.

**The repo-wide setting is unchanged.** CI installs with `--frozen-lockfile`, so these now-locked versions
install fine while the 10-day guard stays in force for every future resolution. `CLAUDE.md` documents this
as the sanctioned escape hatch so the next person doesn't weaken the setting instead.

## API-surface re-verification

`CLAUDE.md` requires re-verifying `src/tweakcc.d.ts` whenever the tweakcc pin moves, because tweakcc
declares a `types` field but ships no `.d.ts`. 4.3.2 still ships no `.d.ts`, and its
`dist/lib/index.mjs` export list is byte-identical to 4.3.0's (`backupFile`, `findAllInstallations`,
`readContent`, `restoreBackup`, `tryDetectInstallation`, `writeContent`, …). The `content-*.mjs` chunk that
implements `tryDetectInstallation` / `readContent` / `writeContent` differs only by minifier variable
renaming (16171 vs 16172 bytes). The existing shim remains accurate; no changes were needed beyond the
version in its comment.

## Residual risk

tweakcc still declares `node-lief: ^1.3.0`, and a published package's `overrides` do not apply to its
consumers — so clodex cannot *force* end users onto ≥ 1.3.2. Anyone resolving exactly 1.3.1 (only reachable
now via a stale lockfile or cache, since 1.3.2 supersedes it) would see the old failure again. `CLAUDE.md`
records this and says to restore the direct `node-gyp-build` dependency if it ever shows up in the wild.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 79 files, 889 tests passing (run with an isolated `CLODEX_HOME`)
- [x] `pnpm build` — succeeds
- [x] **native repack path exercised for real**: with node-lief 1.3.2 and no `node-gyp-build` present,
      `tryDetectInstallation` resolved the installed native Claude Code 2.1.219 and `readContent`
      extracted 21,636,065 chars of JS — the exact operation that fails under the 1.3.1 packaging bug
- [x] confirmed `node-gyp-build` no longer appears anywhere in `pnpm-lock.yaml`


bman654:
Closing as superseded by #129. tweakcc 4.3.3 makes the entry-module workaround unnecessary altogether, so there's no point bumping to 4.3.2 first. #129 tracks removing the workaround safely: it has to keep a fail-closed macOS code-signing check, which the workaround currently provides by accident.

PR #47, opened 2026-07-24
test(logging): cover bounded proxy failure attribution

camjac251:
## Summary

- omit unsupported transport codes from structured failure details
- cover adapter response error, abort, and incomplete-close attribution
- pin abort-before-close ordering at the proxy boundary

Follow-up to the optional hardening notes in #43's final review.

## Test plan

- [x] `pnpm test` (79 files, 897 tests)
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] `git diff --check origin/main...HEAD`
- [x] branch secret scan

### DIFF
--- a/src/upstream-error.ts
+++ b/src/upstream-error.ts
@@ -82,6 +82,7 @@ export function sdkUpstreamErrorDetails(err: unknown): SdkUpstreamErrorDetails |
   const retryAfterSeconds = rawRetryAfter === undefined
     ? undefined
     : clampRetryAfterSeconds(rawRetryAfter);
+  const transportCode = boundedTransportCode(inner.data);
 
   return {
     statusCode: inner.statusCode,
@@ -89,7 +90,7 @@ export function sdkUpstreamErrorDetails(err: unknown): SdkUpstreamErrorDetails |
     isRetryable: inner.isRetryable,
     attemptCount: retry?.errors.length ?? 1,
     ...(retryAfterSeconds !== undefined ? { retryAfterSeconds } : {}),
-    transportCode: boundedTransportCode(inner.data),
+    ...(transportCode !== undefined ? { transportCode } : {}),
   };
 }
 
--- a/tests/http-proxy-server.test.ts
+++ b/tests/http-proxy-server.test.ts
@@ -7,6 +7,7 @@ import * as http from 'node:http';
 import * as net from 'node:net';
 import * as tls from 'node:tls';
 import { EventEmitter, once } from 'node:events';
+import { PassThrough } from 'node:stream';
 import { gzipSync } from 'node:zlib';
 import { ensureHttpProxyCaBundle, ensureHttpProxyCertificates } from '../src/http-proxy/ca.js';
 import { shouldInterceptConnect, startHttpProxy } from '../src/http-proxy/server.js';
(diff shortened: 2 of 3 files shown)


bman654 (approved):
Approving. I verified the one source change rather than taking the commit type on faith, and `test:` is right.

**The `src/upstream-error.ts` change is a behavioral no-op.** The only reader of `transportCode` anywhere in the tree is `src/proxy.ts:685` (`details?.transportCode`), an optional-chained read where present-but-`undefined` and absent are indistinguishable. There is no `'transportCode' in details`, no `Object.keys(details)`, no spread-merge, no structural equality. `exactOptionalPropertyTypes` is not enabled (`tsconfig.json` sets only `strict`), so it is not a type-correctness fix either, and `src/trace-log.ts:477` emits `errorCode` only when truthy. So no JSONL byte and no HTTP response changes — producing no changelog entry and no version bump is the correct outcome.

**The new tests are not vacuous.** Reverting the hunk to `transportCode,` fails both new `tests/upstream-error.test.ts` cases. Mutating `'adapter_response_aborted'` → `'adapter_response_error'` fails the new abort test plus the pre-existing "terminates and logs a translated response when the adapter closes before end"; mutating `'adapter_response_close'` fails exactly the new close test.

Green here too: 79 files / 897 tests, typecheck clean, build clean, commitlint clean. Both touched test files run 3x with no variance (~1.4s of test time against 20s timeouts) and leak no temp dirs.

### Optional follow-ups, none blocking

**Two of the three new proxy tests pin attributions production cannot emit.** The synthetic `PassThrough` lets the test choose an event order a real `http.IncomingMessage` never produces. On Node v24.14.1 (what `.nvmrc` pins) I probed four termination modes and two HTTP parse-error modes:

```
headers-then-socket-destroy   -> aborted | error(ECONNRESET) | close(complete=false)
partial-body-then-destroy     -> data | aborted | error(ECONNRESET) | close(complete=false)
short-content-length-then-end -> data | aborted | error(ECONNRESET) | close(complete=false)
invalid-chunk-size-mid-body   -> RES aborted | RES error(ECONNRESET) | RES close(complete=false)
```

`aborted` always precedes `error`/`close` on an incomplete response, and since `failAdapterResponse` dedupes on `failed`, `adapter_response_aborted` always wins. I confirmed end to end by driving the real proxy with a real adapter server that `flushHeaders()` then destroys its socket before the first body byte — production logged:

```json
{"event":"response_failed","statusCode":200,"phase":"waiting_for_first_byte","bytes":0,"chunks":0,
 "errorType":"Error","failureSource":"adapter_response_aborted","terminationSource":"upstream_failure"}
```

That is `adapter_response_aborted` with **no** `errorCode` — not `adapter_response_error` + `ECONNRESET`. So `errorCode: 'ECONNRESET'` at `tests/http-proxy-server.test.ts:1344` asserts a log line that will never be written. Worth keeping the branches (they are cheap defensive coverage, and I could not *prove* unreachability), but consider dropping that one assertion and rewriting the genuinely valuable abort-before-close dedup pin with a real adapter server that resets before the first byte — that reproduces the asserted entry exactly and needs no `PassThrough`.

**The two transport-code tests exercise the same branch.** `boundedTransportCode` is an exact-equality allowlist against the single literal `'websocket_transport_error'` — it has no length logic, so "omits an **overlong** WebSocket transport code" is rejected by the same `!==` as the "unexpected" case. Either fold them into one table-driven case or rename to something like "omits any code other than `websocket_transport_error`".

**Commit body is slightly over-stated.** "pin transport-code bounding so unexpected upstream values never enter the structured failure schema" — bounding shipped in #43, and unexpected values already could not enter the schema. What changed is key presence on an in-memory object no consumer reads.

PR #54, opened 2026-07-26
fix(proxy): reuse private adapter connections

camjac251:
## Summary

- isolate the loopback translation adapter from Node's process-global HTTP pool
- retain healthy internal connections across ordinary request gaps
- cover private-pool reuse with a local transport regression test

## Problem

Translated requests currently use the process-global HTTP agent for the private loopback hop. Its short socket lifetime causes frequent reconnects between tool turns and couples adapter traffic to unrelated HTTP users in the process. A reset during that connection phase fails the request before the adapter can return response headers.

The proxy now owns a keep-alive agent for that hop. The adapter advertises a 60-second idle window, allowing the private pool to retain healthy sockets while leaving active request timeouts unchanged. Startup failures still destroy the pool immediately, while normal shutdown preserves the existing local-shutdown lifecycle ordering.

## Validation

- `pnpm vitest run tests/http-proxy-server.test.ts`
- `pnpm typecheck`
- `pnpm test` (80 files, 1,026 tests)
- `pnpm build`
- `gitleaks git --staged`

### DIFF
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -455,6 +455,7 @@ function forwardToAdapter(
   rawBody: Buffer,
   adapter: ProxyHandle,
   adapterRequest: typeof http.request = http.request,
+  adapterAgent?: http.Agent,
   lifecycle?: {
     logPath: string;
     requestId: string;
@@ -584,6 +585,7 @@ function forwardToAdapter(
       port: adapter.port,
       method: 'POST',
       path: req.url,
+      agent: adapterAgent,
       headers: {
         'Content-Type': 'application/json',
         'Content-Length': String(rawBody.length),
@@ -730,6 +732,7 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       options.modelAliases,
     );
   }
+  const adapterAgent = adapter ? new http.Agent({ keepAlive: true }) : undefined;
   let shuttingDown = false;
 
   const mitmServer = https.createServer({
@@ -848,6 +851,7 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
           adapterBody,
           adapter,
           options.adapterRequest,
+          adapterAgent,
           messagesEndpoint === 'messages' && options.inferenceLogPath
             ? {
                 logPath: options.inferenceLogPath,
@@ -971,6 +975,7 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       options.host ?? '127.0.0.1',
     );
   } catch (err) {
+    adapterAgent?.destroy();
     adapter?.close();
     throw err;
   }
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -71,6 +71,7 @@ type ProxyLog = (message: string | (() => string)) => void;
 // still surfaced by the SDK idle watchdog rather than masked by pings forever.
 const STREAM_KEEPALIVE_INTERVAL_MS = 20_000;
 const STREAM_KEEPALIVE_PING = 'event: ping\ndata: {"type":"ping"}\n\n';
+const INTERNAL_ADAPTER_KEEPALIVE_TIMEOUT_MS = 60_000;
 
 function createTranslationLifecycle(
   logPath: string | undefined,
@@ -749,6 +750,7 @@ export async function startProxyCatalog(
     // Everything else → 404
     anthropicError(res, 404, `Unknown endpoint: ${req.method} ${req.url}`);
   });
+  server.keepAliveTimeout = INTERNAL_ADAPTER_KEEPALIVE_TIMEOUT_MS;
 
   let address: AddressInfo;
   try {
--- a/tests/http-proxy-server.test.ts
+++ b/tests/http-proxy-server.test.ts
@@ -1200,6 +1200,90 @@ describe('selective HTTP proxy', () => {
     }
   }, 20_000);
 
+  it('keeps translated adapter connections out of the process-global pool', async () => {
+    const certificates = ensureHttpProxyCertificates();
+    let connectionCount = 0;
+    const adapterServer = http.createServer((req, res) => {
+      req.resume();
+      req.once('end', () => {
+        res.writeHead(200, {
+          'Content-Type': 'application/json',
+          'Content-Length': '2',
+        });
+        res.end('{}');
+      });
+    });
+    adapterServer.keepAliveTimeout = 60_000;
+    adapterServer.on('connection', () => {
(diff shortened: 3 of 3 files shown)


bman654 (approved):
Verified this one by measurement rather than by reading — the premise, both halves of the fix, and the shutdown/race questions it raises all held up. Approving.

## The premise checks out, and both halves are load-bearing

I probed Node 24.14.1 directly rather than take the description on faith. Defaults: `http.globalAgent` is `keepAlive: true` with a **5s** socket timeout, and `http.Server.keepAliveTimeout` is also **5s**.

| scenario | connections across a 6s idle gap |
|---|---|
| globalAgent + default 5s server timeout (today's `main`) | **2** — the reconnect you describe |
| globalAgent + 60s server timeout (server half only) | **2** |
| private agent + 60s server timeout (this PR) | **1**, 200 OK |

The middle row is the interesting one: raising `keepAliveTimeout` alone would have accomplished nothing, because globalAgent's own 5s socket timeout is the binding constraint. So the private agent isn't just isolation hygiene — it's what makes the 60s window reachable at all. The two changes only work as a pair, which is worth stating because they read as separable.

## Questions I had, and what measurement said

- **Does bumping 5s → 60s delay shutdown?** No. `server.close()` with an idle keep-alive socket outstanding fires its callback in **0ms** — Node ≥19 releases idle connections on close. And the client-side free socket is unref'd, so leaving `adapterAgent` undestroyed doesn't hold the event loop.
- **Does the asymmetry (server retires at 60s, the agent has no idle timeout of its own) leave a reuse race?** Not in practice. I hammered the boundary — 40 requests timed onto the server's socket-retirement instant — and got **0 ECONNRESET**; every reconnect was clean. I'd expected to raise this and it didn't survive contact with a measurement.
- **`keepAliveTimeout` (now 60s) == `headersTimeout` (default 60s) — premature 408?** No. Reuse at 90% of the window returns 200 on the same connection.
- **Can a pooled socket ever be recycled dirty and desync the next response?** This was the one I most expected to bite, since the adapter streams SSE and `forwardToAdapter` abandons responses on several paths. It doesn't. Aborting mid-stream after 4 events leaves 0 live adapter responses and sends the *next* request over a brand-new connection; three sequential completed streams share one connection with exact framing. The reason is that `IncomingMessage._destroy` only kills the socket while `!readableEnded || !complete`, so the `upstream?.destroy()` in the `res.once('close')` handler is a no-op once the response has completed and the socket has been freed — forcing the collision with `maxSockets: 1` still succeeded on a single connection.
- **Concurrency:** 300 concurrent requests complete in 157ms (serialized would be ~15s), so the default `maxSockets: Infinity` isn't accidentally serializing anything. Idle-FD retention grows from 5s to 60s, bounded by `maxFreeSockets: 256` — far above real Claude Code concurrency.
- Suite at the head: **1026 tests / 80 files green**, typecheck clean, build clean, commitlint exit 0, no `dist/` or version edits. Sits directly on current `main` with no divergence.

Since this touches a launch path, I also ran the manual smokes from the PR head build rather than ask you to — `CONTRIBUTING.md` asks for these and they aren't in the description:

- `clodex claude --proxy -p …` → Anthropic passthrough, `SMOKE_OK_ANTHROPIC`
- `clodex claude --proxy -p … -- --model sol` → translated path over the now-pooled adapter hop, streaming SSE intact, `SMOKE_OK_OPENAI`
- the same translated path multi-turn with a **7s** gap between adapter requests — i.e. straddling the 5s window where `main` drops the socket — clean round trip, no reset

## Non-blocking — none of this needs to hold up the merge

**1. The `src/proxy.ts` half has no test coverage.** Mutation-tested: delete `server.keepAliveTimeout = INTERNAL_ADAPTER_KEEPALIVE_TIMEOUT_MS;` and the full 1026-test suite is still green. The `agent:` half is properly pinned — dropping `agent: adapterAgent` reddens the new test — so this clears the bar in `CONTRIBUTING.md` ("at least one of them actually pins the behavior you changed"). But a bare assignment tucked after a large `createServer(...)` call is exactly the kind of line a future refactor drops silently, and it's the half that carries the user-visible change. A one-liner asserting `handle`-side `keepAliveTimeout`, or an end-to-end test with a >5s idle gap, would close it.

**2. The new test doesn't verify what its name claims.** It's called *"keeps translated adapter connections out of the process-global pool"*, but it can't distinguish a private agent from the global one: since Node 19 `globalAgent` is also `keepAlive: true`, so two back-to-back requests through the global pool reuse a connection too and would count 1 all the same. The signal comes entirely from the stub's own `agent: options.agent ?? false` coercion. Proof: remove that coercion so plain `http.request` is used **and** revert the src fix, and the test still passes. What it actually pins is "an agent object is threaded through to the adapter request" — genuinely useful, just narrower than the name. A rename, or an idle gap longer than globalAgent's 5s timeout, would make the name true.

**3. `INTERNAL_ADAPTER_KEEPALIVE_TIMEOUT_MS` has a wider blast radius than its name.** `startProxyCatalog()` has two callers: the internal adapter hop (`src/http-proxy/server.ts:725`) and endpoint mode (`src/cli.ts:576`), where the client is Claude Code itself, not the adapter. So this also retimes the user-facing gateway from 5s to 60s. I think that's fine-to-beneficial — the server just advertises `Keep-Alive: timeout=60` and undici honors it — but the name implies a narrower scope than the code has. Worth a comment, or a name that doesn't say "internal".

**4. `adapterAgent` is destroyed on the startup-failure path but not on the normal `close()`.** Measured harmless (see above — no hang, no leaked loop reference, and `adapter.close()` FINs the sockets anyway), so this is pure symmetry: `adapterAgent?.destroy()` alongside `adapter?.close()` in the returned `close()` would make the two teardown paths match.

**5.** `expect(firstResponse).toContain('200')` matches any `200` anywhere in the raw response text, including a header value. `toContain('HTTP/1.1 200')` costs nothing.

**6. A side effect worth knowing about, though I could not turn it into a defect.** `copyResponse` forwards the adapter's `rawHeaders` verbatim, so the MITM now advertises `Keep-Alive: timeout=60` **downstream to Claude Code** — while `mitmServer` itself still enforces Node's default 5s, since `src/proxy.ts:753` is the only place a `keepAliveTimeout` is set. That inverts which side retires first on the downstream hop. I tried to break it and couldn't: undici at 8s gaps gives `ok=6 errors=0 serverConns=7` either way, and sweeping the gap across 4990–5016ms gives `ok=14 errors=0`. So it's an inaccurate advertisement with no measurable consequence today — noting it only because it's the kind of thing that gets confusing later.

Nice, tightly-scoped fix — the failure mode it targets is real and the mechanism is right.

PR #56, opened 2026-07-26
fix(adapter): prevent cached input usage inflation

camjac251:
## Summary
- keep local input estimates out of initial streamed usage
- prefer provider-reported input usage and retain estimates only as completion fallbacks
- cover fully cached input and missing-usage streams

## Root cause
The adapter emitted estimated input usage at stream start and cache-split provider usage at completion. Consumers that preserve zero-valued fields could retain the estimate alongside cache-read tokens, inflating context totals and triggering compaction too early.

## Test plan
- [x] `pnpm exec vitest run tests/sdk-adapter.test.ts tests/proxy.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm build`
- [x] `gitleaks dir . --no-banner --redact`

### DIFF
--- a/src/sdk-adapter.ts
+++ b/src/sdk-adapter.ts
@@ -605,7 +605,7 @@ type LogFn = (msg: () => string) => void;
 export interface AnthropicStreamObserver {
   /** Called for every AI SDK fullStream part before Relay translates it. */
   onPart?: (partType: string) => void;
-  /** Local estimate used until the provider reports actual usage at stream completion. */
+  /** Local fallback used when the provider omits usage at stream completion. */
   initialInputTokens?: number;
   abortSignal?: AbortSignal;
   /** Abort if the provider produces no stream event for this long. */
@@ -664,7 +664,7 @@ export async function writeAnthropicStream(
   let openToolId: string | null = null;
   let finishReason = 'end_turn';
   let usage: AnthropicUsage = {
-    input_tokens: 0,
+    input_tokens: observer?.initialInputTokens ?? 0,
     output_tokens: 0,
     cache_creation_input_tokens: 0,
     cache_read_input_tokens: 0,
@@ -679,7 +679,7 @@ export async function writeAnthropicStream(
         id: messageId, type: 'message', role: 'assistant', content: [],
         model: modelId, stop_reason: null, stop_sequence: null,
         usage: {
-          input_tokens: observer?.initialInputTokens ?? 0,
+          input_tokens: 0,
           output_tokens: 0,
           cache_creation_input_tokens: 0,
           cache_read_input_tokens: 0,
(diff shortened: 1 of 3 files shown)


bman654 (approved):
Approving — this is a correct root-cause fix and the tests pin it. Thanks for the clear write-up; the "consumers that preserve zero-valued fields" line is exactly right, and it made this quick to confirm.

## What I verified

I decompiled the 2.1.220 bundle rather than take the mechanism on faith. Claude Code's usage merge is:

```js
t.input_tokens > 0 ? t.input_tokens : e.input_tokens
```

applied per field — **last-non-zero-wins**, not add, not last, not max. `message_start` seeds the accumulator, `message_delta` merges into it, and the result is written to the transcript. The context total is `input + cache_creation + cache_read + output`, and it feeds both the `/context` gauge and the auto-compact preflight. So an estimate emitted at `message_start` can never be corrected by a later `message_delta` that reports `input_tokens: 0` — which is precisely the inflation you describe.

Reproduced on a harness driving the real `writeAnthropicStream` against a port of that merge function: a 173,000-token fully-cache-read turn with a 61,500-token local estimate reports **~235,000 before this change (+35%) and ~173,000 after — the true value**. Uncached, partial-cache, no-usage, and zero-input streams are identical between `main` and this branch.

Also confirmed: `toAnthropicUsage` is untouched so the cache-token mapping is intact with no double counting; the estimate fallback still fires when the provider reports nothing; `estimateAnthropicInputTokens` and its flat per-image estimate are unchanged. I checked every consumer of these fields — `openai-adapter`, the http-proxy observers, `server/router`, `count_tokens`, and the OAuth WebSocket path — and none of them regress. I could not construct a NaN, negative, or double-counted value; `NaN` routes to the estimate, which is an improvement on the previous `"input_tokens": null`.

Test quality clears the `CONTRIBUTING.md` bar: four separate mutations, including narrowing the guard to `input_tokens > 0` alone, each redden a named test for the right reason.

## Non-blocking notes

None of these hold up the merge.

1. **Interrupts and mid-stream errors now record zeros.** Claude Code materialises an assistant message at `content_block_stop` using `message_start` usage, and we never emit a `message_delta` on an interrupt or a mid-stream failure. That orphan message previously carried the estimate and now carries `{0,0,0,0}`. Because context is anchored on the last assistant message with usage, `/context` reads ~0% and the next auto-compact preflight is skipped — the "too late" direction, which is the one that hurts in this codebase. Pressing ESC during a thinking→text transition is enough to hit it.

   I'm explicitly **not** asking for a fix, because I don't think a better emission shape exists: once an estimate lands in `message_start` the merge above makes it permanently uncorrectable, the pre-change value was itself a wrong bytes/4 estimate, and it self-heals on the next turn. Emitting a synthetic delta on the error path would also perturb the client's retry branch. A short source comment recording the trade-off would be welcome, and I'll track the interrupt path separately.

2. Zeroed `message_start` rows now show up in the `--trace` inference log. Diagnostics only, but mildly confusing when reading a trace.

3. The inline model of the client's merge in the test omits `output_tokens` and will drift silently if the client changes. A comment naming the upstream function it mirrors would help the next person.

4. The non-streaming path still has no estimate fallback. Pre-existing, not yours.

5. One test name is broader than what it asserts.

PR #57, opened 2026-07-26
fix: preserve extended effort levels in patched clients

camjac251:
## Summary

- preserve distinct extended effort selections through provider translation
- advertise only capabilities with distinct wire behavior while retaining native defaults and denylist behavior
- issue explicit capability verdicts for configured aliases, including suffixed identities
- preserve canonical alias normalization and rejected-alias reporting from the current base
- build and validate re-patch candidates without replacing the working binary on failure

## Verification

- full suite: 1,127 tests passed
- type checking
- production build
- staged and branch-range secret scans
- two independent review passes and a final patch-lifecycle verification


bman654 (changes requested):
Good change and the analysis behind it is right — I confirmed the divergence you describe is real, and the extended values now reach the wire intact. Three things to fix before I merge, all small.

## Verified

I checked the transform work against real pristine bundles for 2.1.217 through 2.1.220 rather than the test fixture:

- All ten anchors match **exactly once** on every version, with an identical delta.
- The patched 21.6 MB bundle passes `node --check`, and the insertions land exactly where intended.
- Idempotent: a second run reports SKIP for every site and the output is byte-identical under `cmp`. The refresh path rewrites correctly when the config changes, markers stay unique, and the result still parses.
- With no effort metadata configured, the output is **byte-identical to `main`**. With effort configured, the only difference from `main` is the four new snippets — PATCH 1–7 are untouched.
- No `$`-substitution hazard: every new replacement is function-form, verified with a model id containing `$&`, `$1`, `` $` ``, `$'`, and `$$`.
- `NATIVE_EFFORT_LEVELS` exactly matches the client's own ladder (`["low","medium","high","xhigh","max"]`), `none` is correctly dropped, and `projectNativeEffort` provably cannot install a default outside the projected set.
- Existing installs correctly evaluate `stale-config` and repatch, so nobody is left on a stale binary.

On the provider side, the divergence you describe is real — I built the full mapping table before and after and the `xhigh → high` downgrade and the omitted `max` are both there on `main`. The extended set matches `@ai-sdk/openai`'s own documented GPT-5.6 values. I also checked the OAuth WebSocket partition key, since effort participates in it: it hashes the post-mapping wire value, so it becomes strictly more discriminating and can't produce new collisions.

I ran a live OAuth round-trip on `gpt-5.6-luna` at every level. `xhigh` and `max` go out verbatim and come back with real content, so the backend accepts them — worth stating explicitly, because that backend fails with an empty response rather than an error when it dislikes a parameter, and nothing client-side would have caught it.

## Please fix before merge

### 1. Make the new sites `required: false`

All four of PATCH 8a/8b/8c/9 are `required: true`. `applyPatch` restores the pristine backup *before* patching, so a single upstream literal change aborts the whole run and leaves the user **unpatched** — no aliases, no context-window map, nothing.

I reproduced this: changing `default_effort??"high"` to `"medium"` in the real 2.1.220 bundle takes PATCH 1 through 8c to OK, PATCH 9 to FAIL, and the entire patch throws. Effort is the newest and most-churned subsystem in the client, so this is the site most likely to drift, and it's taking the safety-critical context map down with it.

Degradation is graceful in every combination if these are optional — the client clamps `max` and `xhigh` down to `high` on its own, the default falls back to `"high"`, and the capability map is never consulted when the gate is false. So `required: false` costs nothing and removes the failure mode.

### 2. Pin effort in the config hash

Reverting just the `computePatchConfigHash` hunk leaves all 36 patcher tests green, so nothing catches it if the effort dimension stops participating — and if it stops participating, users who change effort metadata are told their binary is current and never get repatched. There's already a precedent test for the display-label dimension; extending it by one assertion covers this.

### 3. Don't advertise `xhigh` where we downgrade it

This is the one I'd most like changed, because it reintroduces the exact problem the PR sets out to fix, one tier down.

`projectNativeEffort` advertises `["low","medium","high","xhigh"]` for `gpt-5.5`, `gpt-5.4`, `gpt-5-codex`, and `o3`. But `mapCodexEffortToOpenAI` is gated on `/^gpt-5\.6/`, so for those models `xhigh` returns `'high'` on the wire.

And this is new behavior, not pre-existing. In the client, the effort value is clamped to `high` unless the per-model xhigh capability gate returns true, and that gate consults a model table with no clodex ids in it — so before this change the client clamped `xhigh` itself and never sent it. PATCH 8b flips that gate on, so now the picker offers `xhigh`, the user selects it, and the request carries `high`. Selected effort and upstream request diverge, silently.

The fix is to feed `projectNativeEffort` the *effective* post-mapping levels rather than the raw capability list, so a level is only advertised if we actually forward it. That also future-proofs the `/^gpt-5\.6/` regex, which won't match the next model generation.

## Non-blocking

- The GPT-5.6 scope guard is unpinned — mutating `isGpt56Model` to `return true` leaves the whole suite green. A negative test for `gpt-5.5` would cover both this and item 3 above.
- The four changed assertions in `tests/sdk-adapter.test.ts` weren't necessary — I re-applied the originals against your `src` and they pass. That's coverage of the pre-5.6 path traded away rather than added. No hidden regression, but I'd restore them.
- `patcher.ts` passes `model.upstreamModelId` into `getReasoningCapabilities` unguarded, and that function calls `.toLowerCase()` immediately. Registry IO doesn't validate model fields, so a cache without that field throws. Low reachability; `?? model.id` fixes it.
- PATCH 8 injects ahead of the client's own runtime effort denylist and its env override, so a model the client has decided doesn't support effort would still get our gate. Nearly unreachable, and trivially fixed by moving the injection after that check.
- The new sites emit no SKIP row under `--trace` when they don't apply, which makes a partial patch harder to diagnose.
- No test covers the effort-refresh path.
- The `CLAUDE.md` edit describing the hash lost some precision — the tuple's elements are no longer spelled out. The other edited claims check out.
- `none` is unreachable from the client ladder and only reachable through the server API. Worth a comment so it doesn't look like dead code.
- One comment overstates how the picker mechanism works.

Full suite, typecheck, build, and commitlint are all clean, and the second commit resolves real defects in the first.

camjac251:
Addressed the requested changes in b2894a99d80ac696ad68a3317a5a68bec9c80a35.

- effort-only anchors now fail independently without aborting unrelated patches
- capability and default fields are covered by patch-state hash regressions
- advertised choices now come from distinct provider options; missing or duplicate wire paths are excluded
- refresh coverage includes reduced and fully removed capability and default tables
- legacy upstream identifiers retain a safe fallback

Verification: 150 focused regressions, 1,053 full-suite tests, typecheck, build, native read-only transform checks, and staged secret scan all passed.

bman654 (changes requested):
All three asks are genuinely addressed — I verified each by re-running the original failure experiment rather than reading the diff. Thank you for that; the `required: false` change in particular I confirmed by re-drifting the anchors in a real bundle.

Unfortunately, decompiling the pristine 2.1.220 binary to check the ask-3 fix end to end turned up two problems that the clodex-side tests can't see. Both are in the patch sites, both affect the default bridge mode, and neither is your fault in the sense of being sloppy — they're the kind of thing only the real binary reveals.

## Blocking

### 1. Omitting a level does not deny it — the fix is inert in proxy mode

`projectNativeEffort` now correctly drops `xhigh` for `gpt-5.5`/`gpt-5.4`/`gpt-5-codex`/`o3`, and at the clodex layer the advertised-vs-wire table is clean, zero violations. But PATCH 8b/8c inject a **positive-only** map, and the binary treats a miss as "unknown", not "no":

```js
function LN(e,t){ let n=ww(r); if(n!==void 0) return n.capabilities.includes(t)?!0:void 0; return XWm?.(r,t) }
function Pye(e){ …; if(LN(r,"xhigh_effort")||r==="claude-mythos-5") return !0; return p9(n_(e)) }
function p9(e=xn()){ return e==="firstParty"||aq(e)||e==="foundry"||e==="mantle" }
```

`LN` returns `void 0` — not `false` — when the capability is absent, so control reaches `p9(n_(e))`. In **proxy mode**, which is clodex's default, Claude Code believes it is talking to first-party Anthropic, so `p9()` is `true`.

Net effect: `/effort` still offers `xhigh` and `max` for every non-GPT-5.6 Responses favorite. The user picks `max`, and `mapCodexEffortToOpenAI('max','gpt-5.5')` returns `undefined`, so clodex sends **no effort field at all** and the provider applies its own default. That is the same silent selected-vs-sent divergence this PR exists to eliminate.

The reason the suite doesn't catch it: the new test fixture's gates `return!1` on fallthrough, while the real binary returns `p9(...)`. The fixture is more restrictive than reality.

**Fix:** bake explicit true/false verdicts for every effort-capable model into PATCH 8b/8c so a miss returns `!1` instead of falling through.

### 2. PATCH 9 silently downgrades default effort from `high` to `medium` for existing users

`lit(e)` is `ww(lo(e))?.default_effort ?? "high"`. On `main`, a clodex model isn't in that table, so the default resolves to **`high`** — verified, and `FI` is true via the same fallthrough as above, so clodex models really do send `output_config.effort: "high"` today.

PATCH 9 bakes an explicit default, and `projectNativeEffort` currently returns `defaultLevel: "medium"` for every model I checked. So after the forced repatch, **every existing proxy-mode user's default effort drops from high to medium**, including users who have never touched effort and don't know this feature exists. That's an unannounced quality regression, and the PR body doesn't mention it.

Either preserve `high` as the baked default, or make the change deliberate and call it out prominently — but it shouldn't arrive silently as a side effect of a patch bump.

## Verified — all three asks met

- **Ask 1.** Re-ran the exact abort experiment on `~/.tweakcc/claude-2.1.220.orig`: `default_effort??"high"` → `"medium"` now yields PATCH 1–8c OK, PATCH 9 FAIL, and **no throw**. I also drifted all four effort anchors simultaneously — identity and context sites still apply — then repacked into a binary copy and confirmed `--version` works with the ctx markers intact.
- **Ask 2.** Reverting the hash tuple hunk reddens two named tests, for the right reason: the two configs collapse to an identical sha256.
- **Ask 3.** At the clodex layer, the full truth table across gpt-5.6-sol/luna/terra, gpt-5.5, gpt-5.4, gpt-5-codex, o3 and case variants shows zero violations — every advertised level round-trips unchanged. `/^gpt-5\.6/` is bounded rather than fixed: a future gpt-5.7 under-advertises instead of mis-advertising, which is the safe direction. (Blocking item 1 above is about the *binary's* behavior on a map miss, not about this projection.)

Also confirmed sound: all four new anchors match exactly once on the real bundle, all ten sites apply, refresh and marker cleanup work, patched 21.6 MB bundles pass `node --check`, hash staleness produces `stale-config` correctly, the OAuth WebSocket partition key is unaffected, and the runtime source path is unchanged for users with no effort configured.

## Non-blocking

- **The ask-3 fix's call site is unpinned.** Reverting only `getPatchReasoningCapabilities` → `getReasoningCapabilities` at `patcher.ts:174` leaves typecheck clean and 1053/1053 green while reintroducing the gpt-5.5 leak. A ~30-line temp-home test closes it.
- `upstreamModelId ?? model.id` is a genuine TypeError fix, but it's repairing a crash this PR introduced at `7d5105a`, and it's untested.
- The new sdk-adapter GPT-5.5 test is pure characterization — it didn't redden under any mutation I tried.
- The stale-effort-table clearing in PATCH 8/9 wasn't requested, but it follows the existing PATCH 2/7 refresh pattern and is test-pinned. No objection.
- Latent, footnote only: the dedup can remove a *middle* level and void the whole ladder; the only model that triggers it is unreachable in clodex today.

Suite, typecheck, build, and commitlint are all clean, and I found no unrequested scope in the two newest commits. Note that this still conflicts with #59 in `src/patcher.ts` and `tests/patcher.test.ts`; whichever lands second will need a rebase.

camjac251:
Addressed the current runtime findings in `ecf45d2`.

- every configured identity now receives an explicit capability verdict, including suffixed forms and identities without projected effort
- the native denylist and high default remain authoritative; only unknown identities fall through
- explicit non-reasoning metadata, normalized collisions, and prototype-like names now fail safely
- required effort-anchor failures no longer persist a partial binary or current manifest

Verified with 121 focused capability and patch regressions, type checking, the production build, the native artifact transform, and a staged secret scan.

bman654 (changes requested):
Requesting changes — but first, credit where it is due: **both blockers from the last round are genuinely fixed**, and I verified them against the pristine 2.1.220 bundle rather than the test fixture, since the fixture is what hid these last time.

## Verified fixed — not re-litigating these

**1. Omitting a level now actually denies it.** The capability snippet emits explicit boolean verdicts instead of a positive-only map, so a configured model that lacks `xhigh` gets a hard `false` and `if(_ccv!==void 0)return _ccv;` short-circuits before control ever reaches `return p9(n_(e))`. Patched with a two-model config (one with the full ladder, one without), the injected `Pye` gate reads:

```js
function Pye(e){if(S2r(e))return!1;/*ccpatch:xhigh-effort*/var _ccv=Object.assign(Object.create(null),
{"sol":true,"sol[1m]":true,...,"g55":false,"g55[1m]":false,...})[...];if(_ccv!==void 0)return _ccv;
```

That closes the proxy-mode leak. All 9 sites report OK on the real bundle.

**2. No more default-effort downgrade.** PATCH 9 now injects `high` for every key, which matches the binary's own `ww(lo(e))?.default_effort??"high"` fallback for clodex ids. Existing proxy-mode users see no change.

**The anchor relocation is also correct and is an improvement.** Moving injection to after `if(S2r(e))return!1;` matters because `S2r` is `Mt.effortUnsupportedModels.has(e)` — a set populated at runtime by `TFn()` when a model is observed not to support effort. Letting that learned denial win over a baked-in verdict is the right precedence.

Also good: `CLAUDE.md` is updated (PATCH 1–9, the widened hash tuple, effort capabilities in the patch description), and `findModelsDevModel` reads the bundled/on-disk cache rather than the network, so `buildDesiredPatchConfig()`'s documented disk-only invariant still holds.

## Blocking: a failed effort patch now leaves the user with a completely unpatched binary

This is the one I would not merge without. `applyPatch` gained:

```ts
const failedEffortPatches = requiredEffortPatchFailures(results);
if (failedEffortPatches.length > 0) { throw new PatchApplyError(...); }
await writeContent(installation, patched.content);
```

Combine that with the restore-first invariant a few lines above. `runPatchCommand` computes `restoreFirst = existsSync(backup) && sha256File(backup) !== sha256File(binaryPath)`, which is **true for every already-patched user**. So the sequence for anyone re-patching is:

1. `copyFileSync(backup, binaryPath)` — the binary is reverted to pristine, wiping their working patch.
2. One effort site reports FAIL.
3. Throw. `writeContent` never runs, the manifest is never updated.

The user is left with a pristine, unpatched binary: no Agent-tool enum entries, no alias resolution, no context windows — because an *effort* anchor drifted. Those are the features people install clodex for, and they are lost to a failure in the newest, most drift-prone sites.

This is reachable on any future Claude Code release that moves one of the four new anchors. I simulated it on the real bundle by renaming the capability string `wde(e,"xhigh_effort")` to `"xhigh_effort2"`:

```
DRIFT patch8b-drift: PATCH 1=OK PATCH 3=OK PATCH 6=OK PATCH 5=OK PATCH 4=OK PATCH 8b=FAIL PATCH 8c=OK PATCH 9=OK
```

Everything else still applies cleanly. On `main` that drift is harmless. This also directly reverses round 1's ask 1, where we asked for `required:false` precisely so anchor drift would degrade instead of throwing — the throw has just moved up a layer.

I do understand the motivation, and it is a real concern: if 8a succeeds but 8b fails, the effort menu is enabled while `xhigh`/`max` fall through to the native `p9()` — reintroducing exactly the silent divergence this PR exists to kill. A partial effort patch genuinely is unsafe. But the remedy should be scoped to the effort feature rather than the whole patch. Either of these works:

- **Treat the effort sites as an all-or-nothing group.** If any of 8a/8b/8c/9 fails, skip the group entirely — including not injecting 8a — and continue applying 1/3/4/5/6/7. The user keeps aliases and context windows and simply has no clodex-managed effort, which is exactly `main`'s behavior today. Surface it as a warning in the outcome.
- **Or, if you prefer to fail loud, do not leave the binary worse than you found it** — restore the previously patched binary (or defer the pristine restore until after `applyClodexPatches` has succeeded) so a failure is a no-op rather than a downgrade.

The second point is worth doing regardless of which you pick: `applyPatch` currently destroys working state *before* it knows whether it can produce replacement state. Moving the restore after a successful `applyClodexPatches` makes every failure mode in this function safe, not just this one.

## Blocking, but the fix is now mostly "delete it": the alias-collision hard-fail

`applyClodexPatches` gained a `fail()` when one alias maps to multiple models, and `registerCapabilityKeys` throws on any identity-key conflict. Both abort the entire patch, through the same restore-then-abort path above.

It is reachable today: `cli.ts` dedupes aliases with an exact string comparison and `MODEL_ALIAS_PATTERN` permits uppercase, so `Sol` and `sol` pointing at different models both persist. Both lowercase to the same key. On the real bundle:

```
CASE case_variant_same_name_diff_models: THREW clodex patch: alias "sol" is assigned to multiple models   (this PR)
CASE case_variant_same_name_diff_models: OK (0 non-OK)                                                     (main)
```

**Here is the important part for your rebase.** #59 just merged as `5fec19a`, and it owns this problem now. `buildPatchModelConfig` on current `main` calls `normalizeModelAliases(aliases)`, which groups aliases by canonical name and, when one name points at multiple targets, rejects **every** occurrence with `conflicting-targets` — preserving the saved entries in config, reporting them through `describeModelAliasRejection`, and letting everything else proceed. That is the policy we want: reject the ambiguous alias, warn, keep patching.

So when you rebase, **take #59's solution and drop the collision checks this PR adds.** After the rebase the config reaching `applyClodexPatches` cannot contain a duplicate alias name — normalization has already removed it — which makes the `fail()` unreachable in the good case and actively harmful in the bad one. Having two of your own PRs implement opposite policies for the same situation is the thing to resolve here, and #59's is the one that survives.

The identity-key conflict check in `registerCapabilityKeys` should go the same way. For what it is worth, the `[1m]` collision it guards against is not reachable anyway: `buildPatchModelConfig` calls `stripOneMContextSuffix` and dedupes with `if (config[id]) continue`, so a config key never carries the suffix. I confirmed a `X` + `X[1m]` favorite pair throws on this branch and patches cleanly on `main`, but only via a hand-built config — so that specific case is hygiene, not a live bug.

## Non-blocking

1. **PATCH 9 is now effectively a no-op.** `projectNativeEffort` validates the model's declared `defaultLevel` is among its levels and then discards it, returning `'high'` unconditionally — and `'high'` is already what the unpatched binary produces for clodex ids. The site is harmless and arguably good defense-in-depth, but a reader will wonder why `defaultLevel` is computed and validated at all. Either drop the field or leave a comment saying the validation is a guard and the value is deliberately fixed.

2. `Object.create(null)` on the internal maps and in the injected snippets is a real (if small) hardening win, since the alias pattern permits names like `constructor`. Unrequested, but I have no objection.

3. The new `provider-factory.ts` early return for `reasoning === false` is fine and covered by the new tests.

4. The added `sdk-adapter` GPT-5.6 `xhigh` test is pure characterization — it passes on `main` too. Harmless.

Suite is green at 1085 on your head, and the effort tuple is properly pinned in the hash. Once the two items above are sorted this should be close to done.

camjac251:
Addressed the latest findings in `5a4159f` after integrating the current base.

- canonical alias normalization now owns conflicting saved aliases: ambiguous entries remain saved, are reported, and are omitted from patch configuration without aborting unrelated work
- re-patching now extracts, transforms, validates, and repacks a temporary candidate while the installed binary remains untouched
- validation, repacking, setup, and candidate-metadata failures preserve the working binary and existing manifest
- only a complete candidate is published, and temporary cleanup cannot override the patch outcome
- regression coverage now exercises failed validation, failed repacking, and successful publication with matching manifest metadata

Verified with 1,127 full-suite tests, type checking, the production build, staged and branch-range secret scans, two independent review passes, and a final patch-lifecycle verification.

bman654 (approved):
Approving. Both blockers are resolved, and I verified them rather than reading the diff — including re-running the binary checks against the pristine 2.1.220 bundle after your rebase, since `patch-transforms.ts` moved underneath you when the alias work landed.

## The destructive-failure blocker — fixed, and your solution is better than the one I suggested

I proposed either grouping the effort sites all-or-nothing or restoring the previous binary on failure. You did something cleaner: build the candidate in a temp dir beside the live executable, patch and validate *that*, and `renameSync` it into place only once it is complete. No failure path touches the working binary at all, and deriving `patchedSize`/`patchedSha256` from the candidate before publication means the manifest can never describe a binary that was never published.

Checked the things that restructure could quietly break:

- The candidate dir is created in `dirname(binaryPath)`, so the rename is same-filesystem and atomic. Good.
- `copyFileSync` preserves mode (verified `755` in, `755` out, and `755` after `renameSync`), so the executable bit survives publication. This was my main worry and it is fine.
- The `finally` block cleans up the candidate dir and swallows cleanup errors behind `--trace`, so a cleanup failure cannot mask the real outcome.

And it is properly pinned. Re-introducing the old destructive `copyFileSync(backup, binaryPath)` before patching turns two of your three new tests red:

```
× preserves the working binary and manifest when a re-patch fails validation
× preserves the working binary and manifest when candidate repacking fails
Tests  2 failed | 75 passed (77)
```

That clears the `CONTRIBUTING.md` bar squarely.

## The alias-collision hard-fail — removed, correctly

You took the canonicalization path rather than keeping your own guard, which is the right call now that `buildPatchModelConfig` normalizes upstream. Reproduced the case that used to abort the whole patch:

```
Sol -> clodex:openai:m1, sol -> clodex:openai:m2
=> no throw; 2 aliases rejected; both models still patched (without aliases)
```

Previously this threw `alias "sol" is assigned to multiple models` and the user got no patch at all. Now the ambiguous alias is dropped with a warning and everything else proceeds — same policy the rest of the codebase applies.

## Round-2 fixes re-verified post-rebase

Patching the real pristine bundle still yields all 9 sites OK, with the capability verdicts intact:

```
Pye: {"sol":true,...,"g55":false,"g55[1m]":false,...}    (xhigh denied for a model without it)
lit: {"sol":"high",...,"g55":"high",...}                 (no default-effort downgrade)
```

So neither the explicit-deny fix nor the PATCH 9 fix regressed in the rebase. Suite green at 1127.

## One residual note — not blocking

The throw on a failed effort site still means a **first-time** patch produces nothing at all if a future Claude Code release drifts one of the four effort anchors, where the pre-effort behavior would have applied the identity, alias, picker, and context sites and simply skipped effort. For an already-patched user this is now harmless — they keep their working patch and get a clear error — which is why this is a note rather than a blocker. But if you ever want the degradation path, grouping 8a/8b/8c/9 so a failure skips the effort feature while the other sites still apply would give you both properties. Entirely your call, and fine to leave as is.

Nice work across three rounds on a genuinely fiddly area.

PR #58, opened 2026-07-27
feat: opt in to Claude Code native computer use

cgasgarth:
## Summary

- add an explicit `CLODEX_NATIVE_COMPUTER_USE=1` opt-in for Claude Code’s bundled `computer-use` MCP
- enable that built-in MCP globally under the same opt-in instead of requiring a per-project toggle
- version the binary patch digest so existing installations restore and repatch after transform changes

## Safety

The patch preserves Claude Code’s HIPAA, macOS, and interactive-session guards. It does not bypass the native per-app approval dialog, computer-use lock, screen filtering, or macOS Privacy & Security controls. Removing the environment variable restores upstream behavior.

## Validation

- Claude Code 2.1.220: `/mcp` reported `computer-use · connected · 24 tools`
- live Sol and Luna smoke tests completed native app approval and screen interaction
- 1,029 tests passed
- `pnpm typecheck` passed
- `pnpm build` passed

### DIFF
--- a/src/patch-transforms.ts
+++ b/src/patch-transforms.ts
@@ -65,7 +65,7 @@ export function formatPatchSiteLine(result: PatchSiteResult): string {
 }
 
 /**
- * Apply the clodex patch sites (PATCH 1–7) to the Claude Code source.
+ * Apply the clodex patch sites (PATCH 1–9) to the Claude Code source.
  * Pure: source string in → patched string + per-site results out. Throws
  * `PatchApplyError` when the config is invalid or a required site fails —
  * nothing should be written to the binary in that case.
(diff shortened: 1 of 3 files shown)


bman654 (commented):
Thanks for the careful work here — the patch sites are well built and the write-up is honest about what they do and don't touch. I verified the claims and they hold: both anchors match exactly once in the 2.1.220 bundle, the HIPAA guard stays first and authoritative, the surrounding call site still requires macOS and an interactive session, and PATCH 9 is correctly scoped to the single `computer-use` built-in rather than all built-in MCP servers.

I'm going to decline this one anyway, for two reasons.

## It's outside what clodex is for

clodex exists to bridge Claude Code to third-party models. The existing patch sites are all load-bearing for that: PATCH 1–7 exist because we inject model identities that Claude Code's own validation, picker, resolver, and context-window map would otherwise reject. Without them the bridge doesn't work.

These two sites have no relationship to model routing, providers, translation, or aliases. They behave identically for someone who never routes a single token to OpenAI. That makes them a Claude Code feature mod rather than clodex functionality, and I don't want to take on maintenance of feature mods against a binary that moves every few days — each additional patch site is another anchor that can break and another way `clodex patch` can fail for everyone.

If there is a genuine clodex angle to computer use, it's the one this PR doesn't address: the screenshots it produces flow through our `tool_result` image-lifting path on the way to a third-party model, which is delicate code with its own token-accounting rules. That would be a real integration question, and it isn't what's here.

## What PATCH 8 actually removes

This is the part that decides it. The predicate is:

```js
function Lbo(){ if(lK("hipaa")) return !1; return QE_() && iEs().enabled }
```

Following both gates:

- `QE_()` is `let e = Aa(); return e === "max" || e === "pro"`, and `Aa()` resolves to the account's `subscriptionType`. That's a paid-plan entitlement check.
- `iEs()` reads the server-controlled config `tengu_malort_pedway`, whose local default is `enabled: false`. There's no supported way to override it locally in 2.1.220 — the env-override reader returns before it ever parses the variable, and the secondary override function is a bare `return`. That value only comes from Anthropic.

So the only thing PATCH 8 removes is Anthropic's own entitlement and staged-rollout control over a feature that drives the user's machine through Accessibility and Screen Recording — which is presumably why it's being staged in the first place. For an account that already has both gates satisfied the patch changes nothing at all; its entire practical effect is enabling the feature for people whose account isn't entitled to it. I'm not comfortable shipping that in a public package, and I don't think the project should be in the business of unlocking paid features of the host application.

Worth noting for your own validation: if your account is on Max or Pro and Anthropic has enabled the rollout flag for you, you can already run the built-in computer-use MCP unpatched, enabling it per project. That's the supported path, and it's the one I'd point people at.

## One thing I am taking

Your `PATCH_SCHEMA_VERSION` change is a real bug fix and it's the best thing in this PR. `evaluatePatchState` decides `current` from the binary path, claude version, patched size, and a hash over the *model* config only — nothing represents the transform set. So a user whose favorites and aliases haven't changed, upgrading to a clodex with new or changed patch sites, gets told "already patched — nothing to do" and never receives them. Folding a schema version into the digest is exactly the right fix.

I've added it to our own backlog rather than asking you to resubmit it, because it needs to be sequenced behind two other open PRs that are already touching `computePatchConfigHash`. You found it; thank you. It'll land with credit to this PR in the commit.

Closing this one — but please do keep the review contributions coming. The quality of the analysis here is not in question.

PR #59, opened 2026-07-27
fix: canonicalize aliases without unsafe fallback

camjac251:
## Summary

- canonicalize saved alias names while preserving exact inactive records
- route only canonical active names and block every known legacy spelling from fallback
- fail closed on malformed, reserved, conflicting, unavailable, or colliding aliases
- preserve exact provider and model identifiers across routing surfaces

## Verification

- focused normalization, routing, diagnostics, and patch regressions: 94 passed
- server routing regressions: 32 passed
- type checking
- production build
- staged secret scan
- two independent review passes


bman654 (changes requested):
The canonicalization itself is sound and unusually well-pinned — but the upgrade path destroys existing user config, so I can't merge this yet. Two blocking items, both small.

## Blocking

### 1. Removing one alias silently erases the others, permanently

This is the one that worries me. Starting from a real `config.json` with five saved aliases and removing exactly one:

```
BEFORE: MixedCase→gpt-5.6-sol, Dup→gpt-5.5, dup→gpt-5.4, keeper→o3, opus→gpt-5.6-luna

$ clodex models --unalias keeper
  ◆  Removed model alias keeper.

AFTER:  mixedcase→gpt-5.6-sol          (1 entry)
```

Four aliases gone from a one-alias removal, `MixedCase` rewritten to `mixedcase` on disk, and the only thing the user is told is "Removed model alias keeper."

The mechanism: `loadPreferences()` normalizes and filters at the load boundary, and every production consumer reads through it, so the `rejected` channel added in `http-proxy/routes.ts` is unreachable in production — `normalizeModelAliases` is idempotent, so by the time anything inspects `rejected` it is always empty. Then the `models --alias` / `--unalias` path read-modify-writes that already-filtered list back over `config.json`. The in-memory normalization the PR description describes as non-destructive becomes destructive on the next write of any kind.

The PR description says "existing preferences are normalized in memory without rewriting the configuration file." That's true of the load path in isolation and not true of the product.

What I'd like:
- **Never write back entries you filtered out on load.** Preserve unknown/rejected entries verbatim through a read-modify-write so an upgrade can't eat a config.
- **Tell the user once**, at the point where it matters. `reportSkippedHttpProxyFavorites` is the natural home, and it would make the existing `rejected` plumbing live instead of dead.

Deactivating a genuinely ambiguous `Dup`/`dup` pair is defensible. Erasing them from disk with no notice isn't, and an alias that silently stops resolving with no diagnostic is the specific failure mode I want to avoid here.

### 2. `default` is missing from the reserved list

`opus`, `inherit`, and `sonnet` are correctly rejected with "That alias name is reserved by the client". `default` isn't — it passes the reserved check and only stops on an unrelated "not a saved favorite" guard, so it becomes assignable as soon as the target is a favorite.

In the 2.1.220 bundle `"default"` is a sentinel at three separate sites, all resolving to the account's default **Claude** model — including a `r.trim().toLowerCase()==="default"` comparison, so casing doesn't help. A saved `default` alias would be baked into the enum, get its own picker row, and silently run the session against Anthropic instead of the aliased model. One-line fix.

While you're in that list: it would help to add a short comment recording where the entries came from and against which client version, so the next person can re-derive it.

## Non-blocking

- **Mixed-case model ids on the wire now miss everything.** The client sends the model id with its original casing (`Ei()` trims but does not lowercase — it returns `bO(t)` where `bO` is the identity). After this change a mixed-case id misses both lookup tables *and* `reservedModelIds`, so proxy mode passes it through to `api.anthropic.com` rather than returning our own 400. Case-folding the lookup key in the three resolution sites would make the change purely additive and fix `--model Sol` for unpatched users, which currently regresses (and there's a new test codifying that regression).
- `--unalias` on a reserved or deactivated name is a dead end: `--unalias Sol` reports "No model alias named sol is saved" when it is saved, and `--unalias opus` refuses because the name is reserved — so entries that exist on disk can't be removed through the CLI. The reserved gate has no purpose on the *removal* path; dropping it there fixes both messages.
- The two new `fail()` calls in `patch-transforms.ts` abort the entire patch run, which takes PATCH 7's context-window map down with them. A skip-with-warning would degrade better.
- `rejected` conflates harmless collapsed duplicates (same target) with real drops (different targets). Only the second kind needs to reach the user.
- `--help` for `models --alias` doesn't mention the new casing or reserved-name rules.

## Verified correct — no need to revisit

I went at the parts most likely to be subtly wrong and they held up:

- `toLowerCase()` is locale-invariant here (`'SIL'` → `sil`, versus a Turkish-locale fold to `sıl`), so there's no dotted/dotless-I hazard. No NFC/NFKC normalization is needed either — every non-ASCII fold fails the ASCII pattern afterwards, with the Kelvin sign the only survivor and harmless. Trim, empty, and length bounds all behave.
- The `enum == sent == echoed == context-map key` chain agrees end-to-end for a mixed-case alias, checked against real binary anchors, and collisions fail closed on both the patch and proxy sides with no half-state.
- `http-proxy/routes.ts` still forwards bodies unrewritten, `server/models.ts` resolution precedence is unchanged, and `/models` still advertises exactly the canonical/masked ids.
- No `$`-substitution hazard reintroduced — validate-after-canonicalize is the right order and the accepted character class is narrower than before.
- All nine consumers were updated, and all seven `src` hunks are pinned by named tests that redden when reverted.

Test gaps worth closing while you're here: the `rejects reserved names` CLI test actually exercises `model-aliases.ts` rather than the CLI path, the `--unalias` reserved branch is untested, and nothing pins the join that matters most — that the MITM route key equals the value baked into the enum.

## On the premise

Worth correcting for the record, because it changes what the fix is defending against. `main`'s `patch-transforms.ts` does lowercase the alias identity, and route lookup is case-preserving, so the bug is real and reachable — the fix is justified. But the stated mechanism ("after the client normalized them") isn't what happens: the client doesn't normalize the id at all. The bug exists because *we* bake a lowercased alias into the enum while matching case-sensitively at the route. Same conclusion, different reason.

camjac251:
Addressed the requested changes in 5fb6193be19c6774145bd9d0af1244d557dc3fc1.

- raw inactive entries survive read-modify-write operations and warnings state they were preserved
- the missing reserved sentinel is covered and its source version is documented
- reserved and deactivated aliases can be removed
- equivalent duplicates collapse without warning; true conflicts remain inactive
- saved aliases resolve case-insensitively across proxy and endpoint paths while exact provider and model IDs retain precedence
- response echo now follows actual alias-resolution provenance

Verification: focused alias and routing regressions, 1,042 full-suite tests, typecheck, build, and staged secret scan all passed.

bman654 (changes requested):
Both blockers are genuinely fixed, and you fixed the first one the right way — reverting `loadPreferences` to main and moving normalization to the consumers is better than what I asked for. I reproduced both fixes rather than reading the diff.

Two things stand between this and a merge: the fix introduced one regression in endpoint mode, and the second commit carries a substantial amount of scope nobody asked for that I'd like split out.

## Verified fixed

- **Destructive rewrite — gone.** Same five-alias fixture: `--unalias keeper` now removes only `keeper` and leaves `MixedCase`, `Dup`, `dup`, and `opus` verbatim on disk, where the old head left one of five. The add path preserves all five and appends. `--list` writes nothing and warns *"4 model aliases skipped … Saved entries were preserved."* Only two writers of `modelAliases` exist and both filter the raw array now. Reintroducing the destruction reddens two named tests.
- **`default` — fixed, and in the right place.** `default`/`DEFAULT`/`Default` are all rejected from `parseModelAliasAssignment`, which runs before the favorite guard (confirmed with a target that *is* a favorite). I also re-derived the list against the pristine 2.1.220 bundle: a full sweep of `toLowerCase()==="…"` comparisons against model names returns only `default` and `inherit`. **Nothing missing, nothing fictional** — that closes the provenance question from last round.

## Blocking

### Endpoint mode now silently serves the default model for a rejected alias

This is a regression: before this PR, case-distinct alias pairs worked. Now, on the endpoint launch path (`src/cli.ts:1240` → `startProxyCatalog`), a request for a collision-rejected alias returns **200 OK against the default model**. Proxy/MITM and the standalone server both return 400.

Silently getting a different model than you asked for is the worst outcome available here — worse than the original destructive rewrite, because nothing at all signals it. It's also the same failure I blocked on last round ("an alias silently stops resolving"), just relocated from the CLI to the endpoint path. And no `reportSkippedHttpProxyFavorites` call site sits on the endpoint or `clodex server` path, so the warning that makes the CLI case survivable never fires there.

The fix looks small: surface the rejected names into `configuredAliasNames` so the endpoint route table rejects them the way the MITM does.

## Please split out: the case-insensitive matching

`src/proxy.ts` and `src/http-proxy/server.ts` are hot-path files that neither blocker required touching. I can show mechanically that the case-folding they add isn't needed:

- Revert both files entirely to `main`, and revert the two **pre-existing** tests that were changed to mixed case (`llama` → `LLaMa`, `sol` → `SoL`), and **all 1042 tests pass**.
- It's unnecessary by construction: all runtime alias names are lowercase after canonicalization, and the patched binary can only send lowercase — I verified `Ei()`/`pYn()` in 2.1.220.

The same applies to the `get()`/`isAlias()` case-folding in `server/models.ts`; reverting it reddens only its own new assertions.

I want to flag one thing carefully, because I don't think it was intentional: **the two mixed-case test edits are what make the new code look necessary.** Changing an existing test so that new code has something to satisfy is a pattern worth avoiding — if the case-folding is worth having, it should stand on a new test that describes a real user scenario, in its own PR. I'd look at that PR on its merits; the idea isn't unreasonable, it just isn't this PR.

## Non-blocking

- Three case-fold sites are unpinned (mutation survivors), and the `proxy.ts` one silently falls back to the default route instead of returning 400 — same failure shape as the blocker above.
- The new case-insensitive reserved-alias leak guard is real but untested; deleting it leaves the suite green, and flipping one fixture string pins it.
- `http-proxy/index.ts`'s `unavailableAliases` change is untested.
- Two edits appear to be no-ops: `normalizeModelAliases` in `server/index.ts` (its only consumer already normalizes internally) and `canonicalModelAliasName` on the write side in `server/models.ts`.
- **`CLAUDE.md:100` now describes something that no longer exists** — it cites `aliasNames` in `ServerOptions`, which this delta deletes. No `.md` file is touched in the PR. Please update it; that file is how the next contributor learns these invariants.
- No command tells the user *which* aliases are inactive or why — the skip warning is count-only, so they have to open `config.json`.
- The MITM's 400 suggests running `clodex patch`, which cannot fix a collision or a reserved name. And `clodex patch` itself silently drops rejected aliases with no warning.
- `--unalias` on a colliding pair deletes both entries but reports one; it's also still a dead end for syntax-invalid stored names (`bad:name`, empty), which stay unremovable with a permanent warning. The reserved and mixed-case halves of that complaint are fixed.
- Concurrent `--alias` saves can lose an update (the read happens outside the lock). Pre-existing, no corruption, and the lock itself holds.
- Design question, narrow config only: alias/catalog-id collisions now echo the masked display name, which reintroduces a context-window key mismatch for that case.

## Invariants — all verified sound, no need to revisit

The MITM forwards byte-identical model strings for canonical, lowercase, mixed-case, and all-caps aliases, and passes real Anthropic ids straight through. The endpoint echo matrix holds across masked/unmasked × alias/canonical/exact. `/models` never advertises aliases. The documented resolution precedence is intact and only *grew* — by mixed-case aliases. The `enum == sent == echoed == context-map key` chain holds end to end, and a mixed-case alias round-trips in both modes to the correct upstream model with verbatim echo, including masked gateway ids.

Mutation score is 18/25 overall, and the shape of it is encouraging: **every mutant of the blocker fixes was killed by a well-named test, and every survivor is in the unrequested case-folding surface.** Typecheck, build, commitlint, and the full suite are clean, with no `dist/`, version, or changelog edits.

camjac251:
Addressed the current alias findings in `576294b`.

- canonical names are the only routable aliases; every exact legacy spelling is retained as blocked-only
- malformed containers and elements, reserved names, conflicting targets, unavailable targets, and catalog collisions now fail closed without erasing saved records
- zero-favorite and inactive-target paths cannot fall through to a default or upstream route
- exact provider and model identifiers, repair diagnostics, and removal behavior are preserved
- prototype-like alias names no longer collide with inherited object properties

Verified with 94 focused normalization, routing, diagnostics, and patch regressions, plus type checking, the production build, and a staged secret scan.

camjac251:
Follow-up `58a5e57` fixes the test isolation failure from the latest run.

- shared cleanup now clears both non-streaming response spies alongside the existing streaming spies
- fail-closed routing assertions now measure only the request under test instead of inheriting earlier successful calls

The prior eight-call failure was reproduced locally. The complete router test file now passes 32 of 32 tests; type checking, the production build, and the staged secret scan also pass.

bman654 (approved):
Approving — both asks from the last round are met, and I verified them rather than taking the diff at face value. Thanks for the careful round; this one got noticeably better.

## Verified fixed

**The endpoint-mode blocker is fixed, and it is pinned.** Rejected and unavailable aliases now feed `configuredAliasNames` through `configuredAliasLookupNames` (covering `name`, `savedName`, and `sourceNames`), so `configuredModelUnavailable` is true and the request gets a 400 carrying the reason instead of silently falling through to `defaultRoute`. The route-binding loop also skips unavailable aliases now. I mutation-tested it: filtering `unavailableReason !== undefined` back out of `configuredAliasNames` — which restores exactly the old 200-with-the-default-model behavior — turns a test red. That clears the `CONTRIBUTING.md` bar of "the test for a fix should fail if you revert the fix."

**The case-insensitive request-model matching is genuinely split out.** `src/proxy.ts` and `src/server/models.ts` no longer case-fold inbound model ids; the only remaining `toLowerCase()` calls in the hot path are the pre-existing host and content-encoding normalizations. Canonicalization moved to the configuration boundary via `canonicalModelAliasName`, which is the right place for it and is what this PR is actually about. That also resolves the concern about existing tests being edited to mixed case: `{ name: 'Deep' }` now exercises canonicalization at construction rather than propping up request-side folding.

## One thing you did better than our previous review

The reserved-name list. We had re-derived it against the pristine bundle last round and concluded only `default` and `inherit` were reserved — our sweep looked for `toLowerCase()==="…"` comparisons and missed the array literal. Your expanded list is correct: 2.1.220 has

```js
h1e=["sonnet","opus","haiku","fable","best","sonnet[1m]","opus[1m]","fable[1m]","opusplan"]
function OI(e){return h1e.includes(e)}
```

and because `isReservedModelAlias` strips the `[1m]` suffix before checking, the bare-name set covers the suffixed entries too. Nothing fictional, nothing missing.

Also confirmed: `CLAUDE.md`'s server paragraph is updated, so the `aliasNames` staleness I flagged last round is resolved.

## Non-blocking notes — none of these hold up the merge

1. **Worth a line in the release notes.** Broadening the reserved list is a user-visible behavior change: anyone who previously saved an alias named `opus`, `best`, `sonnet`, or `opusplan` will find it preserved in config but inactive, with a warning. That is the right call — those names could never have routed reliably, since Claude Code resolves them itself — but people will notice.

2. **Malformed `modelAliases` throws a raw `TypeError` on the launch path.** `normalizeModelAliases` throws for a non-array or for any element without a string `name`, and it is called from `catalog.ts` on the `clodex claude` path, plus `patcher.ts`, `server/index.ts`, and `http-proxy/index.ts`. Only the `--alias` / `--unalias` paths in `cli.ts` have the friendly array guard. To be clear this is **not a regression** — `main` crashes on the same input with a worse message (`alias.name` of `null`, or `.map is not a function`) — and it takes hand-edited config to reach. Just noting it as a candidate for the same friendly treatment the CLI paths got.

3. **`target-not-favorite`** is only produced in `patcher.ts`, never in `normalizeModelAliases` itself. That is fine and intentional as far as I can tell, but it means the reason enum is split across two producers; a comment on the type would save the next reader a grep.

## Merge mechanics

This conflicts with #57 in `src/patch-transforms.ts` and `tests/patcher.test.ts`, and with a small patcher change of ours. Since this lands first, the others rebase onto it — see the note I am leaving on #57 about which collision-handling approach should win.

PR #63, opened 2026-07-28
fix(compaction): preflight oversized context requests

cgasgarth:
## Summary

Prevent oversized translated requests when Claude’s estimate and OpenAI’s measured usage drift.

- compact 32k tokens before the configured native threshold as a safety margin
- keep OpenAI usage authoritative after responses; use Claude estimates only for preflight
- add bounded diagnostics for compaction failures, including estimate, threshold, item count, and payload bytes

## Root cause

A failed compaction fell back to forwarding the full transcript, which could then exceed the model context window.

## Validation

- `pnpm test` — 1,242 tests passed
- focused compaction/WebSocket suite — 89 tests passed
- `pnpm build` passed
- `pnpm exec tsc --noEmit -p tsconfig.json` passed


cgasgarth:
Superseded by the correctly targeted fork PR: https://github.com/cgasgarth/clodex/pull/20

PR #75, opened 2026-08-03
feat(patcher): add opt-in local patch extensions

camjac251:
## Summary

- replace the workflow-specific transform with an explicitly enabled local-patch extension
- load one self-contained ES module from `~/.clodex/local-patches.mjs`, or `$CLODEX_HOME/local-patches.mjs`
- include the captured module bytes in patch freshness so edits rebuild from pristine bytes
- apply local sites after the built-in routing patches as one all-or-none transaction
- preserve built-in behavior with exact postcondition checks and a second verifier pass before publishing local output

## Safety model

File presence alone never loads or executes the module. Users opt in persistently with `clodex patch --enable-local-patches` and can disable it with `--disable-local-patches`, which rebuilds from pristine bytes without local sites. Launch freshness checks only read and hash the module; code executes only during patch application.

The loader accepts only a regular UTF-8 file no larger than 1 MiB at the fixed application-home path. It does not search the current directory, packages, dependencies, or `node_modules`. The module runs with the user's full permissions and is not sandboxed.

Each site gets a host-generated `/*clodex-local:<id>*/` marker and reports `OK`, `SKIP`, or `FAIL` through `--trace`. A local load, validation, transform, marker, or built-in-verification failure discards every local change while still publishing the complete built-in patch.

## Verification

- serialized full suite: 1,275 tests passed across 83 files
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`
- Gitleaks scan: no leaks found


bman654 (commented):
Thanks for this — the investigation is solid and the patch itself is well-built. We're going to decline it, on scope rather than quality, but there's a concrete path to getting it shipped that we'd welcome (see the end).

## The symptom is real

We've seen the `0 tok` regression on the workflow board ourselves, so this isn't hypothetical. Verifying your work against a pristine 2.1.220 extraction:

- your anchor regex matches **exactly once**, applies `OK`, emits exactly one marker, and is idempotent (`SKIP: already patched` on rerun)
- `required: false` correctly keeps it from blocking the routing sites, and the missing-anchor case is covered
- `PATCH_TRANSFORMS_VERSION` bump + digest re-pin follow the repo rule
- typecheck clean, 1242 tests / 82 files green — your stated verification reproduces exactly

We also confirmed it does **not** touch workflow budget enforcement (that reads a separate turn-level counter), which was the one behavioral risk worth ruling out.

## Why we're declining: scope

Every existing patch site exists to make clodex models first-class in Claude Code — model validation, `/model` listing, alias resolution, context windows, effort capabilities. This one doesn't. It's a Claude Code harness bug that's independent of clodex: we traced it and clodex cannot produce the zero-usage record.

- clodex floors `input_tokens` with a local estimate on both translation paths, and when a provider omits usage at stream completion it *retains* that estimate rather than zeroing
- Claude Code accumulates usage across `message_start` + `message_delta` (last-non-zero-wins per field) and yields the assistant event once with the merged total, so an all-zero `message_start` can't zero the counter

So the bug reproduces on native Anthropic models too, and carrying the fix here means every clodex user's binary diverges from upstream in a way unrelated to why they installed clodex. Two costs come with that: the transform-version bump forces every existing install through a full ~250 MB re-extract/repack, and because the site is `required: false` against a minified anchor, it will silently start FAILing on some future release with nobody noticing.

## Two suggestions

**1. Please report it upstream.** One finding should sharpen the report considerably: Claude Code *already* guards the final number with the same idiom you're introducing —

```js
no = vt + (Dr || (Ui ? dIe(Ui.message.usage) : 0))   // the "done" report
```

So the completed total is already correct upstream, and the actual unguarded surfaces are narrower than the PR describes: the mid-run progress events (the flicker you're fixing) and the returned `tokens`, which feeds the retry baseline `vt += It.tokens` — an attempt ending at zero drops its tokens from the cross-retry total. That existing guard is also good evidence the maintainers already know `Dr` can be `0`.

Worth including too: Claude Code's own canonical usage reader deliberately *skips* synthetic messages (`model === "<synthetic>"`) and canned interrupt texts when hunting for usage, while the workflow site reads `message.usage` raw. That inconsistency is probably the cleanest framing of the defect, and suggests the upstream fix is to reuse the existing skip-backwards helper rather than to add a `||`.

**2. We would accept a PR that makes the patcher load additional local patches.** If `clodex patch` discovered extra user-supplied patch definitions in a predefined location and applied them as part of the normal patch run, out-of-scope patches like this one could ride the patcher without us owning them. You could then publish your own patch set as a separate repo, or just keep a private set that runs on every patch. That's a feature we'd like to have, and it's the right home for this change.

If you take that on, these are the existing invariants the design has to respect:

- the local patch set must be folded into `computePatchConfigHash`, or an edited local patch leaves the install `current` and silently never re-applies — the same trap `PATCH_TRANSFORMS_VERSION` exists to prevent
- local sites must never block publication of the built-in routing sites, and must never be able to publish a partial patch; they slot in after the built-ins, against the established pristine bytes
- each local site should emit its own marker so double-patching is still detectable, ideally in a namespace distinct from `/*ccpatch:` (that prefix is the *blocking* tier for pristine-backup detection, and a false positive there is unrecoverable)
- local sites should report `OK`/`SKIP`/`FAIL` through `--trace` like the built-ins, so silent decay is visible
- loading transform code from a filesystem location is code execution against the user's Claude Code binary, so it should be explicitly opt-in and not auto-loaded from anywhere a dependency could write to

Happy to review that one.

camjac251:
The workflow-specific transform has been replaced with the general, explicitly enabled local-patch extension described in review.

The requested invariants are covered as follows:

1. **Freshness:** the versioned identity of the exact captured module bytes is folded into `computePatchConfigHash` only while enabled. Disabled installs keep the historical hash shape, and an edit makes the patch state stale.
2. **Built-in publication:** required built-ins run first. Local load, validation, execution, and verification failures remain local `FAIL`s, so the built-in output can still publish.
3. **No partial local output:** the complete local set runs as one transaction against the built-in result. Any failed site restores the pre-local content, marks earlier successes as rolled back, and skips later sites.
4. **Separate markers:** each site receives exactly one host-generated `/*clodex-local:<id>*/` marker. IDs are validated, duplicate or missing markers fail, and local code cannot alter `/*ccpatch:...*/` markers or prior local markers.
5. **Trace visibility:** every site reports `OK`, `SKIP`, or `FAIL` with a reason through `--trace`.
6. **Explicit trust boundary:** only the fixed `$CLODEX_HOME/local-patches.mjs` path is considered after persistent opt-in. File presence alone does nothing; there is no project, package, or dependency discovery, and the docs state that the module is unsandboxed trusted code running with full user permissions.

There is also a post-local integrity gate: exact postconditions are captured for each successful built-in site, then the built-in verifier is rerun before local output publishes. Launch-time freshness checks read and hash the module but do not import or execute it.

Verification completed on commit `738a9fe`:

- serialized full suite: 1,275 tests passed across 83 files
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`
- Gitleaks scan: no leaks found

bman654 (approved):
This is a strong piece of work, and it's the design we asked for. I put a three-lens panel on it
with adversarial verification, and the headline is: **all five invariants we named when we made
the offer hold, and they hold under experiment rather than under reading.** One lens proved them
across 26 runs against the real `runPatchCommand`, and re-checked the proof spans against the real
22.9 MB 2.1.221 bundle for over-matching. No blockers, no majors survived verification.

Specifically, for the record:

| # | Invariant | Verdict |
|---|---|---|
| 1 | Local set folded into `computePatchConfigHash` | **PASS** — hash is over the module *bytes*; editing it moves the install to `stale-config` and rebuilds from pristine (caveat: entry file only, see NIT below) |
| 2 | Never blocks built-ins, never publishes partial bytes | **PASS** — a local patch that throws, hangs, returns a non-string, or deletes the source leaves the built-ins published and never renames partial bytes over the live binary |
| 3 | Own marker namespace, distinct from `/*ccpatch:` | **PASS** — verified both directions: a local marker does not trip the blocking tier of `isPatchedClaudeSource`, and a local patch stays detectable as already-applied |
| 4 | OK/SKIP/FAIL through `--trace` | **PASS** — and better than asked: failures also ride the default (non-`--trace`) summary |
| 5 | Explicit opt-in, never auto-loaded | **PASS, firmly** — presence alone never executes (verified end-to-end), one explicit writer for the flag, no env var enables it, and `CLODEX_HOME` moves the flag with the payload |

The extension point also does the job it exists for: a lens reconstructed your workflow-token-gauge
patch against the real API and applied it to the real 2.1.221 bundle in 85 ms. That was the whole
point of the offer, so it's good to see it land end-to-end.

Everything below is a refinement. Nothing here is a reason not to merge; your call on which to
take now versus later.

## MINOR — a FAILED local set is recorded as `current`, so it is never retried or re-reported

`src/patcher.ts:914-931`, `:798-807`, `:933-936`. When the local set fails, the manifest is still
written with the hash that includes the module digest. Every subsequent `clodex patch` then reports
"already patched … nothing to do" with no LOCAL line at all — the failure becomes invisible and the
user's only recoveries are editing the module or toggling `--disable`/`--enable`. There's no
`--force`.

That's the same silent-staleness shape `PATCH_TRANSFORMS_VERSION` exists to prevent, just reached
from the other direction. Recording the *successful* configuration rather than the attempted one
would close it.

## MINOR — the README's own example idiom takes the whole local set down

`README.md:198` ships:

```js
if (!source.includes(anchor)) return source;
```

But `src/local-patches.ts:235-246` treats "no marker emitted" as FAIL, and `:121-147` then discards
every other local site. So the documented graceful-skip pattern is precisely what detonates the set
the first time a Claude Code update moves an anchor — the user loses their *other*, still-working
local patches too.

Verified: adding one optional site written that way to the working gauge patch discards the gauge
as well. To be fair to the implementation, it is *not* silent — the default summary prints
`SKIP LOCAL gauge — rolled back after LOCAL optional failed` and `FAIL LOCAL optional`, naming
both. And a "not applicable" return *is* expressible today as `return marker + source` (an inert
leading comment). But that idiom is undocumented and non-obvious, while the one in the README is
actively harmful. Either document the marker-only return as the skip idiom, or give the API a real
"not applicable" outcome that doesn't fail the set.

## MINOR — the built-in postcondition proofs check text presence, not reachability

`src/built-in-patch-proofs.ts:16-24`, `:169-176` with `src/patcher.ts:540-572`. The proofs are a
position-blind substring inventory. A local patch that inserts an early `return` before the
`/*ccpatch:ctx*/` map, or relocates a built-in snippet into dead code, keeps every proof string
present: `proofsChanged=false`, the re-verify pass all-SKIPs, `runPatchCommand` exits 0 logging
`OK LOCAL`, and alias routing is dead — `sol` silently reverts to a 200k window. Four such bypasses
were constructed against the real 2.1.221 bundle and driven end-to-end.

I want to be precise about severity, because a verifier pushed back and was right: this is **not a
security boundary being breached.** Local patches are trusted code the user explicitly opted into,
and your README says so plainly — "executes with your full user permissions; not sandboxed. Only
use code you trust." A module that wanted to break routing could simply overwrite the binary. So
this is claim calibration, not a hole: the proofs protect against *accidental* damage from a
well-meaning local patch, which is genuinely useful, but they don't prove the built-ins still
*function*. Worth either widening the proofs to span-based checks (assert the proof text is still
inside the enclosing function that must reach it) or softening the wording so nobody reads them as
a guarantee.

## MINOR — the opt-in default-off state has no test pinning it

Mutating `localPatchesEnabled === true` to `!== false` at both gates (`src/patcher.ts:917`, `:994`)
leaves the entire suite green — 84 files / 1278 tests. Only the explicit-`false` case is covered
(`tests/patcher-command.test.ts:277`, `tests/local-patches.test.ts:312`); the never-opted-in state
— the one invariant 5 exists to protect — is not.

The shipped code is correct: `loadPreferences` applies no default, the flag is written only by the
explicit `--enable`/`--disable` paths (`src/cli.ts:286-293`, `src/patcher.ts:904`), and
`inspectLocalPatchSource` returns before any `statSync` when disabled, so a user who never opted in
never even reads the file. This is test hardening, not a defect — but it's the single property
where a future refactor going wrong would be worst, so it deserves one end-to-end test: module
present on disk, no preference ever set, assert it is neither read nor executed.

## NITs

- **Multi-file local modules escape freshness.** `src/local-patches.ts:290` — the `data:` URL can
  `import` absolute `file://` helpers; editing only a helper leaves the entry-file hash unchanged,
  so `clodex patch` reports `current` and the binary keeps the old local patch forever. You do
  disclose this in the docs, so it's a known edge — mentioning it because it's invariant 1's one
  soft spot.
- **`LOCAL_PATCH_API_VERSION` has no forcing function.** `tests/patcher.test.ts:424-455` pins a
  sha256 of `patch-transforms.ts` precisely so a transform change can't silently skip a version
  bump. `local-patches.ts` now also determines published bytes (marker format, ordering) and has
  its own version constant with no equivalent guard. Same trap, new file.
- **No timeout or cancellation around a local `apply`.** `src/local-patches.ts:207` runs the
  transform synchronously, so an infinite loop hangs `clodex patch` while holding `patch.lock` and
  leaks the ~250 MB `.clodex-patch-*` candidate directory (the `finally` at `src/patcher.ts:781`
  never runs). The live binary is untouched, so this is availability/hygiene only — but the stuck
  lock is what the user will actually feel.
- **`--enable-local-patches` persists before the run can fail.** `src/patcher.ts:903-905` saves the
  preference ahead of the no-favorites bail (`:907-913`) and the lock check (`:938`), so a run that
  exits 1 still arms local execution permanently. Given this is the security-relevant flag, saving
  it only on a successful run would be tidier.

## Verified clean

- Presence alone never loads or executes the module; no env var enables it; `CLODEX_HOME` moves the
  flag together with the payload, so a redirected home can't inherit an opt-in.
- A local failure never aborts the built-ins and never publishes partial bytes.
- The local marker namespace does not collide with the blocking `/*ccpatch:` tier in either
  direction.
- Local sites cannot corrupt the content-addressed pristine backups or launder patched bytes into a
  trusted content-addressed name.
- Proof spans re-checked against the real 22.9 MB 2.1.221 bundle — no over-match.
- `pnpm typecheck` clean.

PR #78, opened 2026-08-05
fix(proxy): isolate bridge settings from child commands

camjac251:
## Summary

- record external and injected proxy, bypass, and CA values at both bridge launch paths
- restore an external value only while the live child value still equals the bridge injection
- bound PATCH 10 to one function and protect its complete emitted postcondition
- preserve settings-level overrides, nested wrapper baselines, and malformed-metadata fail-open behavior
- document the standard child-command boundary and its explicit exclusions

## Why

A value-only snapshot can mistake an ephemeral bridge port for the external proxy during nested launches, leaving later child commands pinned to a dead listener. Compare-before-revert makes the handoff composable: only values written by the active bridge are undone.

## Verification

- `corepack pnpm typecheck`
- `corepack pnpm test` (83 files, 1,307 tests)
- `corepack pnpm build`
- clean 2.1.224 native bundle: one 1,483-byte anchor, 14 source rewrites, one marker, zero references outside the declaring function
- adversarial preceding-function fixtures across patch, backup, and command tests
- `git diff --check`
- `gitleaks git --staged`


bman654 (changes requested):
Thanks for the rework — this is a real problem and `pH()` is the right seam to fix it at. I
confirmed it's Claude Code's shared child-env builder with 14 call sites on 2.1.221 (Bash-tool
shells and the pty path, the login-shell snapshot, stdio MCP servers, LSP servers, hook spawn and
`${VAR}` interpolation, the `gh` env, the subagent status line), so patching it once genuinely
covers the surface. The `PATCH_TRANSFORMS_VERSION` bump plus widening the digest pin to cover
`network-env.ts` is exactly right — that's the trap the pin exists for, and you didn't fall into
it.

Two things need to change before this can land: the PATCH 10 anchor is wrong on the real binary,
and the snapshot's restore-by-value design has a failure mode that outlives the anchor fix.

## BLOCKER — the anchor binds ~407 KB too early and publishes a broken binary

```
/(function [\w$]+\(\)\{)((?=[\s\S]{0,5000}?CLAUDE_CODE_REMOTE)(?=[\s\S]{0,7000}?CLAUDE_CODE_OAUTH_TOKEN)[\s\S]*?)(\}function [\w$]+\(\)\{let [\w$]+=process\.env\.CLAUDE_CODE_MCP_ALLOWLIST_ENV)/
```

JS regex prefers the **leftmost** start position. Lazy `[\s\S]*?` minimises the body *for a given
start*; it never makes the engine prefer a later `function X(){`. The lookaheads only constrain
the first 5,000 / 7,000 chars after the opening brace — nothing keeps the body inside that
function. So the engine locks onto the first zero-arg function that has those two tokens nearby
and runs the body across every function between it and the MCP-allowlist tail.

Measured on the real pristine **2.1.221** bundle (extracted with `tryDetectInstallation` +
`readContent`, then run through this PR's own `applyClodexPatches` — not a re-implementation):

| | |
|---|---|
| head group binds to | `function mae(){` — **406,863 chars before** the intended `function pH(){` |
| match length | 408,346 chars, containing 989 `function ` declarations |
| `_clodexChildEnv` refs emitted | 132, across **56 distinct functions** |
| extent of the declaring function `mae` | 847 chars |
| refs **outside** that binding's scope | **125** |
| PATCH 10 status / markers | `OK` / exactly 1 |
| `applyOnce` "expected exactly 1" guard | passes — there really is one match |

Each out-of-scope reference is a free identifier. First one:

```js
function aEu({envVar:e,wellKnownPath:t,label:r,getCached:n,setCached:o,skipInReviewOrigin:i=!1}){
  let s=n(); if(s!==void 0) return s;
  let a=_clodexChildEnv[e];   // ReferenceError: _clodexChildEnv is not defined
```

Every site reports OK, one marker is emitted, the repack succeeds, and candidate-and-rename
publishes it. `clodex patch` would hand the user a **broken Claude Code binary with a clean
`--trace`**.

**A fix that works** — forbid the body from crossing a function boundary, which forces the head
group to be the immediately-preceding declaration:

```diff
-)(?=[\s\S]{0,7000}?CLAUDE_CODE_OAUTH_TOKEN)[\s\S]*?)
+)(?=[\s\S]{0,7000}?CLAUDE_CODE_OAUTH_TOKEN)(?:(?!\}function )[\s\S])*?)
```

| variant | matches | head | body | `process.env` rewritten |
|---|---|---|---|---|
| as submitted | 1 | `function mae(){` | 408,268 | 125 |
| with the guard | 1 | `function pH(){` | 1,406 | 14 |

**Your test harness would already have caught this — the fixture just lacks an adversarial
input.** `executeChildEnv` extracts and executes only the `function childEnv(` line, so a
mis-bound injection leaves `_clodexChildEnv` free and the test throws. Adding one line to
`CLAUDE_CORE_FIXTURE` immediately before the `childEnv` entry:

```js
'function cwdOf(){let p=process.env.PWD;return p}',
```

turns the suite red — **3 failed / 81 passed**, all three PATCH 10 tests with
`ReferenceError: _clodexChildEnv is not defined`. Please add that decoy (the same line exists in
the `tests/patch-backup.test.ts` and `tests/patcher-command.test.ts` fixtures). It's the cheapest
possible guard against this whole class.

More generally: "applied once and emitted one marker" cannot distinguish a correct match from a
catastrophic over-match — both produce exactly those numbers. For any bundle regex, assert on the
**matched span** and on the enclosing function of every rewritten reference.

## MAJOR — under `clodex claude --proxy`, the snapshot records the bridge as the "original" and pins a dead port

`buildHttpProxyChildEnv` (`src/env.ts:100`) sets `HTTPS_PROXY=http://127.0.0.1:<ephemeral MITM
port>` plus the clodex CA, and writes no snapshot. `preserveOriginalNetworkEnv`
(`src/wrapper-env.ts:31`) can't tell that the env it's snapshotting is already bridged. All the
ingredients are the documented setup — `docs/background-agents.md:39` exports
`CLAUDE_CODE_PROCESS_WRAPPER` in the **shell profile**, so it's set in the same shell you run
`clodex claude` from, and clodex never strips it from the child env.

1. `clodex claude --proxy` launches claude with `HTTPS_PROXY=http://127.0.0.1:51234`, no snapshot.
2. A background session spawns; the launcher passes `{...process.env}` and keeps the wrapper var,
   so `clodex-claude` runs.
3. `computeWrapperEnv` sees no valid snapshot and records
   `{"HTTPS_PROXY":"http://127.0.0.1:51234", …}` as the ORIGINAL, then overwrites the live vars
   with the standalone server's port.
4. PATCH 10 now "restores" the ephemeral port for every Bash command, MCP server and hook in that
   session.

So the bug isn't fixed there — and once the foreground `clodex claude` exits, port 51234 is gone
and every child command fails with `Proxy CONNECT aborted` **permanently**, surviving a restart of
the standalone server, because `hasValidOriginalNetworkSnapshot` preserves the poisoned snapshot
down every further level of nesting. (Preconditions: a discoverable live standalone server and an
applied PATCH 10.)

This is the sharp edge of a more general point: **restoring by value — "what the env looked like
before the wrapper ran" — is not the same as undoing clodex's own writes.** A restore that only
reverts a variable when its *current* value still equals what clodex injected (the bridge URL, the
clodex CA path) is sound and closes this. The same design gap shows up more mildly with
settings-sourced env (below).

Related and worth deciding explicitly: `computeWrapperEnv` is called only from
`src/claude-wrapper.ts:135`, so the default `clodex claude --proxy` path never writes a snapshot
and PATCH 10 no-ops there — the leak this PR exists to fix is still present on the default path.
Adding the snapshot to `buildHttpProxyChildEnv` closes both that and the poisoning above, but note
the per-session MITM deliberately doesn't register in `~/.clodex/server-runtime.json`, so a nested
wrapper launch under it finds no server and leaves the env untouched. Those two facts have to stay
consistent.

## MINOR — settings-sourced proxy/CA is stripped from children

Claude Code applies `settings.json` `env` into `process.env` at startup —
`Object.assign(process.env, phr(<settings>.env, scope))` per scope — but the recorder `bV_` keeps
only `NO_COLOR`/`FORCE_COLOR`, so those values get no second chance in `pH()`. With a snapshot of
`{}` and a settings-level `HTTPS_PROXY`/`NODE_EXTRA_CA_CERTS`/`NO_PROXY`, the patched builder hands
children `{}`.

Honest reachability, which is why this is MINOR not MAJOR: in **proxy** mode that same
`Object.assign` also clobbers clodex's own MITM URL and CA, so the bridge is already dead in that
configuration and nobody is running there. It needs an endpoint-mode wrapper session to bite fully.
Shell-sourced proxy — the normal corporate setup — restores correctly. Still worth fixing via the
compare-before-revert approach above rather than special-casing.

## MINOR — scope gaps and behaviour changes worth acknowledging

- **`--bg --exec` bypasses `pH()` entirely.** The background pty-host env is built from
  `{...process.env}`, and its deletion lists carry no proxy/TLS key, so `claude --bg --exec 'npm
  ci'` still gets clodex's proxy and CA — the motivating failure is unfixed there. Separately,
  `CLODEX_ORIGINAL_NETWORK_ENV` doesn't start with `CLAUDE_`, so it survives the exec-mode scrub
  and is exported into arbitrary user shell commands where PATCH 10's `delete` never runs.
- **A nested `claude` started from the Bash tool now loses clodex routing.** The process-wrapper
  var is consulted only by Claude Code's own supervisor/pty spawns, so a `claude -p --model sol`
  run from a Bash call, hook, or skill script is an ordinary `pH()` child: after PATCH 10 it gets
  the original network env and the snapshot var is deleted, so nothing downstream can recover the
  route. It reaches `api.anthropic.com` and fails on an unknown model, where before it worked.
  The `clodex-claude` spelling still works — but that's a behaviour change the description doesn't
  mention, and the docs currently present the wrapper as needed only for terminal sessions and the
  process wrapper.
- **Endpoint-mode bridge vars aren't isolated.** `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY`
  are injected by `computeWrapperEnv` (`src/wrapper-env.ts:103-104`) but neither is snapshotted nor
  restored, and `pH()` only scrubs the credential list under sandbox/`local-agent` entrypoints. So
  a Bash-tool `curl "$ANTHROPIC_BASE_URL/v1/messages"` still lands on clodex's gateway. Pre-existing,
  not a regression — but the PR's contract is "isolate bridge settings from child commands" and it
  isolates one of the two bridges. Fine to scope out; worth saying so in the comment at
  `src/patch-transforms.ts:548`.

## MINOR — the injected validation guards have no behavioural coverage

Mutation runs against the new tests:

| mutation | caught |
|---|---|
| drop `else delete _clodexChildEnv[key]` | yes |
| drop `delete _clodexChildEnv[snapshotVar]` | yes |
| drop `&& typeof _clodexNetwork==="object"` | **no** |
| drop `&& !Array.isArray(_clodexNetwork)` | **no** |
| `if(typeof _clodexNetwork[key]==="string")` -> `if(true)` | **no** |

The three misses are the entire defensive layer of the injected JS. Only the digest pin fired, and
its documented remedy is "re-pin the digest" — the reflex that lets a weakened guard ship. Note the
asymmetry: `hasValidOriginalNetworkSnapshot` (the producer) has three negative tests for the same
conditions; the consumer inside the binary has none. Three cases through `executeChildEnv` would
close it: snapshot `'[]'`, snapshot `'null'`, and `{"HTTPS_PROXY": 42}` — each must leave the
bridge vars deleted rather than injecting a non-string.

## NITs

- **`required: true` on PATCH 10.** I initially thought this was a bigger deal and checked it
  properly, so for the record: it is *not* anomalous — PATCH 1/3/6/7 are also `required: true` and
  throw from inside the transform, and inserting a function between the builder and the MCP reader
  does **not** cause a 0-match (the unbounded-lazy body just swallows it; only renaming
  `CLAUDE_CODE_MCP_ALLOWLIST_ENV` gives 0 matches). Failure is loud and candidate-and-rename leaves
  the install intact. So this is only a design question — whether a network-hygiene site should be
  able to abort the model-identity sites — worth a thought, not a change request.
- **Commit 2's rethrow.** `catch(e){if(!(e instanceof SyntaxError))throw e}` sits in the shared
  child-env builder. `JSON.parse` on a string can only throw `SyntaxError`, so it's unreachable —
  but if it ever fired it would take out every Bash call, MCP/LSP spawn and hook at once, with no
  clodex attribution. Fail-open is the better default in that position. Same note applies to the
  non-`SyntaxError` rethrow in `hasValidOriginalNetworkSnapshot`, which propagates out of
  `computeWrapperEnv` into the launch path, against the wrapper's "never break launching claude"
  invariant.
- **CLAUDE.md** still says `applyClodexPatches` applies "the PATCH 1–9 regex sites" and that the
  pin hashes `patch-transforms.ts` (now two files). The new `network-env.ts` module and the
  `CLODEX_ORIGINAL_NETWORK_ENV` contract between `computeWrapperEnv` and the binary aren't recorded
  either — that bullet list is how the next contributor learns these invariants.
- **Two `fix(proxy):` commits** produce two changelog lines, the second describing a bug in code
  that never shipped. Worth squashing or rewording as `chore:`.

## Note on the stack

#79 carries the identical anchor and the same `body.replace(/process\.env/g, …)`, so it inherits
the blocker — worth fixing here at the base.

## Verified clean (so you don't re-derive it)

- `pH()` is the right target; 14 call sites, all genuine child-spawn paths.
- The `utn()`/allowlist branch is correctly left alone — `KIs()` copies only
  `["HOME","LOGNAME","PATH","SHELL","TERM","USER"]`, so there's no proxy leak there to restore.
- The `return process.env` -> copy rewrite is safe: all 14 consumers treat the result as read-only.
- The restore isn't undone later in `pH()` — none of its deletion lists contain a proxy or CA var.
- `computeWrapperEnv` doesn't mutate its input, and the no-server early return can't leak the
  snapshot var.
- `_clodex*` identifiers appear nowhere in the 22.9 MB bundle, so no collision or shadowing.
- The marker carries the blocking `/*ccpatch:` prefix, so pristine-backup detection still refuses
  a PATCH-10-only binary.
- Version bump and digest widening are correct and complete.
- The premise holds: the MITM does tunnel non-intercepted CONNECT (`src/http-proxy/server.ts:938`),
  so leaked child traffic normally works — it fails when the MITM goes away (`close()` destroys
  every tracked socket), which matches the `Proxy CONNECT aborted` you saw.

camjac251:
Review update:

- Rebased onto current `main` and replaced value-only snapshots with a versioned compare-before-revert contract.
- Both bridge launch paths now record external and injected values; nested launches recover the external baseline without preserving an ephemeral port.
- Settings-level overrides remain authoritative, and malformed metadata fails open while being removed from ordinary child environments.
- PATCH 10 cannot cross whitespace-separated function boundaries; adversarial preceding-function fixtures cover all three patch harnesses.
- Clean 2.1.224 native-bundle verification found one 1,483-byte match, 14 source rewrites, one marker, and zero emitted references outside the declaring function.
- The separate `--bg --exec`, endpoint-variable, and plain nested-client boundaries are documented explicitly.

Verification: typecheck; build; 83 files / 1,307 tests; diff check; staged leak scan.

bman654 (changes requested):
Re-reviewed at head `9643799`, covering both the binary patcher and the environment-isolation design.

**Both blocking objections from last round are fixed, and the second one is fixed the right way.** I verified each rather than taking the commit titles' word for it. Coverage went from "guards with no behavioural tests" to a clean sweep. There is one ask left, on the patcher, and it comes with a verified drop-in replacement — details in §3.

---

## 1. The anchor over-match is resolved — verified on 12 real bundles

I extracted the JS out of every pristine binary in `~/.tweakcc/` and ran this PR's real `applyClodexPatches` over each, measuring the **matched span** and the **enclosing function of every rewritten reference** — because "applied once, one marker" reads identically for a correct match and a catastrophic over-match, so counts alone prove nothing.

Versions: 2.1.208, .209, .210, .211, .214, .215, .217, .218, .219, .220, .221, .224 (19.6–23.7 MB of JS each).

| version | old head / span / rewrites | new head / span / rewrites |
|---|---|---|
| 2.1.221 | `mae` / 408,346 chars / 125 | `pH` / **1,483** / **14** |
| 2.1.224 | `$_o` / 239,877 / 67 | `XH` / 1,483 / 14 |
| 2.1.220 | `Woe` / 397,202 / 123 | `MO` / 1,299 / 12 |
| 2.1.208 | `tee` / 369,035 / 207 | `pL` / 1,302 / 12 |

The other eight are the same shape — the old anchor over-matched on **all 12**, not only .221. With the new anchor, on every one: exactly 1 global match, span 1,299–1,483 chars, **0 nested function declarations in the body** (was 528–988), and the enclosing function of every `_clodexChildEnv` reference in the output is the patched function itself (`refs=21..23, outside=0`).

I also **executed** the patched builder extracted from each bundle, stubbing only the bundle's own helpers. The identifier-discovery loop found 8 undefined identifiers per version, all genuine bundle helpers, with `_clodex*` never among them — the direct refutation of the old failure, which left 125 free `_clodexChildEnv` identifiers on 2.1.221.

**Your decoy fixture does its job.** Mutating the anchor back to the pre-fix body and running the file the way CI runs it:

```
CLODEX_HOME=$(mktemp -d) pnpm vitest run tests/patcher.test.ts
#   Tests  9 failed | 83 passed (92)
#   ReferenceError: _clodexChildEnv is not defined
```

Nine tests fail with the exact production symptom. Before the decoy line, that same mutation was green. Thank you for adding it.

Also confirmed: `PATCH_TRANSFORMS_VERSION` bumped 3 → 4 with the pinned digest widened to cover `network-env.ts`; `/*ccpatch:child-network-env*/` is picked up by the blocking-marker prefix; PATCH 10 is captured in the built-in proof set; and publication safety holds — a required-site failure throws before `writeContent`, and I hit that path for real without the input bundle being touched.

## 2. The ephemeral-port pinning is fixed — verified, not read

`9643799` replaces the one-sided snapshot with a two-sided `original` + `injected` contract and reverts a key only while its live value still equals what clodex injected. That is exactly the compare-against-what-we-injected shape, and it holds.

I drove the whole chain rather than reading it — foreground per-session MITM on an ephemeral port, wrapper on top, a second nesting level, then the emitted builder executed for real out of the patched bundle:

```ts
const foreground = buildHttpProxyChildEnv(54321, '/tmp/session-ca.pem', EXTERNAL);
const wrapped    = computeWrapperEnv(foreground, standalone(17645));
const wrapped2   = computeWrapperEnv(wrapped,    standalone(17999));

expect(JSON.stringify(contract(wrapped2).original)).not.toContain('54321'); // dead session port
expect(contract(wrapped2).original['HTTPS_PROXY']).toBe('http://alice:s3cret@corp-proxy:8080');
```

To confirm the harness discriminates, the identical scenario against pre-fix `aae0253`:

```
FAIL  OLD head aae0253 > pins the ephemeral MITM port as the "original"
  Received: {"HTTPS_PROXY":"http://127.0.0.1:54321", …}
```

Same inputs, same path: dead port pinned before, external baseline after.

Two more checks on the new design, both clean. **Third-party mutation mid-session:** if something changes a value after injection the contract no longer matches, the value is left alone, and the next layer re-baselines — per-key, so a meddled key and an untouched one behave independently. **Settings-sourced proxy/CA** (my MINOR last round) is now correct for the right reason: Claude Code applies settings `env` via `Object.assign(process.env, …)` *after* any wrapper snapshot, so at builder time the live value differs from `injected` and compare-before-revert leaves it alone.

**The parse-error rethrow is gone.** I was going to flag `1b32882`, since a wrapper that throws breaks every agent spawn — but `9643799` removed that rethrow entirely, and the contract variable is deleted from the child *before* the `try`. I fed 11 malformed/hostile contract values through all three consumers end to end (not-JSON, unterminated, `[]`, `null`, `__proto__` payloads in both positions, wrong `version`, asymmetric keys, a 200 KB value): none throws, none pollutes `Object.prototype`, all fail open. And mutating the catch into a rethrow is now caught by tests.

**Coverage: 19/19 mutants killed** (was 3 of 8 surviving). Every guard in both the injected JavaScript and `network-env.ts` is individually load-bearing under test — `hasOwnProperty`, `Array.isArray`, `version`, key-symmetry and the type guards all pinned. Stronger than I asked for.

**Also checked, so you don't re-derive it:** PATCH 10 cannot un-bridge Claude Code's own background sessions. That was my main worry about reverting inside the shared builder. I enumerated all 14 call sites on the real 2.1.221 bundle — stdio MCP (×2), LSP, hook env, spawn probe, shell snapshot, Bash exec (×2), `gh`, HTTP-hook interpolation, hook spawn, eval harness, status line. The background-agent supervisor, pty and daemon-worker spawns build their env independently and never call it. Separately, the `process.env` → `_clodexChildEnv` rewrite is lossless: zero writes to `process.env` in the matched body across four real bundles.

---

## 3. The one ask — identify PATCH 10's target by its own code, like every other patch

This is the only thing I'd like changed before merge. It is not a live defect: the patch is correct on every build that exists today. It is that PATCH 10 is the **only** patch in the file that doesn't identify its target by the target's own content.

Every other site pins the code it edits — PATCH 1 the literal enum array, PATCH 3 the alias list, PATCH 6 `case"best":{…}`, PATCH 8a `function X(y){if(z(y))return!1;let a=b(y,"effort")`. PATCH 10 matches `function X(){` … `} function Y(){let z=process.env.CLAUDE_CODE_MCP_ALLOWLIST_ENV`, so the only specific landmark belongs to the *next* function, which it never modifies. The target is whatever zero-argument function happens to sit before it.

The two lookaheads look like they constrain the target, but they don't: they're asserted from the match start over 5,000/7,000-character windows and can be satisfied by text outside the body. Measured on the real bundles, `CLAUDE_CODE_REMOTE` occurs **116–143 times** and `CLAUDE_CODE_OAUTH_TOKEN` **81–90 times**, so as identifiers they carry almost no information.

Three mutations of the real 2.1.221 bundle:

| mutation | current anchor |
|---|---|
| a decoy function mentioning `CLAUDE_CODE_REMOTE` inserted between the builder and the MCP reader | **OK, marker emitted, bound to the decoy** — builder unpatched, isolation silently no-ops, `clodex patch` reports success |
| same, mentioning neither token | **FAIL** → `PatchApplyError` → the whole run aborts for every user (the 2.1.224 / PATCH 1 shape) |
| a nested `function` declaration at the top of the builder body | **OK**, but binds to the inner function — the declaration lands inside it while the rewrites stay in the builder, so the original `ReferenceError` is back, with OK status, one marker, and it publishes |

`}\s*function ` blocks a *preceding* function from swallowing the builder, but not one declared *inside* it.

**A stable signature exists, and I measured it.** The builder is remarkably static across all 12 bundles — only three distinct body lengths (1225 / 1222 / 1406), the single refactor in that window (.220→.221) changed only the middle, and no version's body contains a nested `function`. Several structural signatures from the target's own code match **exactly once in the entire 19–23 MB bundle** in every version; the strongest is its own final statement, whose `` `INPUT_${x}` `` template occurs once bundle-wide.

```js
/(function [\w$]+\(\)\{)(let [\w$]+=[\w$]+\(\),[\w$]+=Object\.keys\([\w$]+\)\.length>0,[\w$]+=Object\.keys\([\w$]+\)\.length>0,[\w$]+=[\w$]+\(process\.env\.CLAUDE_CODE_REMOTE\)\?(?:(?!\}\s*function )[\s\S])*?for\(let [\w$]+ of [\w$]+\)delete [\w$]+\[[\w$]+\],delete [\w$]+\[`INPUT_\$\{[\w$]+\}`\];return [\w$]+)(\})/
```

Pair it with an assertion in the replacement callback — PATCH 1/3/6/7 already throw from inside their transforms, so this is in keeping:

```ts
const REQUIRED = ['CLAUDE_CODE_REMOTE','CLAUDE_CODE_OAUTH_TOKEN','CLAUDE_CODE_SUBSCRIPTION_TYPE',
                  'CLAUDE_BG_PTY_AUTH','"OTEL_"','CLAUDE_CODE_OTEL_DIAG_STDERR'];
const ok = REQUIRED.every(k => body.includes(k)) && !/\bfunction [\w$]*\(/.test(body);
```

The group structure is identical to what your callback already expects (`head` / `body` / `tail`), so the replacement body needs no change.

Measured on all 12: exactly 1 match, same function, byte-identical body, same rewrite counts (12 or 14, matching the current anchor). Swapped into the real `applyClodexPatches` and run over every bundle with a representative favorites config:

```
2.1.208 … 2.1.224:  old=OK  new=OK  identicalOutput=TRUE  markers=1  allResultsSame=TRUE
```

**Byte-identical patched output on every real binary**, and against the three mutations above it binds correctly in the first two and fails *loudly* in the third instead of publishing a broken binary. So this costs nothing on any build that exists and removes the silent-wrong-target mode on builds that don't exist yet. Given the site is `required: true` and rewrites the user's binary, I think it's worth taking now.

Worth encoding as a second fixture too: a builder variant containing a nested function declaration, asserting PATCH 10 fails loudly rather than binding to it. The current fixture proves the leftmost-start case and nothing else.

---

## 4. Smaller notes

**The `CLAUDE_CODE_` prefix puts the contract variable into Claude Code's startup telemetry.** On .221 and .224 the bundle collects every set variable whose name starts with `CLAUDE_CODE_` or `ANTHROPIC_` and reports the **names**:

```js
function MQm(e=process.env){let t=[];for(let r in e)
  if((r.startsWith("CLAUDE_CODE_")||r.startsWith("ANTHROPIC_"))&&!hZT.has(r)&&e[r]!==void 0&&e[r]!=="")
    t.push(r); return t.sort()}
hZT = new Set(["CLAUDE_CODE_ENTRYPOINT"]);   // exclusion list holds exactly one entry
```

Only the name is sent, never the value — so this is namespace hygiene, not disclosure. A `CLODEX_`-prefixed name sidesteps both the telemetry line and any future meaning Anthropic assigns to the prefix. Was the `CLAUDE_CODE_` prefix load-bearing for something I've missed?

**The contract embeds the external proxy URL verbatim, credentials included.** The design needs the value, and the shared builder strips the variable on every path it covers, so exposure is confined to the paths you already documented (`--bg --exec`, anything spawning from raw `process.env`). It isn't a privilege boundary crossing — it's the user's own environment. But it is a change in kind: before this PR a proxy-mode child saw only the local bridge URL, so the secret wasn't present at all, and it now sits in a variable name no redaction tooling recognises. One sentence in `docs/background-agents.md` would close it out.

**The isolation is proxy-mode-only.** You've documented this and deferring is right. Naming it precisely: endpoint mode sets `ANTHROPIC_BASE_URL`/`ANTHROPIC_API_KEY` on the child and neither is in the contract, so a Bash-tool child in endpoint mode can still inherit a gateway URL that outlives its gateway — the same shape this PR just fixed, on a different variable set. Much lower reachability, so a follow-up rather than scope here.

**NIT — the transform rewrites only the literal `process.env` spelling.** Inside the patched function four reads already go through an env accessor (`re.CLAUDE_BG_PTY_AUTH` etc.). Today every one is a `CLAUDE_BG_*` variable clodex never touches, so nothing is wrong. But if a future build routes one of the seven network variables through that accessor, `body.replace(/process\.env/g, …)` misses it and the site still reports OK with one marker — the same trap as before. Asserting the body contains no non-literal read of `HTTPS?_PROXY` / `NO_PROXY` / `NODE_EXTRA_CA_CERTS` would make that fail loudly.

**NIT — the contract has two implementations and no differential test.** `networkEnvBaseline()` and the injected JS both parse the same contract and validate differently (the TS side enforces an allowlist and key symmetry; the injected JS ignores unknown keys and checks `hasOwnProperty` per key). Both inputs are clodex-produced today so there's no live divergence, but the fixtures hand-write the expected outputs on each side, so the coupling is untested by construction. Feeding the same contract shapes to both and diffing would pin them cheaply.

---

## Checks

```
CLODEX_HOME=$(mktemp -d) pnpm typecheck   # clean
CLODEX_HOME=$(mktemp -d) pnpm test        # 83 files / 1307 tests passed
CLODEX_HOME=$(mktemp -d) pnpm build       # success, dist/cli.js 2.80 MB
```

One caveat: `tests/http-proxy-index.test.ts` fails for me in the full-suite run only because my own shell has `HTTPS_PROXY`/`NODE_EXTRA_CA_CERTS` set. With those scrubbed the suite is 1307/1307. Untouched by this PR and CI is green — a local artifact, not a regression.

---

## Verdict

Both blocking objections are genuinely resolved, and the evidence for each is the kind that would have caught the original bug. The redesign is the correct shape, extending it to `buildHttpProxyChildEnv` closed the gap where the whole thing was inert on the default launch path, and 19/19 mutants killed is a real improvement.

The one thing I'd like before merge is §3 — make PATCH 10 identify its target the way the rest of the file does. The replacement is measured byte-identical on all 12 real binaries, so it's a no-op today and a loud failure instead of a silently broken install tomorrow. Happy to re-run everything above against it.

camjac251:
Updated at `65581ca` and rebased onto current `main` (`2.3.0`).

PATCH 10 now identifies the child-environment builder from its own initialization and final `INPUT_${...}` cleanup/return shape. The replacement callback also requires the known body landmarks and rejects nested functions before any patched candidate can be published.

The test fixture now distinguishes all three wrong-target classes discussed in the review:
- a preceding zero-argument decoy;
- a token-bearing zero-argument function between the builder and the next reader;
- a nested function inside an otherwise matching builder, which fails loudly.

Verification:
- `pnpm typecheck`
- 84 test files / 1,330 tests
- `pnpm build`
- diff check and staged leak scan
- required checks: `test` and `commitlint` both successful

bman654 (changes requested):
The one ask from last round is done, and the transform is correct — I executed it rather than reading it. One test-coverage gap is worth a round before this lands.

> **What we are actually requesting: the single test under "The one ask before this lands".** That is the only thing blocking a merge.
>
> Everything under "Minors" is optional and listed only in case you would rather sweep some of it while you are in here. None of it needs to happen in this PR, and we will not hold the merge on any of it — say the word and we will move whichever ones you skip to our own backlog.

## PATCH 10 now identifies its target by its own code — verified

Ran the PR's real `applyClodexPatches` over **13 pristine bundles, 2.1.208 → 2.1.226**:

| | result |
|---|---|
| anchor matches per bundle | **exactly 1**, all 13 |
| bound function | the child-env builder every time (`pL`/`vL`/`PL`/`ML`/`fM`/`PO`/`MO`/`pH`/`XH`/`$M`) |
| matched span | 1237–1421 bytes |
| `process.env` rewrites | 21–23, **0 escaping the declaring function** |
| patched 24 MB bundle re-parses (`vm.Script`) | yes |

The three wrong-target classes, applied to the **real 2.1.226 bundle** rather than the fixture: a decoy inserted before the builder → still binds the real builder; the *next* function's `CLAUDE_CODE_MCP_ALLOWLIST_ENV` renamed → **still binds**, which is the adjacency dependence gone; a nested named function → throws, publishes nothing. Swapping the old anchor back in fails your `token-bearing function follows it` fixture, so it discriminates.

I also extracted the patched `$M` from real 2.1.226 and **executed** it. It reverts correctly on both the early-return and the merge branch, still scrubs credentials, leaves a value another layer changed authoritative, and 12 hostile contract values never throw, never revert, and never leak the contract var. With no contract it returns the **same `process.env` object reference**, so unpatched behaviour is preserved exactly; with a contract it returns a copy, and none of the 14 `$M()` call sites mutate the result, so that difference is inert.

## The one ask before this lands — a missing test, not a bug

**To be clear up front: the transform is correct on both of the builder's exits.** It rewrites every `process.env` in the body, and I confirmed by execution that the merge path reverts properly. Nothing below says the shipped patch leaks.

What is missing is a test that would *notice* if that stopped being true. The builder has two exits:

```js
if (!t&&!o&&!i&&!l&&!s&&!a.length&&!c&&!r) return process.env;   // exit 1: early return
let u = {...process.env, ...tTs, ...e, ...n};                     // exit 2: merge path
```

Only exit 1 is covered. I mutated the transform to rewrite just the final `return process.env`, leaving the merge path on the live object — which would be a real bridge leak, the child inheriting clodex's MITM proxy and CA — and the suite came back **1329/1330**. The single failure was the `patch-transforms.ts` sha256 pin, which fires for *any* edit to the file and says nothing about behaviour. Worse, CLAUDE.md tells contributors to re-pin that digest after a comment-only edit, so the one red signal is one they are instructed to clear.

This matters because the merge path is not the rare one. `$M` takes it whenever any of `t/o/i/l/s/a/c/r` is set — including `jP_()` (entrypoint `local-agent`) and the `Yin` session variables — which is exactly the background-agent case this PR documents.

Poisoning the fixture's merge path directly (`tests/patcher.test.ts:651`) also leaves 94/94 patcher tests green, confirming the fixture never exercises it.

**Ask:** one `executeChildEnv` case that forces the merge branch and asserts the network variables are reverted there too.

## Minors

1. **A `te.X` migration would silently disable this patch.** The rewrite only rebinds the literal `process.env`, and 2.1.226 has begun moving these very variables to the live accessor: `te.NODE_EXTRA_CA_CERTS` appears **8 times in 2.1.226 and 0 times in 2.1.224**, `te.HTTPS_PROXY` 1 and 0. If that migration reaches this function's body, PATCH 10 keeps reporting OK while quietly reverting nothing. A required-literal assertion on the `process.env` spread, or rewriting `te.` reads of the seven contract variables too, would turn that into a loud failure instead of a silent leak.

2. **The TypeScript and baked-JS contract readers disagree.** A differential run over 4,000 generated contract/environment pairs found 165 divergences: the host-side `parseNetworkEnvContract` is stricter and rejects contracts the in-binary reader accepts, which re-creates the dead-bridge-port case the two-sided design was meant to remove. Worth one differential test so the two implementations cannot drift further.

3. **Contract variable name.** Confirmed on 2.1.226 that `Ckh(e=process.env)` collects every name starting `CLAUDE_CODE_`/`ANTHROPIC_` and `tengu_startup_telemetry` ships `set_env_vars: <names>.join(",")`. Names only, no values — a `CLODEX_` prefix keeps clodex out of that field for free.

4. **The contract carries the external proxy URL verbatim, credentials included**, into the child environment. Before this PR a proxy-mode child never saw the user's real `HTTPS_PROXY`. The patched builder strips it before grandchildren, but the claude process holds it and the `--bg --exec` path you document as uncovered does not strip it. A docs line, or a mode-0600 file plus a path, would close it.

5. **`required: true` on PATCH 10 means an unrelated upstream refactor of this one function disables the entire clodex patch**, PATCH 1–9 included, rather than degrading to no child-env isolation. Given how specific the anchor now is, `required: false` with a loud warning may be the better failure mode — worth a deliberate decision either way.

6. **Nit:** the nested-function guard `/\bfunction [\w$]*\(/` requires a space after `function`, so a minified anonymous `function(){}` evades it and the `}\s*function ` temper — injecting one into real 2.1.226 still patches OK. Harmless today (the rewrite stays lexically correct; the real body already contains arrows over the rewritten identifier), so `/\bfunction\s*[\w$]*\(/` is just tidier.

camjac251:
Addressed the remaining merge blocker and the related contract-hardening points at head `21763dc`, rebased onto upstream 2.4.0 (`ad907a5`).

- Added a behavioral regression that forces the child-environment builder's merge branch with a nonempty extra environment, then proves the original proxy and CA values survive while unrelated values merge.
- Made the patch fail closed if the upstream builder no longer spreads `process.env`, and covered both named and anonymous nested-function drift.
- Tightened the baked contract reader to validate both records symmetrically, reject unknown keys and invalid values, and require identical key sets. Differential fixed-case tests now compare it with the host reader.
- Documented why this site remains required and documented that the parent contract can contain a credential-bearing external proxy URL. The separate `--bg --exec` path remains explicitly out of scope.
- Kept the existing contract key for upgrade compatibility. Renaming it in one release can create a transient wrapper/binary version mismatch that silently disables isolation; that change needs a phased dual-read/dual-write migration.

Verification on this head:

- focused environment/patch tests: 168 passed
- full suite: 85 files, 1,394 tests passed
- typecheck and production build passed
- documentation check and staged secret scan passed
- hosted `test` and `commitlint` checks passed

bman654 (approved):
**Approving.** The blocking ask from the previous round is closed, verified by re-running the exact mutation that exposed it. Nothing blocks merge.

Everything under **Optional items** is explicitly optional, including OPTIONAL 0, which was raised as a blocker during review and then demoted — the reasoning is recorded there so it is not re-derived later.

## Prior asks

| From last round | Status |
|---|---|
| **BLOCKING** — one `executeChildEnv` case forcing the merge branch | **ADDRESSED** (`tests/patcher.test.ts:1342-1370`) |
| Minor 1 — a `te.X` migration would silently disable the patch | **PARTIALLY ADDRESSED** — total drift now fails closed, partial drift does not (see OPTIONAL 1) |
| Minor 2 — host and baked contract readers disagree | **ADDRESSED** on the baked side; the host side is still unpinned (see OPTIONAL 2) |
| Minor 3 — contract variable name in telemetry | **DECLINED — accepted, dropped** (see below) |
| Minor 4 — contract carries the credentialed proxy URL | **ADDRESSED** for a current PATCH 10 install; the unpatched-upgrade case is undocumented (see OPTIONAL 3) |
| Minor 5 — `required: true` on PATCH 10 | **ADDRESSED** — kept required, rationale documented, and that is the right tradeoff |
| Minor 6 — anonymous nested function evades the guard | **ADDRESSED** |

## The previous blocking ask — closed

I re-ran the mutation that exposed the missing merge-branch coverage at `21763dc`: after `src/patch-transforms.ts:582` rewrites `process.env`, change the spread expressions back to `process.env` while leaving the early return rebound.

```
tests/patcher.test.ts   Tests  10 failed | 95 passed (105)

× restores the original network environment on the merge branch
    - HTTPS_PROXY: http://proxy.example.test:8080
    + HTTPS_PROXY: http://127.0.0.1:3457
    - NODE_EXTRA_CA_CERTS: /tmp/external-ca.pem
    + NODE_EXTRA_CA_CERTS: /tmp/local-ca.pem
× 8 × "matches the host contract reader for …"
    (these also expose the surviving contract variable)
× PATCH_TRANSFORMS_VERSION digest pin
    (discounted — it fires on any source edit)
```

Nine behavioral reds remain after discounting the digest. The new case at `tests/patcher.test.ts:1342-1370` passes `{ CHILD_ENV_MARKER: 'merge-branch' }` as extra environment, forcing the `let u={...process.env,…}` path. It asserts the restored proxy and CA, the unrelated merged value, and contract removal. The eight fixed reader comparisons at `tests/patcher.test.ts:1438-1490` also force the merge branch and catch the contract leak. Last round, this defect class left the full suite green except for the digest pin; the missing path is now behaviorally covered.

Full suite at head under a dead ambient proxy: 85 files / 1,394 tests. Typecheck and production build also pass, and the hosted `test` and `commitlint` checks are green.

## Optional items

### OPTIONAL 0 — PATCH 10's proof covers the injected prelude, not the transformed builder body

**Raised as a blocker during review and demoted after checking it against the rest of the repo. Recorded in full because the mechanism is real and worth knowing; no change is requested.**

PATCH 10 does two things: it injects the restore prelude and rewrites the original builder body from `process.env` to `_clodexChildEnv` (`src/patch-transforms.ts:582-614`). The proof added at `src/built-in-patch-proofs.ts:165-168` stops at `catch(_clodexError){}}`, immediately before that rewritten body. It therefore protects the prelude but not the merge spread whose correctness this round was specifically meant to pin.

I drove this through the production local-patch API and the `applyPatch` publication path. A valid local patch, with its required `/*clodex-local:…*/` marker intact, changed only:

```js
let u={..._clodexChildEnv,...tTs,...e,...n}
```

to:

```js
let u={...process.env,...tTs,...e,...n}
```

The result was:

- `applyLocalPatchSet` reported `OK`;
- `builtInPatchProofsChanged` returned `false`;
- re-running `applyClodexPatches` returned byte-identical content with PATCH 10 `SKIP`;
- `applyPatch` returned success and handed the local marker plus the broken merge to `writeContent` (`src/patcher.ts:541-572,759-770`).

Executing that builder on the real 2.1.226 bundle leaves the local bridge proxy, local CA, and credential-bearing contract in the standard child's environment.

**Reachability:** this requires local patches to be explicitly enabled, but that is a supported production path, not a test-only option or compromised infrastructure. Local modules are trusted code; this is not a sandbox/security claim. The defect is that an overlapping or mistaken local transform bypasses the repository's separate guarantee that local patches may not alter built-in sites, and silently removes PATCH 10's isolation while the patch is reported successful.

**Why this is not a blocking ask.** Proof coverage in this repository has always been *proof*-granular rather than *emission*-granular, and that predates this PR. PATCH 5 on `main` is the clearest case: its proof captures only the `{value:…,label:…,description:…}` entry object (`src/built-in-patch-proofs.ts:132-143` at `ad907a5`), while the `.forEach(function(_o){…e.push(_o)})` glue that actually inserts the model into the picker is uncovered. Through the same production `applyPatch` path, a local patch neutering that glue publishes with every site reported OK. PATCH 10 is therefore consistent with the established posture, not a departure from it, and local-patch modules are explicitly trusted, opted-in code that already execute arbitrary top-level statements in-process — the layer is a mistake-catcher, not a security boundary.

What is genuinely new is the *consequence*: PATCH 10 is the first uncovered region that is both `required: true` and isolation-bearing. The wholesale "replace the builder" class of mistake is still caught, because it drops the prelude; only a transform that preserves the 2,104-byte prelude verbatim and then rewrites `_clodexChildEnv` inside the body escapes.

**If you want to close it anyway** (optional): extending the proof pattern through the builder's final `for (… delete …); return … }` matches uniquely on all 13 real bundles (lengths 3375/3378/3567), always covers the merge spread, and detects the mutation, with all 105 patcher tests still green. Worth stating explicitly if you take it: this locks the entire builder body against local patches, removing what is otherwise a legitimate extension point.

### OPTIONAL 1 — the fail-closed literal does not pin the merge spread

`requiredBodyLiterals` includes `'{...process.env'` at `src/patch-transforms.ts:564`, and `targetIsValid` only checks that the body contains it (`src/patch-transforms.ts:576-577`). Every tested bundle from 2.1.208 through 2.1.226 contains that substring twice:

```js
n=yr(process.env.CLAUDE_CODE_REMOTE)?mxu(t?{...process.env,...e}:process.env):{}
let u={...process.env,...tTs,...e,...n}
```

Only the second spread produces the returned child environment. Migrating only that spread to the live accessor (`let u={...te,…}`) leaves the first occurrence intact, so validation still succeeds and PATCH 10 reports `OK`. I applied that mutation to the real 2.1.226 body and executed the result: the child retained the local bridge proxy and CA plus `CLAUDE_CODE_CLODEX_NETWORK_ENV`.

The test at `tests/patcher.test.ts:1284-1293` mutates `CLAUDE_CORE_FIXTURE`, whose builder has only one `{...process.env`, so it proves fail-closed behavior for total drift but not this partial drift.

**Reachability:** not reachable in any tested released bundle. All 13 still use `process.env` for the returned merge, so this is future-drift hardening only. A postcondition on the rewritten body is the least brittle fix:

```ts
const mergeVar = restoredBody.match(/let ([\w$]+)=\{\.\.\._clodexChildEnv/)?.[1];
const returnVar = restoredBody.match(/return ([\w$]+)$/)?.[1];
if (!mergeVar || mergeVar !== returnVar) fail('…');
```

I verified `mergeVar === returnVar === 'u'` on all 13 bundles and that this rejects the `{...te,` mutation.

### OPTIONAL 2 — the host-side contract reader is not behaviorally pinned

The fixed comparisons at `tests/patcher.test.ts:1438-1490` compute their expected values with `networkEnvBaseline(baseEnv)`. They therefore pin the baked reader against the host reader, but do not independently pin the host reader.

Deleting the identical-key-set rejection at `src/network-env.ts:51-54` leaves `tests/patcher.test.ts`, `tests/env.test.ts`, and `tests/wrapper-env.test.ts` green except for the digest pin. `networkEnvBaseline` re-checks membership while applying values (`src/network-env.ts:81-84`), so none of the existing fixed cases distinguishes the parse-level rejection.

The changed behavior is concrete: without `:51-54`,

```js
original: { HTTPS_PROXY: 'a', NO_PROXY: 'b' }
injected: { HTTPS_PROXY: 'x' }
```

is rejected wholesale by the baked reader but partially applied by the host reader. `recordNetworkEnvMutation` (`src/network-env.ts:93-105`) always writes identical key sets, so this shape requires hand-editing or corruption. Its consequence is a nested-wrapper baseline computed from a partially trusted contract, not a default-path bridge leak; hence optional.

Two asymmetric fixed cases with literal expected outputs, rather than expectations computed from `networkEnvBaseline`, would pin the host side.

### OPTIONAL 3 — the credential paragraph omits the unpatched upgrade case

The documentation opens with *"After `clodex patch`, …"* (`docs/background-agents.md:88`). That accurately describes a current PATCH 10 install. However, contract recording is unconditional in `buildHttpProxyChildEnv` and both live-server branches of `computeWrapperEnv` (`src/env.ts:214`, `src/wrapper-env.ts:75,83`); it does not consult the patch manifest.

On an unpatched install — including a user upgrading from main's version-3 patch before accepting this PR's version-5 re-patch — the shared child builder has no PATCH 10 stripping. The contract and any credential-bearing external proxy URL inside it therefore propagate to standard child-command environments. In the endpoint-wrapper branch specifically, base removed all four proxy variables and carried no credential URL, while this branch carries that URL in the contract JSON; I executed that base/head differential.

**Reachability:** real but bounded. The same proxy credential ordinarily begins in the user's shell environment, so this is a regression of incidental protection and a new `CLAUDE_CODE_`-prefixed carrier, not disclosure of an otherwise unreachable secret. A sentence naming the unpatched/stale-from-version-3 case is the surgical fix. Patch-state gating is not a cheap helper-level alternative here: `computeWrapperEnv` intentionally stays dependency-light and its wrapper path does not perform the launch-time patch-state probe.

### OPTIONAL 4 (nit) — documentation wrap and timing wording

`docs/background-agents.md:95-97` wraps *"while authenticated proxy settings are present"* across three lines. It renders correctly.

The same paragraph says the parent retains the contract *"until it builds a standard child environment"*. The emitted builder deletes the key from a copy, so the parent actually retains it for its lifetime. *"The parent retains it for its lifetime; standard child builders strip it from their child copy"* would be exact. The warning is conservative and correct either way.

---

## Checked and cleared

- **Contract-variable rename:** ask dropped. I executed base/new and old-name/renamed reader-writer combinations; every mismatch returns normally and leaves the local bridge in place. The key has never shipped on `main`, so renaming before this PR's first release would not itself require migration, but the first-upgrade binary/wrapper skew already lacks isolation under either spelling. Once a spelling ships, a later rename needs dual-read/dual-write.
- **Keeping PATCH 10 `required: true` is correct.** A required failure prevents `applyClodexPatches` from returning, so `applyPatch` never reaches `writeContent` or the candidate rename. The installed binary and manifest remain untouched; making the site optional would silently publish the leak instead.
- **Anonymous nested functions are now rejected.** The stricter `/\bfunction\s*[\w$]*\(/` guard is false on all 13 real builder bodies, and injecting `let nested=function(){};` into the real 2.1.226 body now throws.
- **Current `te.*` migration is harmless.** The real 2.1.226 builder has three `te.*` reads (`CLAUDE_BG_SOCKET_TOKENS_PATH`, `CLAUDE_BG_RV_AUTH`, `CLAUDE_BG_PTY_AUTH`) and no network-variable reads through `te`. I executed the bundle's actual `te=BYi(...)` getter factory and confirmed it reads changed `process.env` values live; the three current keys are identical between the live environment and the restored copy.
- **`computeWrapperEnv`'s no-server return is intentional** (`src/wrapper-env.ts:62`). A background agent under the per-session MITM sees no advertised standalone server; forwarding the inherited bridge keeps that agent bridged.
- **Anchor integrity remains sound.** Re-running the transform over all 13 real bundles produced exactly one match per bundle, the same child-env builder identities and 1,237–1,421-byte spans, 21–23 rewrites, and no rewritten reference outside the declaring function. The patched bundles parse under `vm.Script`.
- **Transform versioning is correct.** `PATCH_TRANSFORMS_VERSION` is 5, and the digest pin now covers both `patch-transforms.ts` and `network-env.ts` (`tests/patcher.test.ts:449-460`).

## Verification scope

I ran the full head and base suites under the same dead ambient proxy, plus head typecheck/build, real-bundle anchor/mutation harnesses, extracted-builder execution, and an `applyPatch` local-proof publication harness. I did not repack a real installed Claude Code native binary, launch a real Claude Code and inspect an actual Bash/MCP child, or run a `claude -p` end-to-end smoke test. The publication harness used mocked tweakcc extraction/repacking while exercising the real `applyPatch` control flow and source handed to `writeContent`.

PR #79, opened 2026-08-06
fix(proxy): contain upstream socket resets

camjac251:
## Summary

- handle request-socket failures that arrive before response headers
- reuse the existing bounded `502` response and lifecycle diagnostics
- detach the temporary socket listener as soon as response handling takes over

## Scope

This branch was rebuilt directly on current `main`. The route-scoped configuration system, child-environment policy, transport replacement, and persisted proxy URL were removed. Raw passthrough support for the existing `HTTP(S)_PROXY`/`NO_PROXY` mechanism is isolated in #92.

## Test plan

- [x] `corepack pnpm typecheck`
- [x] `corepack pnpm test` (83 files, 1,297 tests)
- [x] `corepack pnpm build`
- [x] focused pre-response socket reset test proving one bounded `502`
- [x] `git diff --check`
- [x] `gitleaks git --staged`

### DIFF
--- a/src/http-proxy/server.ts
+++ b/src/http-proxy/server.ts
@@ -160,6 +160,8 @@ export interface HttpProxyOptions {
   anthropicOrigin?: string;
   /** Test hook for a local self-signed Anthropic origin. */
   anthropicRejectUnauthorized?: boolean;
+  /** Test hook for exercising Anthropic request socket failures. */
+  anthropicRequest?: typeof https.request;
   /** Test hook for observing relay-route isolation without calling an AI provider. */
   adapterHandle?: ProxyHandle;
   /** Test hook for exercising adapter request transport failures. */
@@ -269,6 +271,7 @@ function forwardRawAnthropicRequest(
   rawBody: Buffer,
   origin: URL,
   rejectUnauthorized: boolean,
+  requestImpl: typeof https.request = https.request,
   onErrorResponse?: (statusCode: number, body: string) => void,
   onResponseUsage?: (usage: ResponseUsage) => void,
   lifecycle?: {
@@ -337,7 +340,39 @@ function forwardRawAnthropicRequest(
       resolve();
     };
     const errorType = (err: Error): string => (err as NodeJS.ErrnoException).code ?? err.name;
-    const upstream = https.request({
+    const failUpstream = (err: Error) => {
+      if (clientDisconnected || failed) {
+        done();
+        return;
+      }
+      failed = true;
+      stopProgress();
+      const now = Date.now();
+      writeLifecycle('response_failed', {
+        statusCode: 502,
+        phase: responsePhase(),
+        durationMs: now - startedAt,
+        idleMs: now - lastActivityAt,
+        bytes,
+        chunks,
+        errorType: errorType(err),
+        terminationSource: 'upstream_failure',
+      });
+      onErrorResponse?.(502, `Anthropic upstream unreachable: ${err.message}`);
+      if (!res.headersSent) res.writeHead(502, { 'Content-Type': 'text/plain' });
+      res.end(`Anthropic upstream unreachable: ${err.message}`);
+      done();
+    };
+    let requestSocket: Socket | undefined;
+    let onRequestSocketError: ((err: Error) => void) | undefined;
+    const detachRequestSocketError = () => {
+      if (requestSocket && onRequestSocketError) {
+        requestSocket.removeListener('error', onRequestSocketError);
+      }
+      requestSocket = undefined;
+      onRequestSocketError = undefined;
+    };
+    const upstream = requestImpl({
       protocol: 'https:',
       hostname: origin.hostname,
       port: origin.port || 443,
@@ -347,6 +382,11 @@ function forwardRawAnthropicRequest(
       servername: net.isIP(origin.hostname) ? undefined : origin.hostname,
       rejectUnauthorized,
     }, upstreamRes => {
+      detachRequestSocketError();
+      if (failed) {
+        upstreamRes.destroy();
+        return;
+      }
       headersReceived = true;
       statusCode = upstreamRes.statusCode ?? 502;
       lastActivityAt = Date.now();
@@ -392,6 +432,14 @@ function forwardRawAnthropicRequest(
         done();
       });
     });
+    upstream.once('socket', socket => {
+      requestSocket = socket;
+      onRequestSocketError = err => {
+        failUpstream(err);
+        upstream.destroy(err);
+      };
+      socket.once('error', onRequestSocketError);
+    });
     res.once('finish', () => {
       stopProgress();
       if (failed || clientDisconnected) return;
@@ -423,27 +471,8 @@ function forwardRawAnthropicRequest(
       done();
     });
     upstream.once('error', err => {
-      if (clientDisconnected) {
-        done();
-        return;
-      }
-      failed = true;
-      stopProgress();
-      const now = Date.now();
-      writeLifecycle('response_failed', {
-        statusCode: 502,
-        phase: responsePhase(),
-        durationMs: now - startedAt,
-        idleMs: now - lastActivityAt,
-        bytes,
-        chunks,
-        errorType: errorType(err),
-        terminationSource: 'upstream_failure',
-      });
-      onErrorResponse?.(502, `Anthropic upstream unreachable: ${err.message}`);
-      if (!res.headersSent) res.writeHead(502, { 'Content-Type': 'text/plain' });
-      res.end(`Anthropic upstream unreachable: ${err.message}`);
-      done();
+      detachRequestSocketError();
+      failUpstream(err);
     });
     upstream.end(rawBody);
   });
@@ -877,6 +906,7 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
         rawBody,
         anthropicOrigin,
         options.anthropicRejectUnauthorized ?? true,
+        options.anthropicRequest,
         messagesEndpoint === 'messages' && options.inferenceLogPath
           ? (statusCode, errorContent) => writeInferenceResponseErrorLog(options.inferenceLogPath!, {
               requestId,
@@ -919,6 +949,7 @@ export async function startHttpProxy(options: HttpProxyOptions): Promise<HttpPro
       rawBody,
       anthropicOrigin,
       options.anthropicRejectUnauthorized ?? true,
+      options.anthropicRequest,
     );
   });
 
--- a/tests/http-proxy-server.test.ts
+++ b/tests/http-proxy-server.test.ts
@@ -109,6 +109,24 @@ function adapterRequestWithResponseEvents(
   }) as unknown as typeof http.request;
 }
 
+function anthropicRequestWithSocketError(error: Error): typeof https.request {
+  return ((_options: https.RequestOptions, _onResponse: (response: http.IncomingMessage) => void) => {
+    const request = new EventEmitter() as EventEmitter & {
+      end(body?: Buffer): void;
+      destroy(error?: Error): void;
+    };
+    request.end = () => {
+      queueMicrotask(() => {
+        const socket = new EventEmitter();
+        request.emit('socket', socket);
+        socket.emit('error', error);
+      });
+    };
+    request.destroy = () => {};
+    return request;
+  }) as unknown as typeof https.request;
+}
+
 async function adapterResponseFailureEntries(
   logName: string,
   emitEvents: (response: http.IncomingMessage) => void,
@@ -224,6 +242,39 @@ describe('selective HTTP proxy', () => {
       for (const client of clients) client.destroy();
       await proxy.close();
       await new Promise<void>(resolve => upstream.close(() => resolve()));
+    }
+  });
+
+  it('returns 502 when the upstream request socket resets before a response', async () => {
+    const certificates = ensureHttpProxyCertificates();
+    const origin = https.createServer({
+      key: certificates.serverKey,
+      cert: certificates.serverCert,
+    }, (_req, res) => {
+      res.writeHead(200, { 'Content-Type': 'application/json', 'Connection': 'close' });
+      res.end('{}');
+    });
+    const originPort = await listen(origin);
+    const socketError = Object.assign(new Error('socket hang up'), { code: 'ECONNRESET' });
+    const proxy = await startHttpProxy({
+      routes: [],
+      anthropicOrigin: `https://127.0.0.1:${originPort}`,
+      anthropicRejectUnauthorized: false,
+      anthropicRequest: anthropicRequestWithSocketError(socketError),
+    });
+
+    try {
+      const response = await requestMitm(
+        proxy.port,
+        certificates.caCert,
+        '/api/oauth/usage',
+        '{}',
+      );
+      expect(response).toContain('502 Bad Gateway');
+      expect(response).toContain('Anthropic upstream unreachable: socket hang up');
+    } finally {
+      await proxy.close();
+      await new Promise<void>(resolve => origin.close(() => resolve()));
     }
   });
 


bman654 (commented):
Reviewing this as a draft, so this is design feedback rather than a line-by-line pass — the base
has to move anyway (see the note at the end). Two lenses looked at it independently and landed on
the same recommendation: **land it split, and drop most of it.**

First, the good news, and it's the part I was most worried about: **`src/oauth/responses-websocket.ts`
is genuinely untouched behaviourally.** Both lenses audited it. The diff is two lines — an additive
`proxyUrl?` option and passing it to `outboundWsProxyAgent`. `responsesWebSocketPartitionKey` still
takes `Pick<…, 'providerId' | 'accountId'>`, so partition key material is unchanged; chain-head
capture and validation, longest-prefix continuation, safe fallback to full context, and the
`previous_response_not_found` single retry are all unmodified. Thank you for respecting the
do-not-restructure marker on that file — that was the right instinct.

Likewise the socket-reset change does **not** invert retryability. Pre-response upstream errors
already produced `502 Anthropic upstream unreachable`; the refactor moves that into `failUpstream`
and adds a raw-socket listener catching errors Node would otherwise leave unhandled.
`detachRequestSocketError` runs on both paths so no listener accumulates on pooled keep-alive
sockets, and mid-stream failures still `res.destroy()` rather than becoming a status code. No prose
scanning introduced. That part is clean.

## The core design concern: this duplicates `NO_PROXY` with strictly worse coverage

`src/outbound-proxy.ts:31-67` already implements host-selective proxying (`noProxyBypasses`,
`outboundProxyUrlForTarget`), applied process-wide to every `fetch`. So
`HTTPS_PROXY=… NO_PROXY=api.anthropic.com` already expresses "OpenAI through the proxy, Anthropic
direct" — uniformly, across every clodex path. The README concedes the overlap at line 280.

The new surface — `upstream-proxy.ts` / `-command.ts` / `-fetch.ts`, the `upstreamProxy`
preferences key, four CLI subcommands — buys selectivity that already exists, and covers *fewer*
paths than the env mechanism it parallels:

- **`route openai on` misses the whole auth/metadata surface.** `spec.proxyUrl` reaches only
  `createLanguageModel` and the WS transport. The OpenAI device flow and `/oauth/token` exchange
  (`src/oauth/openai.ts:42,70,84`), token refresh (`src/oauth/refresh-http.ts:28`), model list
  (`src/registry/refresh-models.ts:150`), models.dev, and pricing all use bare global `fetch`.
  In the motivating environment — egress only via the proxy — a Codex-OAuth user turns
  `route openai on`, inference works, then the access token expires mid-session, `refresh-http.ts`
  reaches `auth.openai.com` directly, hangs, and the provider dies with an error that looks nothing
  like a proxy misconfiguration. `HTTPS_PROXY` covers all of it.
- **Coverage is silently bridge-mode dependent.** `route openai` is wired only through
  `buildConfiguredHttpProxyOptions` and `clodex server`'s `openAiProxyUrl`; `clodex claude
  --endpoint` builds catalog routes elsewhere and never sets `proxyUrl`. `route anthropic` only
  affects the MITM raw passthrough, so it's a no-op in endpoint mode entirely. `tools` is applied
  in `buildHttpProxyChildEnv` and `computeWrapperEnv` but not in `buildChildEnv`. An endpoint-mode
  user can run all four commands, have `status` report `Anthropic: on / OpenAI: on / Tools:
  upstream`, and have none of it take effect. `README.md:191-195` states the toggles as
  unconditional.

The one genuine gap is real and worth fixing: `forwardRawAnthropicRequest`
(`src/http-proxy/server.ts:490`) uses `https.request` directly and is the single outbound path that
ignores `HTTP_PROXY`/`NO_PROXY`. But that's ~20 lines — reuse `outboundProxyUrlForTarget` and give
the existing `https.request` a proxy `agent`, which you already plumbed as `anthropicAgent` and
then didn't use for this — not a second configuration system.

## What I'd suggest landing, in order

1. **Socket-reset → bounded 502, alone, against `main`.** ~40 lines, no config, no patch bump,
   closes a real unhandled-`'error'` window. No dependency on #78.
2. **Make the raw Anthropic passthrough honour the existing outbound proxy env.** That's the whole
   genuine gap, expressed in the mechanism that already exists.
3. **Child-tool network modes belong to #78's PATCH 10**, not here — they're the only part needing
   `PATCH_TRANSFORMS_VERSION`, and they're coupled to the anchor that has to change anyway.

Everything else I'd drop unless there's a concrete scenario `NO_PROXY` can't express; none is
stated in the PR.

## Specific problems, if any of this does land as-is

**The new undici passthrough is not byte-equivalent to the `https.request` path it shadows.**
`src/http-proxy/server.ts:483-498` hands undici the client's raw header array. undici's
`processHeader` throws `InvalidArgumentError` for `transfer-encoding`, `keep-alive`, `upgrade` and
duplicate `host`, and `NotSupportedError` for `expect` — all of which `https.request` forwards
verbatim today. The rejection surfaces as a synthetic 502. Verified live against `startHttpProxy`
with a real CONNECT tunnel: `Transfer-Encoding: chunked` → `502 … invalid transfer-encoding
header`, origin never saw it; `Expect: 100-continue` → `502 … expect header not supported`.
Streaming SSE control case passed. That's a passthrough that silently rejects legal requests.

**A new `ProxyAgent` is constructed per request on the hot path.**
`src/upstream-proxy-fetch.ts:11` builds `new ProxyAgent(proxyUrl)` on every
`createUpstreamProxyFetch`, and `src/provider-factory.ts:172` calls it on every
`createLanguageModel` — which `src/proxy.ts:589` invokes *inside* the per-request handler, with no
model cache on that path (unlike `server/router.ts:714`, which caches and correctly folds
`proxyUrl` into the key). Every Claude Code turn pays a fresh TCP + CONNECT + TLS handshake and
leaks the previous agent's sockets until `keepAliveTimeout`. Worse, for OAuth Responses models the
branch at `provider-factory.ts:198-209` never uses `proxyFetch` at all — so the agent is built per
request and thrown away unused. Hoist it into a module-level `Map<proxyUrl, ProxyAgent>` and build
only when the consuming branch is selected.

**`clodex-claude` can be bricked for every agent spawn by a config value.**
`src/claude-wrapper.ts:136` calls `getUpstreamProxyConfig()` unguarded, and `main()` runs as
`void main()` with no `.catch`. `resolveUpstreamProxyConfig` throws on a non-`http(s)` scheme, and
`new URL()` throws on an unparseable or non-string value. So a `config.json` that acquires
`"upstreamProxy": {"url": "socks5://…"}` — hand edit, partial write, or a future clodex that adds
SOCKS — makes every `clodex-claude` invocation die on an unhandled rejection, taking out every
background agent and pty host with no output explaining why. That directly violates the wrapper
invariant your own comment at `src/claude-wrapper.ts:127` states: a down server must never break
launching claude. Only `clear` recovers, and nothing tells the user.

**Unknown or stale child-mode fails open and leaks the proxy credential.**
`applyChildNetworkPolicy` (`src/wrapper-env.ts:43-55`) is applied unconditionally without
consulting patch state, so `CLODEX_CHILD_NETWORK_MODE` and the full credential-bearing
`CLODEX_CHILD_UPSTREAM_PROXY` are always placed in the claude child env. On an **unpatched** binary
nothing strips them, so `http://user:pass@proxy:8080` sits in the environment of every command
Claude Code spawns, including model-authored `bash` running `env`. On a **stale** patch (v2/v3) the
old transform deletes only `CLODEX_ORIGINAL_NETWORK_ENV`, so the two new vars survive. And an
**unrecognised** mode string matches none of the `direct`/`upstream`/`original` branches
(`src/patch-transforms.ts:571` defaults only when the var is *absent*), so the snapshot isn't
restored and the child inherits the bridge env — failing open to exactly the behaviour #78 exists
to fix.

**Proxy credentials are persisted in plaintext.** `src/upstream-proxy-command.ts:116` writes the
raw URL, password included, to `~/.clodex/config.json`. `formatUpstreamProxyStatus:66` masks it on
display, so the intent was clearly to treat it as a secret — but clodex has a keychain for exactly
this, and the cleartext value also gets exported verbatim into child environments in `upstream`
mode.

**The self-loop guard degrades from structural to trust-the-user.** It held because clodex derived
its proxy only from env vars it never set on itself. No new `process.env[x] =` writes were
introduced and the mutation guard test still passes — but the dispatcher source is now an arbitrary
user URL with no check that it isn't clodex's own listener. `upstream-proxy set
http://127.0.0.1:17645; route anthropic on` with a `clodex server --proxy` up on the default port
makes the passthrough CONNECT to clodex, which re-intercepts and CONNECTs again. A loopback/own-port
check at `set` time would restore it as an assertion. The header comment in `outbound-proxy.ts` also
still claims env vars are the only source.

## The new test fails for exactly the users this feature targets

`tests/env.test.ts:272-292` sets `HTTPS_PROXY`/`NO_PROXY` and asserts the snapshot `toEqual`s
exactly those two keys. But `preserveOriginalNetworkEnv` snapshots the *ambient*
`CHILD_NETWORK_ENV_VARS`, so any pre-existing `HTTP_PROXY`, `http_proxy`, `no_proxy` or
`NODE_EXTRA_CA_CERTS` also lands in the object and the assertion fails.

Measured on my machine: with a scrubbed environment, 29/29 pass. In my normal shell, 1 fails — my
shell has `HTTPS_PROXY=http://127.0.0.1:17645` and `NODE_EXTRA_CA_CERTS=…/clodex-ca.pem`, because
that session is itself running bridged through clodex. So `pnpm test` is red for the corporate-proxy
developer this feature is aimed at, and for anyone running clodex under clodex. It passes in CI only
because runners have a clean env.

The `finally` block also unconditionally `delete`s the developer's real `HTTPS_PROXY`/`NO_PROXY` for
the rest of the vitest process, contaminating later tests. Fix by stubbing `process.env` for the
case, or asserting `toMatchObject` on just the two keys under test.

## Smaller things

- `src/upstream-proxy.ts:45` has a dead `import { ProxyAgent } from 'undici'` placed after all code,
  unreferenced. It's tree-shaken out of the wrapper chunk, but it drags a dependency edge into a
  module `claude-wrapper.ts` now imports transitively — and CLAUDE.md asks that the wrapper's
  imports stay minimal.
- `{ anthropic: false, openai: false, tools: 'original' }` is repeated verbatim in three signatures
  (`src/env.ts:104`, `src/http-proxy/index.ts:128`, `src/wrapper-env.ts:98`); one exported
  `DEFAULT_UPSTREAM_PROXY_CONFIG` would do.
- `forwardRawAnthropicRequest` now takes ten positional parameters with three new ones inserted
  *before* the existing optional callbacks — a rebase hazard for anything else touching those call
  sites.
- No coverage of an **SSE streaming** response through the dispatcher, which is the production shape
  for `/v1/messages`. The non-streaming CONNECT path is well covered
  (`tests/http-proxy-server.test.ts:270-365`).

## Base note

This branch contains #78's two commits, and #78 has a blocker: its PATCH 10 anchor binds ~407 KB too
early on the real bundle and publishes a Claude Code binary with 125 out-of-scope identifier
references. #79 carries the identical anchor and the same `body.replace(/process\.env/g, …)`, plus
the extra `direct`/`upstream` branches on top. Worth fixing at the base before this moves.

camjac251:
Review update:

- Rebuilt directly on current `main` as one socket-reset change.
- Removed the route-scoped configuration system, persisted proxy URL, child-mode policy, dispatcher replacement, and all dependent patch changes.
- The remaining code handles the pre-response request-socket error window, reuses the existing bounded `502`, and detaches its listener when response handling begins.
- The raw passthrough gap in the existing environment mechanism is isolated in #92 without changing this branch.

Verification: typecheck; build; focused reset coverage; 83 files / 1,297 tests; diff check; staged leak scan.

bman654 (changes requested):
Thanks for splitting this — you did exactly what we asked. This is now a single commit on `main` (`827215a`), 2 files, +104/−22, no `#78` dependency, no patcher changes, out of draft, CI green. That made it reviewable on its own merits, which is what we wanted. Reviewed fresh, from scratch.

The scoping is right. The problem is with the premise: I could not find a Node version, or a failure shape, in which the window you are guarding actually exists. Details and every command below.

---

## Standalone check — clean ✅

```
$ git log --oneline 827215a..HEAD
62102f6 fix(proxy): contain upstream socket resets

$ git diff 827215a..HEAD --stat
 src/http-proxy/server.ts        | 75 +++++++++++++++++++++++++++++------------
 tests/http-proxy-server.test.ts | 51 ++++++++++++++++++++++++++++
 2 files changed, 104 insertions(+), 22 deletions(-)

$ gh pr view 79 --json isDraft,baseRefName,changedFiles,statusCheckRollup
 isDraft: false   base: main   changedFiles: 2
 test: SUCCESS    commitlint: SUCCESS
```

No `src/patch-transforms.ts`, no `PATCH_TRANSFORMS_VERSION` implication, no `#78` content, no config surface, no new dependency. Both call sites of `forwardRawAnthropicRequest` were updated for the new positional parameter (verified by grep — there are exactly two, at `:903` and `:946`).

Local, in a clean env (`env -u HTTPS_PROXY -u HTTP_PROXY -u NODE_EXTRA_CA_CERTS`, `CLODEX_HOME=$(mktemp -d)`):

```
pnpm typecheck  → exit 0
pnpm test       → Test Files 83 passed (83) · Tests 1297 passed (1297)
pnpm build      → ESM dist/cli.js 2.79 MB · Build success in 935ms
```

---

## MAJOR — the guarded window does not appear to exist on any supported Node

The commit message states the premise as:

> A request socket can emit an error before the request object reports the failure.

I tried to reproduce that and got the opposite result in every shape I could construct: **the `ClientRequest` reports the failure first, synchronously, and the socket error either arrives afterwards or never fires at all.**

### Why, from Node's own source

`_http_client.js::tickOnSocket` — offsets within the function body, on the engines floor (v22.14.0):

```
$ node -e "const c=process.binding('natives')['_http_client']; const i=c.indexOf('function tickOnSocket'); const b=c.slice(i,c.indexOf('\nfunction ',i+10));
  console.log('_httpMessage=req  @', b.indexOf('socket._httpMessage = req'));
  console.log('attach socketErrorListener @', b.indexOf(\"socket.on('error', socketErrorListener)\"));
  console.log(\"req.emit('socket') @\", b.indexOf(\"req.emit('socket'\"));"
_httpMessage=req  @ 517
attach socketErrorListener @ 819
req.emit('socket') @ 1119
```

Node assigns `socket._httpMessage = req` and attaches its own `socketErrorListener` **before** it emits `'socket'` — i.e. strictly before your listener can be attached. And `socketErrorListener` forwards unconditionally:

```js
function socketErrorListener(err) {
  const socket = this;
  const req = socket._httpMessage;
  if (req) { req.socket._hadError = true; emitErrorEvent(req, err); }   // → req.emit('error', err)
  ...
}
```

`emitErrorEvent` is a plain `request.emit('error', error)` — no suppression. So the only way a socket error escapes the request is `socket._httpMessage == null`. I grepped every assignment of that field:

- `_http_client.js` — nulled in exactly one place: the **upgrade/CONNECT** branch, and only when `req.listenerCount('connect'|'upgrade') > 0`. This request registers neither.
- `_http_agent.js` — nulled when the socket is returned to the **free pool**, which happens after the response completes, by which point your `detachRequestSocketError()` has already run.

Between `'socket'` and the response callback — precisely your listener's lifetime — `_httpMessage` is always your request.

### Measured, on real sockets

Eight failure shapes against real `https.request`, logging the order of `req 'error'` vs `socket 'error'` (`/tmp/pr79probe/probe.mjs`):

| shape | v22.14.0 | v24.14.1 | v25.8.1 |
|---|---|---|---|
| ECONNREFUSED | REQ then SOCKET | REQ then SOCKET | REQ then SOCKET |
| TCP destroyed pre-TLS | REQ then SOCKET | REQ then SOCKET | REQ then SOCKET |
| TLS up, destroyed pre-response | **REQ only** | REQ only | REQ only |
| TLS up, reset after `req.end()` | **REQ only** | REQ only | REQ only |
| mid-stream reset after headers | neither (response `aborted`) | same | same |
| TLS cert rejected | REQ then SOCKET | REQ then SOCKET | REQ then SOCKET |
| DNS ENOTFOUND | REQ then SOCKET | REQ then SOCKET | REQ then SOCKET |
| garbage response bytes | **REQ only** | REQ only | REQ only |

### Measured, through the real proxy

I instrumented `failUpstream`'s two entry points with a stderr tag and drove the **real** `startHttpProxy` (no `anthropicRequest` hook, so real `https.request`) against five hostile origins:

```
[PR79-PATH] REQ-HANDLER fired code=ECONNRESET
[PR79-PATH] SOCKET-HANDLER fired code=ECONNRESET
[PR79-CASE] A destroy-pre-TLS :: 502 "Anthropic upstream unreachable: read ECONNRESET"

[PR79-PATH] REQ-HANDLER fired code=ECONNRESET
[PR79-CASE] B destroy-pre-response :: 502 "Anthropic upstream unreachable: socket hang up"

[PR79-PATH] REQ-HANDLER fired code=ECONNRESET
[PR79-CASE] C FIN-no-response :: 502 "Anthropic upstream unreachable: socket hang up"

[PR79-PATH] REQ-HANDLER fired code=HPE_INVALID_CONSTANT
[PR79-CASE] D garbage :: 502 "Anthropic upstream unreachable: Parse Error: Expected HTTP/, RTSP/ or ICE/"

[PR79-PATH] REQ-HANDLER fired code=ECONNREFUSED
[PR79-PATH] SOCKET-HANDLER fired code=ECONNREFUSED
[PR79-CASE] E econnrefused :: 502 "Anthropic upstream unreachable: connect ECONNREFUSED 127.0.0.1:55644"
```

`REQ-HANDLER` — the pre-existing `upstream.once('error')` — wins every race. In A and E the socket handler also runs, second, and does nothing because `failUpstream` short-circuits on `failed`. In B, C and D it never runs at all.

**So on this path the new socket listener produces no 502 that `main` did not already produce, and no 502 arrives sooner.** Every one of those five responses is byte-identical to what `main` emits.

**The ask:** can you share the reproduction that motivated this? Node version, the error you saw, and ideally the `response_failed` / hang you observed. I can't prove a negative for every environment, and if there's a shape I haven't constructed I'd rather see it than guess. If it turns out there isn't one, I'd suggest closing this rather than carrying a permanently-inert branch through the passthrough hot path.

---

## MAJOR (test quality) — the new test stages a shape Node cannot produce, and 5 of 8 mutants survive

The fake request in `anthropicRequestWithSocketError` emits `'socket'` from inside `end()` and then errors the socket while **never** emitting `'error'` on the request object. That is the exact inverse of what real Node does in all eight shapes above — the request always errors, and it errors first.

The mutation results (run the way CI runs, full file, not `-t` isolation — those disagree; see below for why that matters):

```
$ pnpm vitest run tests/http-proxy-server.test.ts     # baseline: 20 passed
```

| mutation | result |
|---|---|
| M1 delete the whole `upstream.once('socket', …)` block | **1 failed / 19 passed** ✅ |
| M7 socket handler body → no-op | **1 failed / 19 passed** ✅ |
| M8 `requestImpl(...)` → `https.request(...)` | **1 failed / 19 passed** ✅ |
| M2 drop `\|\| failed` from `failUpstream`'s guard | 20 passed — **survives** |
| M3 drop `detachRequestSocketError()` in the response callback | 20 passed — **survives** |
| M4 drop `if (failed) { upstreamRes.destroy(); return; }` | 20 passed — **survives** |
| M5 drop `detachRequestSocketError()` in the upstream error handler | 20 passed — **survives** |
| M6 drop `upstream.destroy(err)` from the socket handler | 20 passed — **survives** |

Deleting the feature outright does fail something, so this isn't pure theatre. But every *defensive sub-condition* the PR adds is unverified — including the detach, which the commit message headlines ("detach the socket listener once response handling takes over"). One test covering one fictional shape is the whole of the new coverage.

### The detach doesn't do what the commit message says, in the only ordering that occurs

Look again at case A and case E above: `REQ-HANDLER` runs, calls `detachRequestSocketError()`, and `SOCKET-HANDLER` **still fires**. That's not a bug in your code — `EventEmitter.emit` clones the listener array before iterating, so removing a listener from inside the same emit can't cancel it. What actually prevents the double response is the `|| failed` guard, not the detach. The detach only bites across ticks (the response-callback case). Worth correcting in the commit message if this lands in any form.

---

## Verified clean — please don't re-litigate these

- **Echo invariant intact.** The change is a function-reference swap only. `rawBody`, `req.url`, `req.method` and `requestHeadersWithoutProxyHeaders(req)` are untouched; `copyResponse` still does `res.writeHead(statusCode, upstream.statusMessage, upstream.rawHeaders)` verbatim. Nothing can rewrite a response `model` field, so auto-compaction / patched-alias context-window lookup is unaffected. `options.anthropicRequest` being `undefined` in production correctly falls through to the `= https.request` default parameter.
- **Retryability is not inverted or masked.** Since the socket path never wins the race, the failure-shape → status mapping is identical to `main`, byte for byte (see the five 502 bodies above). No fast retryable failure becomes a slow terminal one, and nothing terminal is newly dressed up as retryable.
- **Streaming is untouched.** The socket listener is detached at the top of the response callback, so mid-stream resets are still handled entirely by `copyResponse`'s `upstreamRes.once('error')` → `res.destroy()`. I verified the client-visible bytes for both a mid-stream RST and a mid-stream FIN:
  ```
  [PR79-CASE] F mid-stream-reset :: raw="HTTP/1.1 200 OK\r\nContent-Type: text/event-stream\r\n
  Transfer-Encoding: chunked\r\n...\r\n\r\n54\r\nevent: message_start\ndata: {...}\n\n\r\n"
  ```
  The body ends after the last chunk with **no terminating `0\r\n\r\n`** — compare the 502 responses above, which all end `...\r\n0\r\n\r\n`. A conformant client (undici / Node `fetch`, which is what Claude Code uses) surfaces that as a premature close, not a successful completion. So no silent truncation, and no content duplication — `res.end(<502 text>)` is unreachable after headers are sent, because `failed` is already set by then.
- **No crash from double-error.** Cases A and E exercise the ordering where `upstream.destroy(err)` is called on a request whose `once('error')` listener has already been consumed. No unhandled `'error'` — `ClientRequest.destroy` early-returns on an already-destroyed request. Verified live.
- **Resources.** No per-request agent construction (this path passes no `agent` at all, so it's `https.globalAgent`), no new buffering, no unbounded growth. The listener is removed on the normal path, so nothing rides into the keep-alive free pool; on the abnormal paths the socket is destroyed anyway.

---

## Minors / nits

- **`forwardRawAnthropicRequest` is now nine positional parameters**, with the new one inserted in the middle. Both call sites are correct, but a mis-ordered argument here would be a silent behavioural change with no type error in some permutations. An options object would be worth it if this function grows again.
- **`HttpProxyOptions` now carries four test-only hooks** (`anthropicOrigin`, `anthropicRejectUnauthorized`, `anthropicRequest`, `adapterHandle`/`adapterRequest`). Consistent with what's already there, so no objection — just noting the exported surface is drifting toward test scaffolding.

## The gap I think is actually there

If the goal is "contain upstream failures on the passthrough route", the shape that genuinely isn't contained today is not a reset — resets already produce a bounded 502 in ~10ms, as measured. It's a **stall**: `forwardRawAnthropicRequest` sets no socket timeout, no headers timeout, and no idle deadline. The only timer in the function is the lifecycle progress interval, which logs and nothing else, and it's `unref()`'d. A silently wedged upstream connection therefore holds the request open indefinitely and the promise never resolves. That's an unbounded failure with a real client symptom, it's about the same size as this change, and it needs no config. If you're up for it, I'd much rather review that.

---

**Verdict: changes requested** — on the premise, not the craft. The split was done exactly as asked and the code itself is careful and self-consistent. I just can't find the window it guards. Show me the reproduction and I'll re-run all of the above against it.

camjac251:
I could not produce a counterexample to the request-first ordering described in the review. The new listener therefore has no demonstrated runtime path, and its synthetic test encodes an event ordering the supported Node implementations do not emit.

I’m closing this instead of carrying inert logic in the passthrough path. The separate raw-passthrough proxy gap remains in #92, now rebased and updated independently. A bounded upstream-stall policy would be a separate behavior and configuration change, so it is not being folded into this PR.

PR #80, opened 2026-08-06
fix(oauth): snapshot function_call arguments in the sanitized downstream shape

### LINKED ISSUE #84: OAuth chain lost on every tool call carrying a filler null: sanitizeToolInput strips outbound, head snapshots raw upstream arguments
## Summary

`sanitizeToolInput` strips keys from tool input on the way **out** to Claude Code, but the OAuth chain head snapshots the **raw upstream `arguments` string**. Claude Code then echoes back exactly what it was given — which is strictly smaller than what the head holds — so the compare fails, the chain is abandoned, and the turn is re-sent with full context and `cached_tokens=0`.

This fires on any tool call where the model emits a filler `null` or a non-required empty array in its arguments.

## Mechanism

- `sanitizeToolInput` (`src/sdk-adapter.ts:416-427`) drops `null`-valued keys and non-required empty arrays.
- It is applied on the outbound leg only: `src/sdk-adapter.ts:788` and `:806` (SSE `input_json_delta`) and `:1027` (non-streaming tool-call block).
- It exists for a good reason — Claude Code forwards some tool inputs verbatim into server-side API calls, e.g. `WebSearch` domain lists, where an empty list is a 400.
- The head's expected assistant item comes from the upstream `response.output_item.done` item via `expectedAssistantItems` → `withoutEphemeralFields(done)`, which keeps `arguments` verbatim.

So the head holds the pre-sanitization arguments and the client can only ever echo the post-sanitization ones.

## Reproduction

Against `main` (`54621d1`), driving the real WebSocket transport:

```
head arguments    : {"path":"file.ts","offset":null}
Claude Code echo  : {"path":"file.ts"}
  -> previous_response_id = undefined, new socket opened      (chain lost, full-context resend)

control (echo keeps the null): {"offset":null,"path":"file.ts"}
  -> previous_response_id = resp_head, input items = 1        (continues)
```

Same kind on both sides (`function_call`), which is why it shows up in the logs as
`history mismatch starting an additional chain ... expected=function_call actual=function_call`.

## Evidence that it is live

From local WebSocket diagnostics since the 2.1.8/2.1.9 continuation fixes — 19,083 head decisions, 5,738 distinct connections:

- 858 `history_mismatch_new_head` (4.5% of decisions)
- of those, **37** are `expected=function_call actual=function_call` (0.19% of decisions, 4.3% of mismatches)
- **37 of 37** sit at `firstMismatch == expectedPrefixItems - 1` — always the final item of the expected prefix, i.e. the head's last assistant `function_call` against the client's echo of it

The class did not exist before those fixes (0 occurrences pre-2026-08-01), because comparison previously stopped earlier at a reasoning item; once reasoning could match, comparison ran deeper and this surfaced.

## Candidate fixes

1. **Snapshot what we actually emitted** (preferred). Store the sanitized arguments in the head, since that is what the client will echo. Keeps a single source of truth and covers both the `null` and empty-array cases.
2. **Normalize at compare time.** Strip nullish values inside the parsed `arguments` before canonicalizing in `normalizeToolCallJson`. Verified to fix the `null` case, but it cannot cover the non-required-empty-array case without the tool's `required` set, which that module doesn't have — so it would only be a partial fix.

Option 1 also avoids duplicating the sanitization rules in two places.

## Credit

Surfaced by @zachspartofaday in #80 while investigating this exact mismatch class; the `expected_hash`/`actual_hash` addition in that PR is what made it diagnosable from the logs.

zachspartofaday:
Retargeted per review — the original compare-time ephemeral-field strip is withdrawn (the review's reachability analysis is correct; details in the thread), and the PR now fixes the live bug it turned out to be circling: **#84**.

## The fix

The head snapshot stored the raw upstream `arguments` string, but the client never sees that string: `sanitizeToolInput` strips `null`-valued keys and non-required empty arrays on the way out, and the client echoes the sanitized object back. A head whose call carried filler `null` or a non-required empty array could never match its own echo, and the chain was lost at that item on every subsequent request.

Per the review's suggestion, the fix is on the snapshot side: `expectedAssistantItems` now snapshots `function_call` arguments in the same downstream shape. The missing piece for the empty-array half — the tool's `required` set — turns out to be available inside the module: the request's own `tools` array carries each function tool's JSON schema as `parameters`, passed through unmodified by the Responses provider (`prepareFunctionTool` in `@ai-sdk/openai@4.0.11` emits `parameters: tool.inputSchema` and sets `strict` only when explicitly configured, which this codebase never does). So the snapshot consults the exact `required` sets the sanitizer consulted, with no cross-module plumbing.

Compare-only: the payload sent upstream is untouched, and the do-not-restructure invariants (head capture, longest-exact-prefix continuation, full-context fallback, single `previous_response_not_found` retry) are unchanged.

## Diagnostics (second commit, per-review tidy-ups applied)

- `expected_hash`/`actual_hash` on the mismatch line, unchanged in behavior; the "(0600, local-only)" comment now states that `--trace` re-prints matching lines to stdout on exit.
- Dump lines are capped at 2,000 chars with a `[truncated]` marker, matching the other content-bearing writers.
- A side past its history's end renders `(absent)` instead of `undefined`.
- `CLODEX_MISMATCH_DUMP` is documented in CLAUDE.md next to the other websocket knobs, privacy tradeoff called out.

## Tests

- Sanitized-echo continuation: null filler dropped (fails on `main`), non-required empty array dropped (fails on `main`), required empty array retained.
- Negative coverage on the equality predicate, per the review's safety note: a genuinely different argument value and a different `call_id` each still start a new chain. (Verified these guard the over-strip direction: the previously-green `if (true)` + `delete call_id` mutation now fails.)
- Full suite: 1,280 passed, typecheck clean.

Branch name is historical — renaming it would close this PR and lose the review thread.

Fixes #84


zachspartofaday:
One correction to the closing observation after reading CLAUDE.md more carefully: the fan-out/partition-sharing behavior is documented, with `CLODEX_WS_MAX_CONNECTIONS`/`CLODEX_WS_MAX_NURSERY_CONNECTIONS` as the tuning knobs. In the traced workload the caps were not the binding constraint (evictions were 17× `idle_ttl`, zero `*_lru_cap`) — the isolated-socket volume came from in-flight collisions on the shared partition, which the caps don't address. Scoping stands: out of scope here, happy to file separately if useful.

zachspartofaday:
Sustained results beyond the initial sweep, from a subsequent long orchestration session on the patched build (~2,500 requests, concurrent subagents, ChatGPT OAuth):

- `function_call`-class history mismatches: **0** (previously the dominant recurring class).
- Continuation decisions: **1,334 continued** vs 142 mismatches vs 562 isolated — continuation is now the majority mode.
- The 142 remaining mismatches are all previously-known residual classes (cross-conversation head additions at `first_mismatch=0`, and the trailing reasoning-omission variant that the `omitted_reasoning` path doesn't cover) — no new signatures introduced by the fix.
- Zero-`cached_tokens` requests: ~10% (79/711 and 141/1327 on the two affected models) vs ~55% pre-fix; cached:uncached input ratio ≈ 4:1.

The `expected_hash`/`actual_hash` additions to the mismatch line are what made the residual classes identifiable from the log alone.

bman654 (commented):
Thanks for this — and especially for the follow-up numbers and for adding the hashes to the mismatch line. That diagnostic is the reason we were able to take this apart properly, and it changed what we understand about our own cache behaviour. Some of what follows disagrees with the diagnosis, but the investigation you did here landed on something real, and it turns out to be our bug rather than the one described.

Verdict up front: **not blocking, but please don't merge the `normalizeToolCallJson` change as-is.** Details below, with the evidence, and a concrete offer at the end.

## The mechanism is real, but it is not reachable in clodex

Your reasoning about the asymmetry is sound: `withoutEphemeralFields` strips `id`/`status`/`phase` when a head's expected assistant items are snapshotted, and the compare side did not. We reproduced your new test failing on `main` exactly as you describe.

The problem is that clodex can never be the client that trips it. `createResponsesWebSocketFetch` is wired in exactly one place — `src/provider-factory.ts:190` — as the `fetch` of the `@ai-sdk/openai` Responses provider. So every payload this transport can ever observe is rendered by that SDK's `convertToOpenAIResponsesInput`, and there is no Responses-format ingress anywhere else (`src/server/router.ts` exposes only `/anthropic/v1/models`, `/openai/v1/models`, `/anthropic/v1/messages`, `/openai/v1/chat/completions`).

On the pinned `@ai-sdk/openai@4.0.11` that converter emits:

- `function_call` → `{type, call_id, name, arguments}` (`dist/index.js:3892-3898`) — no `id`, no `status`, unconditionally
- `function_call_output` → `{type, call_id, output}` (`:4344-4348`)
- `reasoning` → `{type, encrypted_content, summary}` (`:4013-4027`)

The branches that *do* attach `id` require either a provider-defined tool (`shell`/`apply_patch`/custom) or `providerOptions.openai.itemId`. clodex sets neither — `src/sdk-adapter.ts:307` sets only `{ openai: { reasoningEncryptedContent } }`, and there is no `itemId` anywhere in `src/`. `phase` is only ever emitted on role-shaped assistant messages, which your five-way type gate deliberately excludes, so `delete out.phase` is dead even for a hypothetical client.

To avoid resting that on reading alone, we ran a differential over the real pipeline: 60 varied conversations (null-valued tool inputs, non-required empty arrays, nested nulls, `::ts::`-suffixed tool ids, thinking blocks, `WebSearch`/`TodoWrite`) pushed through the real `translateRequest` into the real `provider.responses(...)` with a capturing fetch, then both normalizers — `main`'s and this PR's — run over every emitted item:

```
ITEMS=342 TYPED=270 DIFFS=0
KEYS SEEN=["function_call.arguments","function_call.call_id","function_call.name","function_call.type",
           "function_call_output.call_id","function_call_output.output","function_call_output.type",
           "reasoning.encrypted_content","reasoning.summary","reasoning.type"]
```

Byte-identical on every item, and the key census contains no `id`, no `status`, no `phase`, and no null-valued key. Deleting the new block entirely leaves a real two-turn tool loop continuing identically. On the shipped configuration this change cannot alter the outcome of any comparison.

## Where the reported improvement most likely came from

We have a reasonably large local corpus to check this against: **33,492 head decisions over 17,806 distinct WebSocket connections**, split across a version boundary.

| | before 2026-08-01 | after 2026-08-01 |
|---|---|---|
| head decisions | 14,409 | 19,083 |
| distinct WS connections | 12,068 | 5,738 |
| `history_mismatch_new_head` | 7,412 (51.4%) | 858 (4.5%) |
| `continuation` | 16.3% | 69.9% |
| `expected=function_call actual=function_call` | **0** (0.00%) | **37** (0.19%) |

Three continuation fixes landed in that window: #69 in **v2.1.7**, #71 in **v2.1.8** (match reasoning heads on the `encrypted_content` blob), #73 in **v2.1.9**. That single boundary accounts for 51.4% → 4.5% mismatches and 16.3% → 69.9% continuation, with no part of this PR applied — and note the connection count *falling* while decisions rose, which is what fewer new heads looks like.

So our best reading of your before/after is that the two measurements were taken against different clodex versions — the "before" on **≤ 2.1.7** and the "after" on **≥ 2.1.8** — rather than with and without this patch. That would also explain why `function_call` looked like the dominant class to you: pre-#71, comparison stopped early at a reasoning item (93.6% of our pre-boundary mismatches were `reasoning`/`reasoning`), and once reasoning could match, comparison ran *deeper* into history and began surfacing whatever came next. In our corpus that class goes 0 → 37 across exactly the same boundary.

On the "zero `function_call` mismatches after" result specifically: our post-fix rate is 0.19% of decisions, so over your 2,038 decisions (1,334 + 142 + 562) you would expect roughly **4**. Seeing 0 instead of ~4 in a single run isn't a strong signal, particularly since the two runs weren't the same workload.

## The residual class you found is real — and the cause is ours

This is the part worth your attention. Those 37 are not noise, and they have a very specific shape: **37 of 37** sit at `firstMismatch == expectedPrefixItems - 1` — always the final item of the expected prefix, i.e. the head's last assistant `function_call` compared against the client's echo of it. That is precisely the place you pointed at.

The cause is a clodex-internal asymmetry that has nothing to do with ephemeral fields. `sanitizeToolInput` (`src/sdk-adapter.ts:416-427`) strips `null`-valued keys and non-required empty arrays from tool input on the **outbound** leg to Claude Code (`:788`, `:806`, `:1027`) — it exists so OpenAI models emitting filler nulls don't break Claude Code tools like `WebSearch`, where an empty domain list is a 400. But the head snapshots the **raw upstream `arguments` string**. So Claude Code faithfully echoes back what it was given, which is strictly smaller than what the head holds, and the chain is lost.

Reproduced against `main`, driving the real transport:

```
head arguments   : {"path":"file.ts","offset":null}
Claude Code echo : {"path":"file.ts"}          -> previous_response_id = undefined, new socket   (chain lost)
control (raw echo): {"offset":null,"path":"file.ts"} -> previous_response_id = resp_head, input items = 1 (continues)
```

The same test run against this PR's head produces the identical failure — **this change does not fix it.** Filed as #84, with you credited for surfacing it.

## If you'd like to re-target this

The compare-side instinct is right, just aimed one level too high: the divergence is *inside* the parsed `arguments`, not in the item's own keys. Stripping nullish values within the parsed JSON before canonicalizing makes the sanitized echo continue (verified). That only covers the `null` half though — `sanitizeToolInput` also drops non-required empty arrays, and that needs the tool's `required` set, which this module doesn't have. So the more durable fix is probably on the snapshot side: store what clodex actually emitted downstream rather than the raw upstream `arguments`. Happy for you to take it if you want it, or we'll pick it up.

## The diagnostics half is worth keeping

Independently of the above, the hashes and the dump are genuinely useful and we'd take them. A few things to tidy first:

- **The "(0600, local-only) debug log" comment is incomplete.** The mode is right, but `clodex claude --trace` ends by calling `printTraceLog(...)` (`src/cli.ts:641`, `:1032`, `:1419`), which re-reads that file and prints matching lines to stdout — so raw conversation content can reach the terminal.
- **Dump lines are uncapped** (400,116 chars measured on one item). Every other content-bearing writer in `trace-log.ts` caps (200/500, 2000 with a `[truncated]` marker, 240 for `CLODEX_LOG_REQUEST_PREVIEW`). An `input_image` data URI here would be brutal. The canonicalization also runs even when tracing is off, since `onDebug` is a truthy no-op.
- **`CLODEX_MISMATCH_DUMP` is undocumented.** The CLAUDE.md section covering this file documents `CLODEX_WS_MAX_CONNECTIONS`/`CLODEX_WS_MAX_NURSERY_CONNECTIONS`; this belongs there too, with the privacy tradeoff called out.
- Minor: the hash suffix is conditional, not "always" as the description says, and prints `none` when only one side has an item. When the histories agree over the whole comparable range, `firstMismatch` is an end-of-range index and the dump renders the bare word `undefined`.

## Verified clean (so you don't have to re-derive it)

- `normalizeToolCallJson` is genuinely compare-only — all six callers traced, the upstream delta is a raw slice of `payload.input`, and stored heads keep raw items. Nothing you changed reaches the wire.
- The dump indexes the right array. We suspected `details.firstMismatch` might be an index into a differently-built array than `[...requestInput, ...expectedAssistant]`; it isn't — both sites build the prefix identically, confirmed by hashing the dumped bytes and matching them to the reported hashes.
- No reachable over-strip in the change itself: `call_id` is preserved, reasoning identity (`encrypted_content`) is preserved, and `output: null` vs `output: 'x'` still differ.
- The do-not-restructure invariants are untouched: head capture, longest-exact-prefix continuation, full-context fallback, and the single `previous_response_not_found` retry.

## One safety note if any version of this lands

The over-strip direction has no test coverage at all. Replacing the whole five-way type gate with `if (true)`, or additionally deleting `call_id`, `arguments`, or `output`, leaves all 70 tests in `tests/responses-websocket.test.ts` green. Because this block sits on the equality predicate that gates `previous_response_id`, an over-broad edit here doesn't cost a cache hit — it makes a genuinely divergent history compare equal to a stored head and sends only the tail. Worth one test asserting that two same-kind items differing in a *meaningful* field are still rejected.

Also, since the comment claims both sides now "normalize identically" — they don't quite. `withoutEphemeralFields` also strips `role`, and `expectedAssistantItems` snapshots assistant text as a hand-built `{role:'assistant', ...}` message that the new block doesn't cover.

Thanks again — the hash addition is what made this diagnosable, and you surfaced a live cache bug we'd have kept missing.

zachspartofaday:
This is the kind of review that's worth more than the PR it landed on — thank you. You're right, and I've retargeted rather than defended.

**Conceding the reachability point in full.** I re-derived your chain locally and it holds at every link: single wiring site, `convertToOpenAIResponsesInput` on the pinned 4.0.11 emitting `{type, call_id, name, arguments}` unconditionally, no `itemId` anywhere, no other Responses ingress. My regression test was hand-crafting items the real pipeline cannot produce — exactly the failure your CONTRIBUTING quality bar leads with, so that one stung appropriately.

**And conceding the measurement critique.** After your review I rebuilt the longitudinal picture from every local session log rather than the two runs I'd compared: the real break in our corpus is at a stock version update days before the patch, and the sessions immediately before the patched build were already running at roughly the same zero-cache and cached-share rates as the sessions after it. The "~55% → 5–10%" framing in my earlier comments was not a clean with/without comparison and I retract it. Your table told the true story; mine was two points sampled from either side of a confound. (The workload also changed shape between those runs — we'd deployed staggered subagent launches at the same time, which cuts exactly the in-flight-collision isolation you'd expect to dominate a fan-out workload.)

**Retarget, per your offer.** The PR now fixes #84 on the snapshot side, and there's one discovery worth flagging: the obstacle you named — this module not having the tool's `required` set — dissolves on inspection. The request's own `tools` array carries each function tool's schema as `parameters`, and `prepareFunctionTool` passes `inputSchema` through unmodified (`strict` only when explicitly set, which this codebase never does). So `expectedAssistantItems` can consult the exact `required` sets `sanitizeToolInput` consulted, with no cross-module plumbing, and both halves — null filler and non-required empty arrays — are covered. Your repro shape is the first new test and fails on `main`.

**Safety note taken.** The equality predicate now has negative coverage: a genuinely different argument value and a different `call_id` each assert a new chain. I verified they bite — the `if (true)` + `delete call_id` mutation that previously left the suite green now fails.

**Diagnostics tidy-ups all applied:** dump lines capped at 2,000 with `[truncated]`, `(absent)` instead of the bare `undefined` at end-of-range, the "local-only" comment corrected to note `--trace` re-prints to stdout, and `CLODEX_MISMATCH_DUMP` documented in CLAUDE.md with the privacy tradeoff.

Rebased on v2.2.0; 1,280 tests + typecheck green. The branch name is historical — renaming would close the PR and lose this thread.

bman654 (changes requested):
This is very close. The redesign is the right one, the fix is correct, and I'd rather flag a short hygiene pass now than have any of it land as folklore. Nothing below is structural — it's four small things and a documentation correction that is **our** mistake, not yours.

First, credit where it's due. You dropped the inert compare-time strip, kept the diagnostics half, and went after the real cause. That's exactly the outcome the last round was hoping for, and the diagnosis in the PR body is accurate.

## The fix is right, and here's the evidence rather than an opinion

Four independent lenses went looking for a way this breaks. **No BLOCKER, no MAJOR.** The result I care most about is the negative one: the scary failure mode — continuing a chain the server never had, because the snapshot got stripped until it accidentally matched some *other* echoed call — **cannot happen**. Over-stripping makes the snapshot smaller than the echo, under-stripping makes it larger, and both directions fail safe to a new chain. That was established by mutation, not by argument.

Two more things verified by execution: the change is genuinely compare-only (every byte sent upstream is a verbatim client item in a literal tail slice), and the `ws_head_decision` schema is byte-identical to base, so our production ledger stays comparable across this boundary.

**Differential harness.** Your new tests hand-write the sanitized echo, so I built the missing half: drive the *real* downstream sanitizer (`generateAnthropicResponse` → `sanitizeToolInput`) against the *real* WS snapshot path over 11 filler shapes, with neither side reimplemented.

- your branch: **11/11 agree**
- `main`: **7/11 fail** — and the 4 that pass are exactly the shapes with no top-level filler

**Production ledger.** 11,399 head decisions, 462 `history_mismatch_new_head`, of which **42 carry the #84 signature** (`function_call → function_call` at the tail index), 42/42 at the tail. The bug is real, it's live, and this closes it.

Your test matrix deserves a specific mention: covering the **required** empty array that must be *kept*, plus two over-strip guards on value and `call_id`, is the part people usually skip. Reverting the source fails exactly your 2 positive tests and leaves the 3 guards green — they're real regression tests.

## Asks

**1. MINOR — a blank `arguments` string still loses the chain, inside the function this PR adds.** `src/oauth/responses-websocket.ts:1055-1057`

`ai@7.0.22` (`dist/index.js:3861`) has an explicit branch: `toolCall.input.trim() === "" ? {} : …`. So for a zero-argument tool the client receives `{}` and echoes `arguments:"{}"`, while your snapshot keeps `""` because `JSON.parse("")` throws and the guard returns the item untouched.

Reproduced on your head, through the real WS path — `previous_response_id = undefined`, new socket, full history resent — for both `""` and `"   "`. **Not a regression**, `main` does the same. It matters because it produces the *identical* tail-index signature as #84, so once the obvious cause is gone the remaining signature-matching records will be quietly mis-attributed to filler. Zero-property tools (many MCP tools) are the population at risk. One line, mirroring the SDK:

```ts
const raw = item.arguments.trim();
let parsed: unknown;
try { parsed = raw === '' ? {} : JSON.parse(raw); } catch { return item; }
```

**2. MINOR — the `--trace` sentence is wrong, and that one is on us.** `CLAUDE.md:73` and the comment at `src/oauth/responses-websocket.ts:554-559`

Our last review told you `printTraceLog` can echo dumped content to stdout. You wrote it up faithfully. **It's false, and we're sorry for the detour.** The dump goes to the *adapter* log (`getSessionLogPath('claude-proxy-debug')`); `printTraceLog` has three call sites (`cli.ts:662`, `:1053`, `:1440`) and every one is handed the *Claude Code* debug log. Different files. `readFileSync` over a log appears exactly once in `src/`.

Worth correcting rather than deleting, because the true story is a **stronger** privacy story and the current wording pins 0600 on the wrong file:

- it takes a double opt-in — `--trace` **and** `CLODEX_MISMATCH_DUMP=1`
- the write path runs `redactTraceLine`, so bearer tokens and `sk-`/`sk-ant-`/`AIza`/`gsk_` keys are scrubbed
- the real exposure is the durable on-disk artifact at the path clodex prints as `Adapter debug log:` — the path users paste into bug reports
- `proxy-debug.log` is `0600`; `claude-debug.log` is `0644`

Since the paragraph is added by this PR, easiest to fix it here.

**3. MINOR — share the strip rule instead of duplicating it.** The rule now lives twice: `sanitizeToolInput` (`sdk-adapter.ts:416`) and `sanitizedCallArguments` (`responses-websocket.ts`). They agree today — I verified `@ai-sdk/openai@4.0.11` passes `parameters: tool.inputSchema` through unmodified and omits `strict` unless set — but nothing pins them, and a divergence reintroduces #84 silently under a different filler shape.

To be clear about one thing so it doesn't become received wisdom: **the import cycle does not block this.** I tested it — adding the direct import plus the export gives clean typecheck, 162 tests passing, a successful build, and a working runtime import; both functions are called at request time, so there's no TDZ hazard. The refactor is worth doing on maintenance grounds, not because the cycle forbids it.

Suggested shape: a leaf module (e.g. `src/tool-input-sanitize.ts`) importing nothing from `src/`, exporting the rule with the signature both call sites already use. `sdk-adapter.ts` uses it verbatim; `responses-websocket.ts` keeps your thin wrapper for parse/serialize/blank-string handling. The two *derivations* of the required set read genuinely different data and can't usefully share code — so a pinning test is what covers those. If extracting from a DO-NOT-RESTRUCTURE file feels too invasive, the cheaper equivalent is one table-driven test feeding `[raw, requiredSet]` pairs through **both** paths and asserting equality; a test can import both regardless of cycle direction.

**4. MINOR — commit 2 has no tests.** Nothing asserts the summary-line hashes, the env gate, the 2000 cap, `[truncated]`, or `(absent)`. Mutating the gate to always-on, shrinking the cap to 20, and deleting the hash append simultaneously leaves **1280/1280 green**. The always-on direction is the one that matters: it would start writing conversation content to `proxy-debug.log` for every user with CI silent. The ledger side already pins this exact data (`tests/responses-websocket.test.ts:1524` asserts `expectedHash: expect.any(String)`); the new content-bearing channel deserves the same.

## Optional, take or leave

- `out[key] = value` at `:1060` invokes the `__proto__` setter, so `{"__proto__":{…}}` silently vanishes from the snapshot. **No reachable harm** — `out` is function-local and immediately stringified, and the downstream SDK rejects such arguments anyway. Pure hygiene, but it's a prototype-mutating assignment on model-controlled keys in a hot path: `Object.create(null)` costs nothing.
- `&& log` doesn't mean "a logger is active" — `makeProxyLog(false)` returns a defined no-op, so with the env var set and no `--trace` both items are fully canonicalized and discarded (the 2 KB cap truncates *after* the full string is built). Measured ~9% on a pathological mismatch-heavy run; negligible in practice, and resolving the env knob once at `createResponsesWebSocketFetch` time would match the file's existing pattern.
- When a client replays a history identical to the head, both dump sides render `(absent)`. Gating the dump on the same `expectedHash || actualHash` condition the line above already uses removes the noise and a dead `index >= 0` branch.
- Non-object `arguments` (scalar, array, malformed) return untouched while the downstream still transforms them. The conservative guards are the **right** call — don't change them — but the turns are silent and will be mis-attributed by the #84 signature. A `sanitizedArgumentsSkipped` field on the head-decision event would make the class countable.

## Checked and cleared — so you don't re-derive these

- **Key order and whitespace cannot cause a mismatch.** `normalizeToolCallJson` canonicalizes `arguments` through `JSON.parse` + `canonicalJson`.
- **The two required-set sources agree today.** `@ai-sdk/openai@4.0.11` (`dist/index.js:844`, `:5860`) serializes `parameters: tool.inputSchema` unmodified and omits `strict`.
- **`ctx.originalPayload.tools` is never missing at snapshot time**, so required empty arrays can't be wrongly stripped on a live path.
- **No over-strip or false-match hazard exists** — every divergence fails safe to a new chain.
- **Invariants intact.** Chain-head validation, the `previous_response_not_found` single retry, `store:false`, reasoning `encrypted_content` identity, pool/nursery promotion and eviction reasons are all behaviorally unchanged.
- **Perf is a non-issue.** Rebuilding the required map costs ~1.3% of a `JSON.parse` the path already pays, once per turn.
- **The namespace branch can't collide today** — `translateTools` sets no `providerOptions`, so `@ai-sdk/openai` never emits a `type:"namespace"` tool. Keying by bare name is the correct key regardless. Noted so nobody rediscovers it as new.

Happy to take the CLAUDE.md correction off your hands if you'd rather not carry our mistake — say the word. Otherwise this is a merge as soon as the blank-string case and the doc sentence are handled.

zachspartofaday:
All four asks are addressed in `ce667a9`, plus three of the optional items.

**1. Blank arguments** — mirrored the SDK: `raw === '' ? {} : JSON.parse(raw)` after a trim, so a zero-argument tool's `"{}"` echo now matches its head. Two continuation tests drive both shapes (`""` and `"   "`) through the WS path; reverting the source line fails exactly those two and nothing else.

**2. The `--trace` sentence** — rewritten in both CLAUDE.md and the inline comment with the story you corrected: double opt-in (`--trace` **and** `CLODEX_MISMATCH_DUMP=1`), the write path running `redactTraceLine`, the durable mode-0600 adapter log as the real exposure (the file users paste into bug reports), and never re-printed to stdout (`printTraceLog` reads the separate Claude Code debug log). No detour regretted — the corrected paragraph is a stronger privacy story, as you said.

**3. Shared strip rule** — took the extraction, since you'd already proven the cycle is a non-issue: `src/tool-input-sanitize.ts` is a leaf module importing nothing from `src/`, and both `sdk-adapter.ts` and the snapshot wrapper now call the same `sanitizeToolInput`. The module header states why the rule lives there exactly once. The shared rule also builds its output with `Object.create(null)`, which folds in your `__proto__` hygiene note and covers both call sites.

**4. Commit-2 test coverage** — five new tests: the hash suffix's shape on the summary line, the env gate in the off state, dump content on opt-in, the 2,000-char cap with `[truncated]`, and `(absent)` for a truncated-replay side. Verified by mutation that each guard bites: making the gate always-on fails 1, shrinking the cap to 20 fails 3, deleting the hash append fails 1.

**Optional items taken:** the env knob resolves once at `createResponsesWebSocketFetch` time (matching the connection-cap pattern), and the dump is gated on the same `expectedHash || actualHash` condition as the suffix — the `(absent)`/`(absent)` noise case and the dead `index >= 0` branch are gone.

**Optional item left:** `sanitizedArgumentsSkipped` on the head-decision event. Left out deliberately to keep the ledger schema byte-identical across this boundary per your comparability note — happy to add it here or in a follow-up if you'd like the class countable.

1,287 tests + typecheck green at `ce667a9`.

zachspartofaday:
Commitlint flagged the round-2 commit header at 101 characters, so it's amended (header shortened, diff and body content unchanged): `ce667a9` → `b26c539`. The green-test claims in the comment above now refer to `b26c539`.

bman654 (approved):
Approving — all four asks are addressed, and I re-verified against the head rather than the commit message.

**1. Blank arguments.** `raw === '' ? {} : JSON.parse(raw)` behind a `.trim()`, mirroring `ai@7.0.22` (`dist/index.js:3861`) exactly. Re-ran the differential harness with both blank cases added: **13/13**, where `""` and `"   "` previously lost the chain outright.

**2. The CLAUDE.md correction is better than what I asked for.** You didn't just delete the false sentence — you replaced it with the accurate mechanism: adapter log vs Claude Code debug log as separate files, the double opt-in, `redactTraceLine` scrubbing token shapes, and 0600 attributed to the file that actually has it. That paragraph is now a genuinely useful privacy note instead of a wrong one. Thanks for carrying our mistake without complaint.

**3. The shared rule — and the pinning test turned out to be unnecessary.** `src/tool-input-sanitize.ts` as a true leaf (imports nothing from `src/`), duplicate deleted from `sdk-adapter.ts`, both consumers on the one function.

I asked for a test pinning the two implementations together. You didn't add one, and after mutating I don't think you should: sharing the rule collapsed the risk the test existed to cover.

- Mutation — drop the `requiredProps` exemption inside the shared rule: fails from **both** sides independently, `writeAnthropicStream > preserves an intentional empty array for a schema-required property` and `keeps a required empty array in the snapshot and still continues`. A fork of the rule now breaks both consumers by construction.
- Mutation — make the WS layer's `requiredToolProps` return an empty map, i.e. exactly the sourcing divergence I was worried about, with `sdk-adapter` untouched: caught by your own `keeps a required empty array in the snapshot and still continues`.

So both the rule and its sourcing are pinned. Ask withdrawn.

**4. Diagnostics coverage.** Hash append, env gate off, dump on, the 2000 cap with `[truncated]`, and `(absent)`. The mutation that previously shipped green — gate always-on, cap 20, hashes deleted — now fails.

You also took three of the four optional nits: `Object.create(null)`, the env read resolved once at transport creation, and gating the dump behind the same `expectedHash || actualHash` condition (which removes the `(absent)/(absent)` noise and the dead `index >= 0` branch in one move).

**State:** 1287 tests passing, typecheck clean, CI green on `b26c539`, and it merges into current `main` with zero conflicts. I verified `ce667a9` first and you'd already amended to `b26c539` — trees are byte-identical, so everything above holds.

Nice work on this one. The redesign was the right call, and #84 was ours — you found it, then fixed it.

PR #82, opened 2026-08-07
fix(proxy): make upstream retry budget configurable

### LINKED ISSUE #81: Fan-out 429 storms exhaust the fixed maxRetries=2 and kill client agents — env-tunable upstream retry policy?
## Summary

Under concurrent-subagent workloads (Claude Code orchestration fanning out 10–20 agents on ChatGPT-plan OAuth), upstream 429 storms exhaust the AI SDK's default `maxRetries=2` inside clodex, surface downstream as 429s, then exhaust Claude Code's own retry budget — and the agent dies with a terminal API error. Session-ledger data from real workloads (`response_usage`/`upstream_error` events):

- One 15-agent sweep today: 145 `upstream_error` events, 138 of them status 429 → `AI_RetryError` (`translation_failed`), correctly mapped downstream as 429 with retry-after.
- Heavier historical days: 2,932–7,443 such events per session ledger during large fan-outs.
- The remainder is small: two true 500s and two `websocket_transport_error` (1006) in today's sweep.

clodex's classification is already right (pre-stream rate limits go downstream as 429 + retry-after; `upstream-error.ts` handles the mid-stream-can't-retro-signal case as documented). The gap is that two bounded SDK retries are far too few to ride out a burst that the client's own retry budget also can't cover, so a transient plan-level rate limit becomes an agent death.

## Proposal

An env-tunable retry policy for the upstream call, following the existing knob pattern (`CLODEX_WS_MAX_CONNECTIONS`-style): e.g. `CLODEX_UPSTREAM_MAX_RETRIES` (default 2, current behavior unchanged) and honoring upstream `retry-after` between attempts where present. Fan-out operators could then set a higher ceiling so storms are absorbed inside clodex instead of cascading to the client.

Out of scope / understood as inherent: mid-stream failures after first output (can't change an already-sent status; #68 already recovers provider messages there) and true transport drops (#29 covers pre-frame retries).

Happy to PR this (small, config + adapter call site + tests) once #80 settles — filing as an issue first per CONTRIBUTING since it changes behavior users may depend on.

camjac251:
## Summary

- add a bounded retry override across streaming and non-streaming provider calls
- preserve the SDK default when the setting is unset, blank, or malformed
- clamp valid values above 5 and emit one process-level warning
- retain provider retry headers, exponential fallback timing, and the existing 120-second idle deadline

## Why

Concurrent request bursts can outlast the default retry budget and turn transient throttling into terminal worker failures. A strict cap prevents the override from consuming the streaming deadline before a response can begin.

## Verification

- `corepack pnpm typecheck`
- `corepack pnpm test` (84 files, 1,313 tests)
- `corepack pnpm build`
- focused coverage for zero, whitespace, malformed, bounded, and over-cap values
- `git diff --check`
- `gitleaks git --staged`

Closes #81


zachspartofaday:
We independently built the same knob locally right after filing #81 (same env name, same unset/invalid → SDK-default semantics) and have been running it in real orchestration sessions at `CLODEX_UPSTREAM_MAX_RETRIES=8` — so consider the semantics field-validated. This version is better-factored than our local patch in one way worth keeping: deriving the 0–8 bound from the ten-minute request ceiling is the right call (our local range allowed budgets that would die of the total timeout mid-backoff anyway).

Motivating data from before any mitigation, if useful for the description: in one session with 15+ concurrent subagents, 138 of 145 `upstream_error` events were 429s that exhausted the fixed two-retry budget and surfaced as `AI_RetryError`; heavy days logged 3–7k such events per session ledger, with the overflow terminating client agents. We'll retire our local patch when this lands. +1

bman654 (changes requested):
Thanks — the wiring here is clean and the problem in #81 is real. One thing needs changing before this lands, and it's specifically the upper half of the documented range.

## The 0–8 range is calibrated against the wrong clock

The rationale in both CLAUDE.md and `upstream-retry.ts` is that eight retries produce 510s of backoff, which "fits inside the adapter's ten-minute total timeout". The arithmetic is exactly right — `initialDelayInMs = 2000`, `backoffFactor = 2`, so 2+4+8+16+32+64+128+256 = 510s, and `SDK_TOTAL_TIMEOUT_MS` is indeed 600s.

But that isn't the binding clock. `streamAnthropicResponse` arms a **120s idle timer** (`SDK_STREAM_IDLE_TIMEOUT_MS`, `src/sdk-adapter.ts:866`) *before* calling `streamText`, passes its signal in as `abortSignal`, and only resets it when a stream part arrives (`:887-890`). Retry backoff yields no parts, so the timer runs uninterrupted across the whole retry sequence — and `prepareRetries` wires that same signal into the retry loop, so it aborts mid-backoff.

Driving the shipped retry helper against that timer:

```
ATTEMPT_OFFSETS_MS=[0,2000,6000,14000,30000,62000]
ATTEMPTS=6 (1 initial + 5 retries)
OUTCOME=Delay was aborted
```

The 7th attempt would land at 126s. It never happens. So on both Claude Code paths the effective ceiling is **5 retries**, and 6/7/8 are unreachable.

That would be merely cosmetic if the excess were free, but it isn't. At `maxRetries >= 6` the request stops failing fast with the upstream's 429 + `retry-after` — which Claude Code knows how to act on — and instead stalls for a full 120s before dying with a generic `no data received from provider for 120s`. For the fan-out scenario in #81 that's a strictly worse outcome than the default, and an operator reading the README will reasonably pick 8.

To be precise about the boundary: **5 is fine** (all attempts complete by 62s, intact). The problem starts at 6.

Any of these works for us — your call:
- cap the range at 5 and adjust the docs, or
- reset/extend the idle timer across retry backoff so the 600s budget is genuinely the ceiling, or
- keep 0–8 but document that values above 5 trade a fast, actionable 429 for a 120s stall on the streaming paths.

Whichever you choose, the CLAUDE.md and README rationale needs to name the 120s idle timeout rather than the 600s total, since that text becomes the canonical explanation in this repo.

## Verified clean — worth stating, since these were the obvious objections

- **No duplicate-output hazard.** We checked whether the SDK can replay after chunks have been yielded: in `ai@7.0.22` the retry wraps only the `doStream` call, the body is never consumed inside it, and `@ai-sdk/openai` throws only before the first output chunk. The README's "cannot be replayed safely after output begins" is accurate, not just asserted.
- **It does reach the OAuth WebSocket path.** We expected this to be a no-op for ChatGPT/Codex OAuth models given the custom transport, which would have meant it didn't fix #81 at all. It isn't: the WS fetch returns 200 with an in-band error frame, which the error helper converts into a retryable 429, and `maxRetries=8` was measured producing 9 fetch calls over 510s on a path with no idle timer. Good.
- **Completeness holds.** There are 6 `streamText`/`generateText` sites in `src/`, not 5, and all 6 are wired. No other SDK generation entry point exists.
- **Test counts match your claim exactly**: 83 files / 1,255 tests, typecheck clean. Every one of the 6 wiring reverts is killed by a test.
- `maxRetries: undefined` correctly falls through to the SDK default (`maxRetries != null ? maxRetries : 2`), and `prepareRetries` would throw on a non-integer — so returning `undefined` on invalid input is the right shape.

## Smaller things

1. **`src/openai-adapter.ts` has no `abortSignal` and no total timeout.** The three sites there (`:208`, `:216`, `:252`) pass `maxRetries` but nothing can cancel them, and unlike `sdk-adapter.ts` there's no idle timer and no `SDK_TOTAL_TIMEOUT_MS`. So on the OpenAI-format gateway route `maxRetries=8` really can block a request for ~510s with no way out — the documented "fits inside the ten-minute total timeout" doesn't describe this surface at all, because there is no such timeout here.

2. **The invalid-value warning is effectively unreachable.** Only `streamAnthropicResponse` (`:881`) passes a logger; the other five sites call `upstreamMaxRetries()` bare. Even at that one site, `src/server/router.ts:445` passes literal `undefined`, and `src/proxy.ts:644` passes `plog`, which `makeProxyLog` returns as a no-op unless `--trace` is on. Net effect: a user who sets `CLODEX_UPSTREAM_MAX_RETRIES=lots` gets silence on every route unless they happen to be tracing. Since the module-level `reportedInvalidValues` Set exists specifically to avoid spamming, it would be worth having something to spam — stderr once, like the WS pool knobs do.

3. **Two surviving mutants** worth a test each:
   - Deleting the logger wiring at `:881` — the only logger in the feature — leaves the full suite green.
   - Dropping the `.trim()` in `upstreamMaxRetries` leaves the suite green, but changes behaviour materially: with it, `CLODEX_UPSTREAM_MAX_RETRIES='   '` hits the `raw === ''` guard and keeps the default; without it, `Number('   ')` is `0`, which is a valid in-range value and silently *disables* retries.

No concerns about the default path — unset behaviour is unchanged, and that's well covered.

zachspartofaday:
Chiming in as the #81 reporter: we applied the review's finding to our local build the same day, and can offer one operator's data point on the three options.

We went with **cap at 5, clamping higher values with a one-time stderr warning** rather than documenting the tradeoff — and having run both configurations, I'd argue for that option here too. Before the review we were running `=8` in production orchestration sessions, which per the 120s-idle analysis means we were unknowingly sitting in the stall zone: the exact operator failure the review predicts ("an operator reading the README will reasonably pick 8") is not hypothetical — it's what we did. A knob whose top documented value silently degrades behavior below its middle values is a trap documentation doesn't reliably defuse; the clamp turns the mistake into a one-line warning instead.

Two small corroborations from porting the findings onto our equivalent patch:

- The review's **six-site count is the right checklist** — auditing against it exposed that our local version had only wired five (we'd missed the `generateText` on the OpenAI-gateway non-stream branch, which is also the surface with no abort signal or total timeout at all, i.e. exactly where an uncapped budget hurts most).
- The **whitespace-only `.trim()` mutant is worth its test**: `Number('   ')` is `0`, so losing the trim silently disables retries rather than erroring. We added the regression test and it's cheap.

Running clamped `=5` live now; happy to report back once it's been through a real burst. Still planning to retire our local patch the moment this lands.

camjac251:
Review update:

- Valid retry values now clamp at 5; larger values emit one process-level warning.
- Blank, whitespace-only, and malformed values preserve the SDK default instead of becoming zero.
- The trace-only logger dependency was removed so the clamp warning remains visible in normal runs.
- Documentation now reflects the existing 120-second idle deadline.

Verification: typecheck; build; focused boundary coverage; 84 files / 1,313 tests; diff check; staged leak scan.

bman654 (approved):
Re-reviewed at `bb67f7d`. **The blocking finding is fixed** — verified by measurement, not by reading the commit message. Thanks for taking the idle-timer argument seriously; the clamp lands in exactly the right place and the docs now name the real deadline. Everything below is optional polish.

---

## The blocker is resolved

I re-ran the same measurement that produced the original finding, this time driving the real `retryWithExponentialBackoff` from `@ai-sdk/provider-utils` (ai 7.0.22) and reading the deadline out of the shipped source instead of a literal:

```js
const IDLE_MS = Number(/const SDK_STREAM_IDLE_TIMEOUT_MS = ([0-9_]+);/
  .exec(readFileSync('src/sdk-adapter.ts','utf8'))[1].replace(/_/g,''));
// IDLE_MS=120000  MAX_UPSTREAM_MAX_RETRIES=5
```

Driving a retryable 429 against an abort signal armed at `IDLE_MS`, exactly as `streamAnthropicResponse` arms it before `streamText`:

| configured | attempt offsets (ms) | attempts | outcome |
|---|---|---|---|
| `5` (the new ceiling) | `[0, 2000, 6000, 14000, 30000, 62000]` | 6 | `Failed after 6 attempts. Last error: AI_APICallError: rate limited` |
| `6` (what the old head accepted) | `[0, 2000, 6000, 14000, 30000, 62000]` | 6 | `Delay was aborted` |

That is the whole finding in two rows. At `5` the budget is fully spent by 62 s, 58 s of idle headroom remain for the final attempt's first stream part, and the client gets the **actionable** provider error. At `6` the sixth backoff is still pending when the 120 s abort lands and the user gets the generic stall. The clamp removes the entire 6–8 window that produced it.

Boundary behaviour, driven through the real `upstreamMaxRetries`:

```
raw=4  -> 4  warned=0
raw=5  -> 5  warned=0
raw=6  -> 5  warned=1  "clamping CLODEX_UPSTREAM_MAX_RETRIES=6 to 5 (higher values exceed the 120s streaming idle budget)"
raw=8  -> 5  warned=1
raw=99 -> 5  warned=1
```

Clamping is **not** silent, and it is documented in both `README.md` and `CLAUDE.md`. Good.

## Previously raised minors — two are fixed

- **Warning wired at only 1 of 6 sites, and `--trace`-gated — FIXED.** All six call sites (`sdk-adapter.ts:860,930,990`, `openai-adapter.ts:208,216,252`) now call `upstreamMaxRetries()` with no arguments, so the `console.error` default applies everywhere. No route can swallow the notice any more.
- **Surviving mutants on `.trim()` and on the logger wiring — FIXED.** Both are now killed; see the mutation table below.
- **`openai-adapter.ts` has no `abortSignal`/timeout — still true, but the severity is largely gone.** `grep -n "abortSignal\|AbortController\|setTimeout" src/openai-adapter.ts` still returns nothing, so those three sites remain uncancellable. What changed is the size of the window: the worst case an operator can configure drops from ~510 s to ~62 s of backoff. I'd leave it; noting it only so it isn't mistaken for having been fixed.

## Documentation

Both files now name the binding timer and correctly say it — not the ten-minute total — is the effective ceiling. Accurate as written for the exponential-backoff case. One caveat below.

## MINOR — the ceiling and the deadline it is derived from are two unlinked constants

`MAX_UPSTREAM_MAX_RETRIES = 5` lives in `src/upstream-retry.ts`; `SDK_STREAM_IDLE_TIMEOUT_MS = 120_000` lives in `src/sdk-adapter.ts` and is not exported. The `120s` in the warning string is a third copy. Nothing connects them, and I confirmed the gap is not just theoretical:

```bash
sed -i '' 's/const SDK_STREAM_IDLE_TIMEOUT_MS = 120_000;/const SDK_STREAM_IDLE_TIMEOUT_MS = 20_000;/' src/sdk-adapter.ts
CLODEX_HOME=$(mktemp -d) pnpm test
#  -> the ceiling of 5 (62 s of backoff) is now provably wrong,
#     and no test that asserts the relationship fails
```

The same blind spot covers the other input to the arithmetic: the schedule is the SDK's `initialDelayInMs = 2e3`, `backoffFactor = 2` defaults (`@ai-sdk/provider-utils/dist/index.js:3076-3079`), so an `ai` bump that changes either silently invalidates `5` too.

The value itself *is* pinned in both directions (`5 -> 4` and `5 -> 2` both fail tests), so this is about the derivation, not the number. Cheapest fix: export `SDK_STREAM_IDLE_TIMEOUT_MS` and add one test that walks the real `retryWithExponentialBackoff` for `MAX_UPSTREAM_MAX_RETRIES` retries and asserts the last attempt starts before it — essentially the table at the top of this review, as a regression test. That also lets the warning string interpolate the deadline instead of hardcoding `120s`.

## MINOR — the "five retries complete before the 120 s timeout" claim does not hold when the provider sends `retry-after`

The README says in the same paragraph that the SDK honours `retry-after`/`retry-after-ms`, and it does: `getRetryDelayInMs` (`node_modules/ai/dist/index.js:2650`) uses the header verbatim whenever `ms < 60_000 || ms < exponentialBackoffDelay`. So the header, not the exponential schedule, sets the pace. Measured at the new ceiling of `5`:

| `retry-after-ms` | attempt offsets (ms) | outcome |
|---|---|---|
| 10000 | `[0, 10000, 20000, 30000, 40000, 50000]` | `Failed after 6 attempts…` |
| 20000 | `[0, 20000, 40000, 60000, 80000, 100000]` | `Failed after 6 attempts…` |
| 25000 | `[0, 25000, 50000, 75000, 100000]` | **`Delay was aborted`** |
| 30000 | `[0, 30000, 60000, 90000]` | **`Delay was aborted`** |
| 59000 | `[0, 59000, 118000]` | **`Delay was aborted`** |

From roughly 24 s of `retry-after` upward, the same generic-stall failure mode reappears at the new ceiling. Note the SDK's own default of 2 retries is safe even at a 59 s header (`0, 59000, 118000`, all inside 120 s), so this is strictly a property of opting in to 3+.

I'm filing this as MINOR rather than blocking: it needs both an opt-in and a co-operating provider, and 429s with tens of seconds of `retry-after` are a genuinely bad situation regardless. But the documented claim is stated unconditionally, and it isn't. Either qualify it ("…assuming exponential backoff; provider-supplied `retry-after` can still exceed the deadline") or, if you want the guarantee to be real, have the retry wrapper stop honouring a header that would push the next attempt past the remaining idle budget.

## NIT — the warning fires on the first request, where the TUI may paint over it

`console.error` runs lazily inside `upstreamMaxRetries` on the first generation call. Under `clodex claude` that lands on the shared terminal while Claude Code's TUI is already rendering, so a user who mis-set the variable may never see the line. The value is knowable from `process.env` at launch — emitting it once during startup, where clodex already prints its banner, would be more reliable. (The per-value dedup via `reportedValues` is right; keep it.)

## NIT — the clamp message names a deadline that does not apply to every route

`generateText` on the non-streaming path has only the 600 s total timer, and the three `openai-adapter.ts` sites have no timer at all. A user hitting only `/openai/v1` non-streaming will be told their value "exceeds the 120s streaming idle budget", which isn't true for their route. A single global ceiling is the right simplification — just worth a word or two of hedging in the message.

## Mutation testing

Run as full-file (`npx vitest run tests/upstream-retry.test.ts tests/sdk-adapter.test.ts`, 77 tests), not `-t` isolation:

| mutation | result |
|---|---|
| ceiling `5 -> 8` | **killed** (2 failed) |
| clamp branch deleted entirely | **killed** (3 failed) |
| clamp made silent (warning removed) | **killed** (3 failed) |
| default `console.error` warn `-> () => {}` | **killed** (1 failed) |
| `.trim()` removed | **killed** (1 failed) |
| ceiling `5 -> 4` | **killed** (3 failed) |
| ceiling `5 -> 2` | **killed** (5 failed) |
| `SDK_STREAM_IDLE_TIMEOUT_MS` `120_000 -> 20_000` | **survived** — the MINOR above |

Seven of eight killed. That is a real improvement over the previous head, where the `.trim()` and logger-wiring mutants both survived.

## Build and test results

```
CLODEX_HOME=$(mktemp -d) pnpm typecheck   # clean
CLODEX_HOME=$(mktemp -d) pnpm test        # 84 files / 1313 tests passed
CLODEX_HOME=$(mktemp -d) pnpm build       # Build success, dist/cli.js 2.79 MB
```

(For reference, two tests — `credential-helper > force-kills a helper that exceeds the runtime limit` and `http-proxy-index > records the standalone proxy server start and clean shutdown lifecycle` — fail when those two files are run as a standalone subset, on an unmodified tree as well. Neither imports `sdk-adapter`; it's load/ordering sensitivity that predates this PR, not anything you introduced.)

---

**Verdict:** the blocking concern is closed on the evidence above. The two MINORs are worth a follow-up commit if you're willing — the constant coupling especially, since it's a few lines and it turns the reasoning in your commit message into something the suite enforces — but neither should hold this up.

PR #83, opened 2026-08-07
fix(transport): replay control-only socket failures

camjac251:
## Summary

- gate replay on whether any downstream or payload data was emitted, not on a control-frame count
- retain cumulative control-frame diagnostics across attempts
- keep failures terminal after any caller-visible output
- emit retry-exhaustion diagnostics only for transport retries

## Why

A connection can emit several buffered control frames and still fail before any caller-visible output. Replay is safe in that state; a numeric frame allowlist rejected valid recovery and could also pair exhaustion diagnostics with unrelated continuation retries.

## Verification

- `corepack pnpm typecheck`
- `corepack pnpm test` (83 files, 1,297 tests)
- `corepack pnpm build`
- focused coverage for four buffered control frames, payload-terminal behavior, cumulative diagnostics, and continuation retry separation
- `git diff --check`
- `gitleaks git --staged`


bman654 (changes requested):
Thanks for this. The reasoning is careful and the implementation is genuinely surgical — this file is marked do-not-restructure and you respected that. We verified the safety properties you'd want verified, and they hold (details at the bottom, so you don't have to re-derive them).

The problem is narrower and more awkward: **the widened window doesn't appear to catch any failure that actually happens.**

## The new window has no observed hits; the real pre-output failures stay terminal

`transportReplaySafe` opens replay for `frameCount === 0`, or exactly one frame that is `response.created`. We checked that against a local corpus of roughly 33,000 WebSocket requests (`ws_head_decision` 33,492 / `ws_response_usage` 33,318, 135 `ws_response_error`):

```
ws_response_error (source, frameCount, emittedModelData, retried, closeCode, errCode) -> count
   ('socket_error',        0, False, True,  None, 'ENOTFOUND') -> 88
   ('unexpected_response', 0, False, False, None, None)        -> 24
   ('socket_close',        4, False, False, 1006, None)        -> 4    <-- the only pre-output closes
   ('socket_close',        5, True,  False, 1006, None)        -> 3
   ('socket_close',       19, True,  False, 1006, None)        -> 2
   ... every remaining record has emittedModelData=True (genuinely mid-stream)

records at frameCount == 1: 0
```

There is not one transport failure at `frameCount === 1` in the corpus. Every `ws_transport_retry outcome:"started"` fired at `frameCount === 0` — the window that already existed before this PR. (The 25 `recovered` records showing `frameCount: 1` are post-retry state, not failures: pre-PR, `resetContextForRetry` didn't zero `frameCount`, so the first frame on the replacement socket makes it 1.)

Meanwhile the failures this PR is written for — abnormal close after setup but before any output — do occur, 4 of them, and they all look like this:

```json
{"event":"ws_response_error","source":"socket_close","closeCode":1006,"frameCount":4,
 "emittedModelData":false,"responseIdReceived":true,"inFlightMs":3841,"retried":false,...}
```

`response.created` arrived and three more buffered control frames followed before the drop, 3.4–4.4s in. Reproducing that exact shape against this branch (four control frames, then close 1006) yields `transport_retry records = []` — it stays terminal. So the summary bullet "replay one complete request after a pre-output socket failure" is doing more than the code does: it replays a one-frame subset, and the region that's actually hit is untouched.

## Suggested change

Drop the frame-type allowlist and the frame counting, and gate on what you actually care about — that nothing has been committed downstream and no output has accumulated:

```ts
function transportReplaySafe(ctx: RequestContext): boolean {
  return !ctx.emittedDownstreamData && !ctx.emittedModelData && ctx.outputByIndex.size === 0;
}
```

That recovers all four real cases and is, as far as we can tell, no less safe. `outputByIndex` is only ever populated by `response.output_item.added` and the `.delta` events (`captureOutput`, `:932-949`), which are exactly the events `isModelDataEvent` covers — so `emittedModelData === false` already implies nothing accumulated, and the third clause is belt-and-braces rather than a new constraint. It also removes the dependency on `recentUpstreamEventTypes`, which is a bounded 20-entry diagnostic ring (`:803-804`) and a slightly uncomfortable thing to make load-bearing for a correctness gate.

If you'd rather keep the allowlist, it needs to at least cover `response.in_progress` and the reasoning-summary part frames that show up in the real 4-frame shape.

## Verified clean

These were the things worth being suspicious about, and none of them bit:

- **No double-emit.** The abandoned `response.created` sits in `ctx.pendingEvents` and is cleared by `resetContextForRetry`. We enumerated every mutable field on `RequestContext`; each is either reset or provably untouched by a lone `response.created`. Adding `ctx.frameCount = 0` to the reset is *necessary*, not incidental.
- **No chain-head poisoning.** `entry.responseId` is set only in the terminal non-failed branch and `deleteEntry` unregisters the entry, so `resp_abandoned` can't become a head or come back as a `previous_response_id`.
- **The shared `ctx.retried` budget is safe in both orderings** — the reset clears `continued` and restores the full payload, so a transport replay followed by `previous_response_not_found` (or the reverse) can't produce two retries or surface a `previous_response_not_found` that should have been retried.
- **`encodeSse` is the only `controller.enqueue` site**, so `emittedDownstreamData` can't be bypassed.
- The four do-not-restructure guarantees are behaviourally intact.

One correction to something we initially flagged and then disproved, in case it comes up: widening the second predicate in `handleTransportFailure` looked like it might change the client-visible error, but that branch is diagnostic-only — it always falls through to `failContext`. Not a problem.

## Smaller things

1. **`emittedDownstreamData` can never be the deciding clause.** Given the other conditions, there's no reachable state where it's `true` while the rest of the predicate would still pass — `encodeSse` is reached only from `failContext` (which closes the context) and `flushPending`, and the `flushPending` cases either set `emittedModelData` or are terminal. Deleting the assignment leaves the suite green in both mutation directions. Worth keeping as defence-in-depth, but it isn't currently doing work, and the diagnostic field implies it is. Under the simplified gate above it becomes genuinely load-bearing.

2. **`exhausted` can now fire without a matching `started`.** `resetContextForRetry` zeroes `frameCount` and is also called by the `previous_response_not_found` path (`:1368`), which sets `ctx.retried` too. Previously that path left `frameCount >= 1` so the exhausted arm couldn't trigger; now it can. Relatedly, a single request can now emit `started -> recovered -> exhausted`, which was impossible before (the two were mutually exclusive). Both are diagnostics-only, but they'll confuse anyone aggregating these.

3. **`frameCount` quietly changes meaning on the `previous_response_not_found` path too** — cumulative-across-attempts to per-attempt. It's operator-facing on every `ws_response_error` / `ws_transport_retry` / `ws_head_decision`, and it's the `sequence_number` on the synthetic error event. The new meaning is arguably the better one; it just deserves a line in the PR description since it isn't scoped to the transport path.

4. **Two of the four sub-conditions survive weakening** — dropping `ctx.emittedDownstreamData`, and dropping `ctx.frameCount === 1`, both leave all 71 tests green. `frameCount === 1` is genuinely load-bearing (without it, a `response.created` plus a valid-JSON non-object frame would pass the gate).

5. **Wording.** "unknown, repeated, or emitted frames make replay terminal" reads as though a family of benign control frames is recognised and only strangers rejected; the allowlist has exactly one member, and the three other normal pre-output control frames are terminal too. Same for the replacement test name calling `response.in_progress` "unrecognized" — it's a perfectly normal frame. Worth rewording, and note the replaced test ("does not retry after any upstream response frame has arrived") was guarding a real invariant that's now gone, so it'd be good to keep an explicit test for "a frame that carries model data is always terminal".

camjac251:
Review update:

- Replay eligibility now depends on whether downstream or payload data was emitted, not on a numeric control-frame allowlist.
- Four buffered control frames followed by a transport close are covered and replay successfully.
- Control-frame diagnostics remain cumulative across attempts.
- Retry-exhaustion diagnostics are emitted only for transport retries, so continuation retries cannot create an unmatched exhaustion record.

Verification: typecheck; build; focused transport coverage; 83 files / 1,297 tests; diff check; staged leak scan.

bman654 (approved):
## Re-review of `567c5b6` — the gate now does what the PR claims

Thank you for the rework. The headline objection from the first round is **resolved**, and I verified that rather than taking the commit title's word for it. Details, commands and output below.

---

## 1. The blocking finding is resolved: the new gate recovers the real failures

Round 1's objection was not that the change was unsafe — it was that the window you opened (`frameCount === 1` plus one allow-listed event type) had **zero hits** in a ~33k-request corpus of real traffic, while the only genuine pre-output socket deaths in that corpus all arrived at **four** frames. So the previous head recovered 0 of them.

I re-mined the same diagnostics ledger to make sure I was measuring against the right target, and the four cases are still exactly four, with a very consistent shape:

```
$ node /tmp/mine83.mjs     # streams ~/.clodex/logs/sessions/*websocket-diagnostics*.jsonl
pre-output ws_response_error by source: [["unexpected_response",25],["socket_error",88],["response_event",2],["socket_close",4]]

{"source":"socket_close","closeCode":1006,"frameCount":4,"emittedModelData":false,"responseIdReceived":true,"continued":true,"retried":false,"inFlightMs":4441}
{"source":"socket_close","closeCode":1006,"frameCount":4,"emittedModelData":false,"responseIdReceived":true,"continued":false,"retried":false,"inFlightMs":3841}
{"source":"socket_close","closeCode":1006,"frameCount":4,"emittedModelData":false,"responseIdReceived":true,"continued":true,"retried":false,"inFlightMs":3403}
{"source":"socket_close","closeCode":1006,"frameCount":4,"emittedModelData":false,"responseIdReceived":true,"continued":true,"retried":false,"inFlightMs":4392}
```

I then reconstructed that shape as a test (four control frames — `response.created`, `response.queued`, `response.in_progress`, `response.reasoning_summary_part.added` — then a 1006 close) in both the fresh-request and the continuation variants, and ran the identical file against three revisions:

```
$ pnpm vitest run tests/zz-pr83-real4.test.ts        # at 567c5b6
R1 transport_retry records = [{"connectionId":1,"continued":false,"retried":true,"frameCount":4,
  "emittedModelData":false,"emittedDownstreamData":false,"responseIdReceived":true,
  "event":"ws_transport_retry","outcome":"started","source":"socket_close","closeCode":1006}]
R2 transport_retry records = [{"connectionId":1,"continued":true,"retried":true,"frameCount":4,
  "emittedModelData":false,"emittedDownstreamData":false,"responseIdReceived":true,
  "event":"ws_transport_retry","outcome":"started","source":"socket_close","closeCode":1006}]
Tests  3 passed (3)

# same test file, src/oauth/responses-websocket.ts swapped for main (827215a):
R1 transport_retry records = []
R2 transport_retry records = []
Tests  2 failed | 1 passed (3)

# same test file, src/oauth/responses-websocket.ts swapped for the previous head (f98de8ed):
R1 transport_retry records = []
R2 transport_retry records = []
Tests  2 failed | 1 passed (3)
```

The emitted diagnostic matches the ledger records field for field (`socket_close` / 1006 / `frameCount:4` / `emittedModelData:false` / `responseIdReceived:true`), now with `outcome:"started"` instead of a terminal error. **Recovery went from 0/4 to 4/4.** Your new test `retries after multiple buffered control frames when no output was emitted` pins exactly that shape, which is the right thing to pin.

The replay itself is correct in both variants: `previous_response_id` is dropped, the full original `input` is resent, and in the continuation case turn 3 continues from the **recovered** id, never the abandoned one.

---

## 2. The safety argument holds — with one correction to my own round-1 reasoning

I claimed `outputByIndex` is populated only by `response.output_item.added` and `.delta` events, so `emittedModelData === false` already implies it is empty. **That was incomplete.** `captureOutput` also seeds `outputByIndex` from `response.output` on a terminal frame (`responses-websocket.ts:1140-1153`), and terminal frames are not model-data events. That path is unreachable at failure time — a terminal frame always ends in `closeContext`, and every failure entry point short-circuits on `ctx.closed` — but my stated reason was wrong, so I want it on the record.

The argument that actually carries the gate is simpler and stronger than the one I gave: **`encodeSse` is the only `controller.enqueue` site in the file**, so `emittedDownstreamData` is a *direct observation* that no byte reached the client rather than an inference from event types.

```
$ grep -n "enqueue" src/oauth/responses-websocket.ts
1260:  ctx.controller.enqueue(ctx.encoder.encode(`data: ${JSON.stringify(event)}\n\n`));
```

That matters more than it looks, because it means the gate does not depend on `isModelDataEvent`'s enumeration staying complete. If OpenAI ships a new frame type tomorrow that carries user-visible content, it will be buffered in `pendingEvents`, never flushed (flushes happen only on a model-data event, a terminal event, `failContext`, or an unparseable frame), and therefore discarded and regenerated by the replay. The client cannot see it twice.

---

## 3. Duplicate-output probing: I tried to break it and could not

This was the highest-consequence failure mode, so I attacked it directly with six probes rather than reading the code:

```
$ pnpm vitest run tests/zz-pr83-doubleemit.test.ts
Tests  6 passed (6)
```

| probe | scenario | result |
|---|---|---|
| D2 | the abandoned socket delivers a **full** response late, interleaved with the live replay, twice | zombie output never reaches the client; exactly one `response.created` and one `response.completed` downstream |
| D3 | an unparseable frame among the control frames (which flushes the buffer downstream) then a 1006 close | terminal, no replay, `retries = []` |
| D5 | `error` **and** `close` fired on the same dying socket | exactly one `started` record, exactly one replacement socket, one copy of the output |
| D7 | the replayed attempt also dies after four control frames | terminal, buffer flushed exactly once, retry budget respected |
| D8 | chain-head integrity across a replay | `previous_response_id` on the next turn is the recovered id; the abandoned id never appears |
| D9 | a reasoning-summary delta before the close | terminal; the partial text is emitted exactly once |

Plus the ten-case replay harness I built during round 1, re-run against this head: **9 of 11 pass**, and both non-passes are intentional behaviour changes rather than defects (see §5).

On chain-head poisoning specifically, the two conditions that made it impossible before are both still in place: `entry.responseId` is assigned only in the terminal non-failed branch (`:1652-1655`), and `resetContextForRetry` clears `ctx.responseId` before the replay is dispatched (`:1497`). `createReplacement` always builds a fresh connection, so a replay can never land on an entry that is serving another request.

---

## 4. Mutation testing the new gate

Run as a **full-file** run (`tests/responses-websocket.test.ts`, 90 tests), not `-t` isolation, because those two disagree and the disagreement has hidden a hole before.

| mutation | result |
|---|---|
| M0 baseline | 90 passed |
| M4 — gate returns `true` unconditionally | **killed** — `does not retry after model output has reached the downstream stream` |
| M5 — gate reverted to main's `frameCount === 0 && !emittedModelData` | **killed** — `retries after multiple buffered control frames when no output was emitted` |
| M6 — drop `ctx.transportRetryPending` from the exhausted predicate | **killed** — `does not report transport exhaustion for a continuation retry` |
| M1 — drop `!ctx.emittedDownstreamData` | survives (90 passed) |
| M2 — drop `!ctx.emittedModelData` | survives (90 passed) |
| M3 — drop `ctx.outputByIndex.size === 0` | survives (90 passed) |
| M7 — `encodeSse` never sets `emittedDownstreamData` | survives (90 passed) |

M4/M5/M6 are the important ones and they all die: the feature is genuinely covered, over-widening is caught, and the diagnostic fix is pinned. That is a much better mutation profile than the previous head.

---

## 5. Findings

### MINOR — the `emittedDownstreamData` guard is reachable-decisive but completely untested

M1 and M7 above are the same hole seen from two sides: deleting the clause, or never setting the flag, leaves all 90 tests green. That would be fine if the clause were pure defence in depth, but it is not — it is the **only** thing standing between a real code path and duplicated downstream content.

The path is `handleSocketMessage`'s unparseable-frame branch (`:1527-1531`): a frame that fails `JSON.parse` is pushed and **flushed downstream immediately**, without ever setting `emittedModelData` or touching `outputByIndex`. So after such a frame, two of the three clauses are still `true` and only `emittedDownstreamData` blocks the replay. With M1 or M7 applied, my D3 probe flips from terminal to replayed:

```
# M1 (clause deleted):
D3 sockets = 2 retries = [{... "frameCount":2,"emittedModelData":false,
  "emittedDownstreamData":true,"outcome":"started","closeCode":1006}]
× D3: a non-JSON frame among the control frames flushes downstream and makes the close terminal

# M7 (flag never set):
D3 sockets = 2 retries = [{... "emittedDownstreamData":false,"outcome":"started",...}]
× D3: a non-JSON frame among the control frames flushes downstream and makes the close terminal
```

The client would receive the stray frame from attempt 1 *and* the complete replayed response. **The ask:** one test in the style of your new positive case — `response.created`, then a non-JSON frame, then a 1006 close — asserting a single socket and no `ws_transport_retry` record. It is a handful of lines and it converts the new field from decoration into a covered invariant.

(M2 and M3 also survive, but those two really are redundant: any event that sets `emittedModelData` or populates `outputByIndex` also flushes downstream in the same call, so `emittedDownstreamData` is already `true`. I asked for the three-clause form, so I am not asking you to trim it — a one-line comment saying the last two clauses are belt-and-braces would stop the next reader from trying.)

### NIT — the docs say "control frames"; the gate is frame-type-agnostic

`CLAUDE.md` now reads "buffered control frames do not close that safe window". The gate does not inspect frame types at all — the previous head's allow-list is gone, which is the improvement. Concretely, a typeless JSON frame is now replayable:

```
$ pnpm vitest run tests/zz-pr83-lens.test.ts
× G: a typeless JSON frame is terminal (no replay)
  expected [ FakeWebSocket, … ] to have a length of 1 but got 2
```

That is correct behaviour under the new design — the frame was buffered, never emitted, and is discarded by the reset — but the sentence describes a narrower rule than the code implements. Something like "any frame that was buffered but never emitted downstream does not close that safe window" would match what ships.

### NIT — the synthetic error frame's `sequence_number` now counts across a replayed attempt

Dropping `ctx.frameCount = 0` from `resetContextForRetry` was the right call (it restores main's semantics and un-does the per-attempt drift I flagged last round). The visible consequence is that `failContext`'s synthetic frame reports a cumulative count:

```
$ pnpm vitest run tests/zz-pr83-lens.test.ts
× C: … expected 'data: {"type":"error","sequence_number":2,…}' to contain '"sequence_number":1'
```

This matches the convention the `previous_response_not_found` path has always used, so nothing is wrong — just worth knowing when reading a log, since a transport replay could previously only happen at frame 0 and so always produced `1`.

---

## 6. Round-1 minors: all four resolved

- **"`emittedDownstreamData` can never be the deciding clause."** Resolved — it is now the deciding clause on the unparseable-frame path (§5 proves it), which is precisely why it deserves a test.
- **"`exhausted` can fire with no `started`, and `started → recovered → exhausted` is newly possible."** Resolved by adding `ctx.transportRetryPending` to the predicate, and pinned by a test (M6 kills it). Re-running the sequencing probe at this head prints `OUTCOMES ["started","recovered"]` — no spurious third record.
- **"`frameCount` silently becomes per-attempt."** Resolved; `resetContextForRetry` now matches main exactly.
- **"The replaced test dropped a real invariant."** The invariant it dropped (a socket death after `response.created` is terminal) is the very thing this PR intentionally inverts, and the invariant worth keeping — model output blocks the replay — is still asserted and kills M4. Fine as it stands; §5 is the one piece of coverage still missing.

Everything I cleared in round 1 was re-checked at this head and still holds: no double-emit, no chain-head poisoning, `ctx.retried` safe in both retry orderings, `encodeSse` still the sole enqueue site, and the widened second predicate in `handleTransportFailure` still falls through to `failContext` so it cannot change the client-visible error.

---

## 7. Build, types, tests

```
$ pnpm typecheck        # clean
$ pnpm test             # Test Files  83 passed (83)
                        # Tests  1297 passed (1297)
$ pnpm build            # ESM dist/cli.js 2.79 MB — Build success in 288ms
```

One note for anyone reproducing this: my first full run showed `1 failed | 1296 passed`, on `tests/http-proxy-index.test.ts > records the standalone proxy server start and clean shutdown lifecycle` (20s timeout). That is an artifact of my own shell — it was running bridged through clodex, so `HTTPS_PROXY` / `NODE_EXTRA_CA_CERTS` were set. With those scrubbed the whole suite is green, and the file passes in isolation either way. Nothing to do with this PR.

---

## Verdict

**Approve.** The gate is the right shape, it recovers the failures that actually occur in real traffic, and I could not construct a duplicate-output or chain-corruption path against it. The one MINOR — a test for the unparseable-frame case so `emittedDownstreamData` stops being mutation-invisible — is worth adding, and I am happy for it to land either in this PR or as a follow-up. Nice piece of work on the rework.

PR #85, opened 2026-08-07
fix(patcher): accept the 2.1.224 minified Agent-tool model enum shape

zachspartofaday:
Claude Code **2.1.224** (shipped 2026-08-06) breaks `PATCH 1: Agent tool model enum`: the build minifies the Agent tool's model zod enum from `.enum(["sonnet","opus","haiku","fable"])` to a bare helper call on the property — `model:xr(["sonnet","opus","haiku","fable"])` — so the anchor misses. Because PATCH 1 is `required`, the whole patch run aborts:

```
■  Patch failed: clodex patch: required patch failed: PATCH 1: Agent tool model enum
●    FAIL PATCH 1: Agent tool model enum — anchor not found
●  clodex patch: 0 applied, 0 skipped, 1 failed
```

Every clodex install with model aliases configured hits this on the first launch after Claude Code auto-updates to ≥2.1.224.

## The fix

Accept both spellings in the transform and in the built-in proof pattern, keyed on the aliases array plus the trailing `.optional().describe(` so the match stays unique:

```
/((?:\.enum|model:[A-Za-z_$][\w$]*)\()(\["sonnet","opus","haiku"(?:,"[^"]+")*\])\)(\.optional\(\)\.describe\()/
```

The alternation cannot double-match on pre-2.1.224 builds: at `model:Ne.enum(`, the `model:<ident>(` alternative dies on the `.` and only `.enum(` fires. `PATCH_TRANSFORMS_VERSION` is bumped to 3 so existing installs re-patch instead of trusting the stale config hash (the deliberate-re-pin digest guard is updated accordingly).

## Verification

- **Uniqueness sweep over 11 real binaries**: ran the exact regex against every pristine backup on hand (2.1.215 → 2.1.224) plus two live patched binaries — exactly one match in every one (`.enum(` through 2.1.223, `model:xr(` at 2.1.224), including already-extended arrays, so re-patch/skip detection keeps working.
- **End to end**: `clodex patch` against a real 2.1.224 applies all 10 patches; a second run reports "already patched — nothing to do" (idempotency intact).
- **Proof-pattern parity**: the updated built-in proof pattern uniquely matches the patched 2.1.224 text, so local-patch-extension postcondition capture keeps working.
- **Mutant check**: reverting the alternation makes the new `model:xr` regression test fail (plus the digest-pin guard), so the fix can't regress silently.
- **Failure-mode note**: a hypothetical binary containing *both* shapes would fail loudly on the >1-match ambiguity check rather than patching the wrong site.

Suite: 84/84 patcher tests, full run and typecheck green.

### DIFF
--- a/src/built-in-patch-proofs.ts
+++ b/src/built-in-patch-proofs.ts
@@ -107,7 +107,7 @@ export function captureBuiltInPatchProofs(
 
   addPattern(
     'PATCH 1: Agent tool model enum',
-    /\.enum\(\["sonnet","opus","haiku"(?:,"[^"]+")*\]\)\.optional\(\)\.describe\(/,
+    /(?:\.enum|model:[A-Za-z_$][\w$]*)\(\["sonnet","opus","haiku"(?:,"[^"]+")*\]\)\.optional\(\)\.describe\(/,
   );
   addPattern(
     'PATCH 3: known-alias validator list',
--- a/src/patch-transforms.ts
+++ b/src/patch-transforms.ts
@@ -33,7 +33,7 @@ import { isReservedModelAlias } from './model-aliases.js';
  * and never receive the new transforms, silently. `tests/patcher.test.ts` pins a
  * hash of this file to force that decision to be made rather than forgotten.
  */
-export const PATCH_TRANSFORMS_VERSION = 2;
+export const PATCH_TRANSFORMS_VERSION = 3;
 
 export interface PatchScriptModelEntry {
   alias?: string;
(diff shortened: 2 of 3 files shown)


bman654 (approved):
Approving after the fact — this was already squash-merged as 2ec0916 to unblock 2.1.224 users, but the review record belongs on the PR. Thanks for the fast turnaround on this one, @zachspartofaday; the diagnosis and the write-up were both spot on.

I re-verified the claims independently rather than taking the description at face value:

**Uniqueness sweep, reproduced.** Ran both the old and new anchors against every pristine backup on hand — 13 binaries, 2.1.208 through 2.1.224, plus a live `native-binary.backup`. The new alternation matched **exactly once in every one**. The old regex matched once through 2.1.221 and **zero times at 2.1.224**, which pins the break precisely where you said it was. The `built-in-patch-proofs.ts` pattern tracks 1:1 with the transform, so proof capture stays in sync.

**The alternation is fail-safe, not just lucky.** The broadened `model:<ident>(` branch is wide enough to be worth worrying about, but `applyOnce` hard-fails a `required` site on `count > 1` (patch-transforms.ts:248). So the worst case on some hypothetical future binary carrying both shapes is a loud abort, never a patch landing on the wrong site. That's the right failure mode for a transform that rewrites a 270MB binary, and it's what makes the wider regex acceptable.

**End to end on a real 2.1.224.** Built the branch and ran `clodex patch --trace` against a live install: **10 applied, 0 skipped, 0 failed**. Second run correctly reported "already patched — nothing to do", so `noopIsSkip` idempotency survives. Inspected the resulting binary directly:

```
model:xr(["sonnet","opus","haiku","fable","luna","sol","terra"]).optional().describe(
```

Aliases appended inside the minified shape, constructor spelling preserved — exactly what the capture-group restructuring was for. Launch smoke test passed on both paths (`--model luna` and `--model haiku`).

**`PATCH_TRANSFORMS_VERSION` 2 → 3 is the part that makes this actually ship.** Without it, every existing install whose favorites hadn't changed would keep trusting its stale config hash and silently never receive the new transform — the fix would be in the package and absent from the binary. Good catch bumping it in the same commit and re-pinning the digest guard.

Typecheck clean, 1276/1276 tests green. One process note, no action needed from you: both workflow runs were parked in `action_required` (the contributor gate), so `gh pr checks` reported no checks at all rather than pending — I approved the runs and both passed (commitlint 11s, test 36s).

PR #87, opened 2026-08-07
feat(patcher): recognize configured identities in the autocompact threshold resolver

zachspartofaday:
Claude Code **2.1.223** introduced unknown-model window enforcement in the autocompact threshold resolver: a model it does not recognize gets `source:"unknown-model"` — the session is held to an assumed window with an unrecognized-model notice, and a conversation whose prefix already exceeds that threshold is **blocked with a synthetic "Prompt is too long"** instead of being sent. Configured clodex identities are exactly such models, despite PATCH 7 baking their true windows into the window resolver this same function reads.

Observed in production: concurrent subagents on a 272K-window alias whose early context was inflated past the threshold (a ~850KB bundled skill) died instantly at turn 2 with `AgentApiErrorTerminationError: Prompt is too long`, preceded by `autocompact: routing through reactive (thresholdSource=unknown-model)` and `level=blocked` — the requests never reached clodex.

## The fix — PATCH 10

Short-circuit the unknown-model branch: a configured identity returns `source:"model-default"` with its resolved window — the same treatment a recognized model gets (correct compaction semantics at its real window, no notice).

- The injected lookup **reuses PATCH 7's key set**, so the two baked tables cannot drift; keys are trimmed/lowercased at build matching the lookup.
- **Null-prototype object** for the lookup, so prototype names (the existing `constructor`-alias test case) cannot false-match.
- **In-place refresh** on re-patch preserves the build-specific identifier names and re-bakes the key set on a config change.
- **Best-effort by design** (`required: false`): on builds without the enforcement feature (≤2.1.222) or a future anchor drift, the patch records a non-fatal FAIL and everything else still applies — losing alias recognition degrades to the pre-PATCH-10 status quo instead of aborting every patch (the failure mode #85 just fixed for PATCH 1).

## Verification

- **Anchor sweep over 11 pristine binaries** (2.1.215 → 2.1.224): exactly one match on 2.1.223 and 2.1.224 (both alternatives of the enforcement code), zero on every earlier build — confirming both uniqueness and when the feature shipped. Also exactly one match on a live already-patched binary, so re-patching resolves correctly.
- **Behavioral test executes the patched resolver**: configured alias (any case/padding) and canonical id resolve `model-default` at their window; a genuinely unknown model still gets the enforcement branch; `constructor` stays unrecognized.
- **Refresh test**: re-patching with an added model rewrites the baked key set in place.
- **Mutant check**: widening the injected gate to always-true fails 9 tests (the unknown-model negative assertions and the deliberate-re-pin digest guard), so the patch cannot silently regress in either direction.
- Applied live: all patches green against a real 2.1.224, marker and key set confirmed in the patched binary; a session on the previously-affected workload shape runs with the aliases recognized.

`PATCH_TRANSFORMS_VERSION` bumped so existing installs re-patch. Based on v2.2.1 (includes #85), single commit, full suite + typecheck green.

### DIFF
--- a/src/built-in-patch-proofs.ts
+++ b/src/built-in-patch-proofs.ts
@@ -159,6 +159,10 @@ export function captureBuiltInPatchProofs(
     effortProofPattern('/*ccpatch:max-effort*/'),
   );
   addPattern(
+    'PATCH 10: autocompact alias recognition',
+    /\/\*ccpatch:acw\*\/if\(Object\.assign\(Object\.create\(null\),\{[^{}]*\}\)\[String\([\w$]+\|\|""\)\.trim\(\)\.toLowerCase\(\)\]!==void 0\)return\{window:[\w$]+,configured:[\w$]+,source:"model-default"\};/,
+  );
+  addPattern(
     'PATCH 9: default effort',
     /\/\*ccpatch:default-effort\*\/var _cce=Object\.assign\(Object\.create\(null\),\{[^{}]*\}\)\[String\([\w$]+\|\|""\)\.trim\(\)\.toLowerCase\(\)\];if\(_cce!==void 0\)return _cce;/,
   );
--- a/src/patch-transforms.ts
+++ b/src/patch-transforms.ts
@@ -33,7 +33,7 @@ import { isReservedModelAlias } from './model-aliases.js';
  * and never receive the new transforms, silently. `tests/patcher.test.ts` pins a
  * hash of this file to force that decision to be made rather than forgotten.
  */
-export const PATCH_TRANSFORMS_VERSION = 3;
+export const PATCH_TRANSFORMS_VERSION = 4;
 
 export interface PatchScriptModelEntry {
   alias?: string;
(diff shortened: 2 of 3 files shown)


bman654 (changes requested):
Thanks for the detailed writeup and the production evidence — the diagnosis of *what* goes wrong is accurate and well documented. I put this through a 4-lens adversarial review and then verified the key mechanics myself against the decompiled 2.1.224 bundle. The conclusion is that the patch doesn't move the behaviour you're targeting, and there's a supported one-line alternative that does. Details below, then a proposal.

## First, real credit: the anchor is clean

This is the part I was most worried about, because a patcher PR failed here badly once before (#78's anchor bound to a function ~407KB earlier than intended, spanned 989 function declarations, rewrote 132 references leaving 125 free identifiers — and still reported OK with one marker and would have published).

Yours is not that, and the evidence is unambiguous:

- **149-char matched span, provably inside `function Q9`** — a pure insertion (`commonPrefix + commonSuffix == originalLength`), **0 function declarations in the span, 0 references rewritten**.
- Exactly **1 match** on pristine 2.1.224 *and* on a live already-patched 2.1.224; **0 matches** on eleven older builds.
- The patched 23.7 MB bundle passes `node --check`.
- Fails closed against real-bundle decoys.

I also verified the fail-open claim myself by running your real `applyClodexPatches` against real extracted bundles: on **2.1.220** and **2.1.221** PATCH 10 records a non-fatal FAIL, the other 9 sites apply, nothing throws, and the patch publishes. On 2.1.224 all 10 apply. `required: false` does exactly what you said it does. Given #85, that mattered, and you got it right.

## Why the patch doesn't change the behaviour

There is exactly **one** consumer of the `source` string in the entire bundle:

```js
function KJe(e,t){ return Q9(e,t).source !== "auto" }
```

`KJe` is the on/off switch for the whole enforcement feature. Its 5 call sites gate the blocking branch in the query path (`ur = fx() && cCe() && !KJe(...)`, where `!ur` is what lets the "Prompt is too long" block run), the autocompact check (`if (cCe() && !KJe(t,r)) return !1`), and the status-line display. **Nothing anywhere tests for `"unknown-model"` specifically.**

That leaves three states:

| situation | `source` | `KJe` | enforcement |
|---|---|---|---|
| unpatched, unrecognized model | `"unknown-model"` | **true** | ON → blocked |
| `CLAUDE_CODE_DISABLE_UNKNOWN_MODEL_WINDOW_ENFORCEMENT=1` | `"auto"` | **false** | OFF → not blocked |
| **with PATCH 10** | `"model-default"` | **true** | **ON → still blocked** |

`"model-default"` and `"unknown-model"` are both `!== "auto"`, so `KJe` returns true either way. And the two returns are otherwise byte-identical — the enforcement branch returns `{window:o, configured:o}` and your injected line returns `{window:o, configured:o}` with the same `o`:

```js
if(fx()&&!ee.CLAUDE_CODE_DISABLE_UNKNOWN_MODEL_WINDOW_ENFORCEMENT&&…)
  return{window:o,configured:o,source:"unknown-model"};
return{window:o,configured:o,source:"auto"}
```

So the threshold is computed from the same window and the synthetic "Prompt is too long" fires at the identical token count. What genuinely changes is the startup notice and the `/config` label text.

That's also worth checking against your own verification note — "a session on the previously-affected workload shape runs with the aliases recognized" is consistent with the notice disappearing, which is easy to read as the block being fixed.

## Second problem: it disables the documented escape hatch

The snippet is prepended **before** the enforcement guard rather than substituted into it, so it runs unconditionally — including when the user has set `CLAUDE_CODE_DISABLE_UNKNOWN_MODEL_WINDOW_ENFORCEMENT=1`.

Today that env var makes a clodex identity fall through to `source:"auto"` → `KJe` false → enforcement genuinely off. With PATCH 10 applied it instead returns `model-default` → `KJe` **true** → enforcement back on. So the patch turns the block back on for exactly the users who explicitly turned it off. That's a one-line fix if the patch survives in some form, but see below.

## The proposal

**Have you tried the changelog-documented solution — setting `CLAUDE_CODE_DISABLE_UNKNOWN_MODEL_WINDOW_ENFORCEMENT=1` in your environment?**

> Changed auto-compact to keep sessions on unrecognized model IDs within the assumed context window instead of letting them grow past it; set `CLAUDE_CODE_DISABLE_UNKNOWN_MODEL_WINDOW_ENFORCEMENT=1` to restore the previous behavior.

It can be set in the shell environment, or in the `env` block of `.claude/settings.json`. Per the trace above it reaches `KJe` and actually flips the switch, which is the thing PATCH 10 cannot do. It's also narrowly scoped: only models that fall through to the `unknown-model` branch are affected — i.e. exactly clodex's aliases and canonical ids. Recognized Anthropic models return `settings`/`clientdata`/`experiment`/`model-default` from earlier branches and are untouched.

**If that resolves the blocking you hit, this PR could become a small documentation change** telling users to set that env var, and we'd avoid a new patch site entirely — no `PATCH_TRANSFORMS_VERSION` churn, and nothing to re-verify against every Claude Code release. Given how much of this repo's maintenance cost is anchors drifting, not adding one is worth a lot.

**And if the env var *doesn't* work for your case, that's the more interesting outcome — please say why.** A concrete reason it falls short (a path where `KJe` isn't the deciding gate, a workload where losing enforcement wholesale costs you something you need, an interaction with PATCH 7's baked windows, or simply that you tried it and the block still fired) would tell us something neither of us knows yet, and would justify a patch on much firmer ground than the current one. I'd rather revisit this with that information than have you rework the patch around my analysis if my analysis is the thing that's wrong.

## Two smaller notes, only relevant if a patch survives

- **The sweep claim mentions 2.1.223, and there's no 2.1.223 binary on this machine** (`~/.tweakcc` has .208–.221 and .224), so I could verify the 2.1.224-and-older legs but not that one. Entirely possible you have binaries I don't — flagging it only so the record is accurate about what was independently reproduced.
- **PATCH 10 is already claimed by #78** (child command network environment), which is still open. Whichever lands second needs to renumber. Unrelated to your logic.
- Minor, same category: the site is gated on `CONTEXT_BY_KEY`, so a configured identity whose window is unknown or equal to the 200k default is skipped and stays unrecognized, with no trace row explaining why.

Happy to be wrong about the inertness — if you're seeing the block actually lift with the patch applied, a `--debug` trace showing `level=` before and after would settle it quickly, and I'd want to understand what I'm missing in `KJe`.

zachspartofaday:
Conceding the core finding, with one correction to the record and the answer to your question — which changes what this PR should become.

**Verified your enforcement analysis against my own 2.1.224 bundle, and it holds.** `function KJe(e,t){return Q9(e,t).source!=="auto"}` — found verbatim, and the two returns share the same `o`, so the block fires at the identical threshold under `"model-default"`. I also confirm the escape-hatch regression by inspection of my own snippet: the injected line precedes the enforcement guard, so with the env var set a configured identity now returns `"model-default"` instead of falling through to `"auto"` — the patch re-enables enforcement for exactly the users who turned it off. That one is disqualifying as written regardless of anything else.

**One correction so the record stays accurate**: "nothing anywhere tests for `unknown-model` specifically" isn't quite right. Two more consumers do — `eOv` (the unrecognized-model startup notice: `if(n!=="unknown-model")return null`) and the auto-compact setup wizard, which both labels the source (`Xbe==="unknown-model"?"default for an unrecognized model":Xbe==="model-default"?"default for this model"`) and seeds its initial value differently for `unknown-model` vs `model-default`. Both are cosmetic, so your *behavioral* conclusion stands unchanged — but they're precisely what the patch actually moves, and it's telling that the bundle already renders `model-default` as a first-class label.

**Your question — did I try the env var, and does it fall short?** It works mechanically, and it isn't what I want. With PATCH 7 baking each identity's real window, enforcement computes the correct threshold and auto-compaction behaves correctly — enforcement at the right window is a feature, and I'd rather keep it than disable it wholesale. The blocking incident that motivated this patch turned out to have a different cause entirely: the affected agents' first tool call injected ~230k tokens of bundled-skill content into a 272k-window model, exceeding any threshold ("compaction cannot help" → blocked). The patch was authored under the earlier, wrong theory of that incident. What it genuinely delivers is notice suppression and an accurate `/config` label — cosmetics that don't justify a patch site, per your own maintenance-cost framing, and certainly not one that breaks a documented escape hatch.

**So: withdrawing this PR.** No docs change needed from my side either — the changelog line you quoted already covers the env var for users who do want enforcement off.

Two record notes: the 2.1.223 leg of the sweep was real — `claude-2.1.223-fcbe0b8d47570c50` is on my machine and PATCH 10's anchor matched exactly once there; happy to share the extracted bundle hash if you want it reproduced. And noted on the #78 numbering collision — moot now.

Thanks for the depth here. The 4-lens pass caught what my own behavioral test missed: asserting `source==="model-default"` at the right window proved the resolver changed, not that anything downstream cared.

PR #91, opened 2026-08-07
feat(oauth): omitted-reasoning alignment and abandoned-head canary coverage

zachspartofaday:
Closes two blind spots in the #89 canary where a genuinely forked strip rule stays silent. Both surfaced while adversarially reviewing the canary as ported to my fork; re-derived and re-verified on the merged form after #89 landed. One coherent change — both are coverage gaps in the same detector, in the same file.

## Blind spot 1: an omitted reasoning item hides the gap

`continuationMatch` explicitly supports Claude omitting a stored reasoning item. When a stored prefix is `[…, reasoning, function_call, …]` and the echo omits the reasoning, the exact-prefix divergence lands on a reasoning-vs-call pair — so `toolArgumentNormalizationGap`'s type gate rejects it, and a forked rule on the very next call produces neither the warning nor the diagnostic. Precisely the workload shape the omitted-reasoning mode exists for is the one the canary couldn't see.

Fix: when the divergence pair is reasoning-vs-`function_call`, align the canary's expected side to the first non-reasoning stored item. Everything else still uses the exact divergence pair; a message or other item diverging earlier remains an ordinary mismatch, as documented.

## Blind spot 2: a gap on an older head goes silent behind a newer head

Only `diagnosticEntry` — the most recently used idle candidate — passes through the warning-enabled path; the per-head diagnostic map deliberately uses `warnOnGap = false`. With multiple retained heads (the concurrent-subagent fan-out case the connection-pool section documents), a strip-rule regression on an older head plus an ordinary mismatch on the newest head means clodex starts a full-context chain and says nothing.

Fix: on a `history_mismatch_new_head` decision, every candidate was abandoned this turn, so each idle candidate passes through the warning-enabled path. This deliberately also lets the reasoning-gap warning fire for abandoned non-newest heads — on a no-match turn "the head clodex gave up on" is every candidate — and the existing per-signature dedup + hard cap keep stderr bounded regardless.

CLAUDE.md's "narrower than it looks" paragraph is updated to stay accurate on both points.

## Evidence

- **Reachability**: both paths execute in real configurations — omitted-reasoning is an explicitly supported continuation mode, and multi-head partitions are the documented fan-out behavior.
- **Behavioral tests, mutation-verified**: reverting the alignment fails exactly the omitted-reasoning test; reverting the loop fails exactly the multi-head test; nothing else moves in either direction.
- **Composed probes** (run during review, not committed): an omitted-reasoning gap on an older abandoned head warns end-to-end through the WS path, and the same gap present on two heads warns exactly once through the dedup.
- The abandoned-candidate pass runs only on mismatch turns and only over idle candidates, and `toolArgumentNormalizationGap`'s cheap identity gates run before any tool-schema walk (your thunk), so the added cost is bounded by retained heads on already-diagnostic turns.

`pnpm typecheck && pnpm test && pnpm build` green: 1,298 tests.

Happy to split the two if you'd rather take one without the other — they're independent; the tests are per-delta.


bman654 (changes requested):
Thanks for this — both blind spots are real, and I verified both fixes are correct **on the production path**, not just against the tests you shipped. The analysis in the description is right on both counts, and the CLAUDE.md rewrite is accurate.

My findings are entirely about the tests. That matters more than usual here, because the deliverable *is* a detector: if the tests don't fail when the detector is removed, we've shipped coverage we can't trust. We got bitten by exactly this on #89 — its first version shipped tests that stayed green when the whole feature was deleted — so this is a check we now run on every detector PR, including our own.

## First, what I confirmed is correct (please don't re-derive)

- Reverting the alignment fails **exactly** the omitted-reasoning test and nothing else.
- The alignment is region-agnostic (`prefix = requestInput ++ expectedAssistant`), so it works wherever the pair lands.
- It changes **only** the tool-gap input — `expectedKind`/`expectedHash` still report the true divergence pair, as your description says.
- `[reasoning, message, call]` correctly yields no gap (the scan stops at the non-reasoning item and the type gate rejects it).
- No false-positive path: the `call_id` + `name` identity gate is unchanged.
- The loop cannot warn on a head that wasn't abandoned — it's reached only when `!selected` and no candidate is `inFlight`.
- No new throw risk: `reasoningNormalizationGap` already ran per-candidate for the `heads` map.
- stderr stays bounded via per-signature dedup + the cap of 3.

## MAJOR (tests): the multi-head test passes with the feature deleted

Deleting the abandoned-head loop leaves **all 91 tests in `tests/responses-websocket.test.ts` green** under a normal `pnpm test`.

It does fail under `vitest -t "warns for a gap on an older head"`. The two runs disagree, and the full-file run is the one CI does.

Root cause, confirmed with a probe on `idleCandidates`: the test injects no clock, so both heads can be created within the same `Date.now()` millisecond and tie on `lastUsedAt`. The tie-break is

```ts
[...idleCandidates].sort((left, right) => right.lastUsedAt - left.lastUsedAt)[0]
```

which is a **stable** sort — so on a tie the **older** head becomes `diagnosticEntry` and warns through the pre-existing `continuationMismatchSummary` path, with or without your loop. I saw ties on 3 of 6 probe lines in a single run.

Fix — `now` is already a documented test override on `ResponsesWebSocketFetchOptions`:

```ts
let clock = 1_000_000;
const wsFetch = createResponsesWebSocketFetch(WS_URL, undefined, {
  accountId: 'acct-tool-gap-multi', now: () => (clock += 1),
});
```

With that, the test fails 3/3 full-file runs with the loop removed, and passes with it. Worth asserting the newer head really is the diagnostic entry too, so a future tie-break change can't silently re-hide it.

## MAJOR (tests): both new tests stage in `requestInput`, where the bug cannot occur

Both tests put the reasoning item and the `function_call` in the **first request's input**, so they land in `entry.requestInput`. Two consequences:

1. `sanitizedCallArguments()` is applied **only** in `expectedAssistantItems()`. Items in `requestInput` are stored verbatim as the client sent them, so a forked strip rule can never physically produce a divergence there.
2. More specifically for the first test: `continuationMatch`'s omitted-reasoning mode filters reasoning **only** out of `entry.expectedAssistant`, and bails immediately when there is none there —

```ts
const echoedAssistant = entry.expectedAssistant.filter(item => conversationItemKind(item) !== 'reasoning');
if (echoedAssistant.length === entry.expectedAssistant.length) return undefined;
```

With the reasoning in `requestInput`, `expectedAssistant` holds only the assistant text, so **the omitted-reasoning mode the test is named after never engages**. The test exercises the alignment arithmetic through a client that simply dropped an item from its own previous input.

This is the same staging defect #89 shipped in its first version, which is why I'd rather fix it than wave it through.

The good news: I re-staged both scenarios the way production stages them — emitted by upstream via `response.output_item.done` so they land in `expectedAssistant` — and **both fixes hold**:

| harness | feature present | feature reverted |
|---|---|---|
| realistic omitted-reasoning fork | `warned=true`, `toolArgumentNormalizationGap {tool: 'Grep', equalAfterStrip: true}` | `warned=false`, no gap recorded |
| realistic multi-head, older abandoned head | `warned=true` | `warned=false` |

So the code is right; only the staging is wrong. Your existing `runToolArgumentMismatch` helper already does upstream staging correctly — moving both tests onto it (with a reasoning item emitted at `output_index 0` for the first) is the whole fix. Happy to hand over my two harnesses if useful.

## NITs

- `candidate.inFlight` in the new loop is dead — the preceding `else if (candidates.some(entry => entry.inFlight))` already guarantees no candidate is in flight by the time you get there. Harmless as defensive code, but it reads as if the case is live.
- `continuationMismatchDetails` now runs twice per idle candidate on mismatch turns (your loop, then the `heads` map). `decision` is known before `emitDiagnostic`, so the `heads` map could take `warnOnGap` directly and the separate loop could go away. I'm **not** asking for this — it touches a file CLAUDE.md says not to restructure, and the cost is bounded — just noting it in case you prefer that shape.

Nothing here is a design objection: split or combined, both deltas are ones we want. Once the two tests are re-staged and the clock is injected, this is good to land.

zachspartofaday:
Thanks — both findings were right, and the multi-head one was worse than "flaky staging": that test passed a full-file run with the feature deleted, which is exactly the coverage-you-can't-trust case. Pushed as `0e0ce74`.

## Both tests re-staged

Both now go through `runToolArgumentMismatch`, so the stored items arrive from upstream via `response.output_item.done` and land in `expectedAssistant`. For the omitted-reasoning test that also means the mode it's named after actually engages now — with the reasoning in `requestInput`, `expectedAssistant` held only the assistant text and `continuationMatch` bailed at the length check, so the old test never entered the path at all.

The multi-head test injects `now: () => (clock += 1)` and additionally asserts the newer head is the diagnostic entry, so a future tie-break change can't silently re-hide it.

## Both nits, and one thing I did beyond what you asked

Nit A is gone with the loop. I also did nit B, which you explicitly weren't asking for — flagging that clearly since it touches the file `CLAUDE.md` says not to restructure, so please push back if you'd rather have the simpler shape.

Folding the loop into the `heads` map removes the double `continuationMismatchDetails` run, but it makes a user-visible stderr warning a side effect of building a diagnostic payload — and that payload is an argument to `emitDiagnostic`, which returns early without a listener. The warning would survive only because arguments are evaluated before the call. Since `onDiagnostic` is wired by `clodex server --ws-diagnostics` alone, anyone later making `heads` lazy would kill the detector for every ordinary `clodex claude` user, silently.

So instead: each candidate's mismatch is computed once into a map *before* `emitDiagnostic`, warnings happen there, and the `heads` map reads the cached value. Same double-evaluation fix you were pointing at, without putting warnings on the diagnostics path. `continuationMismatchSummary` takes the diagnostic entry's precomputed details so it doesn't recompute either.

The warn set is unchanged — diagnostic entry through the summary, every other abandoned candidate through the loop. I kept the `entry !== diagnosticEntry` exclusion deliberately: the trace line in `warnToolArgumentNormalizationGap` fires before the dedup check, so warning it twice would double-print under `--trace`. There's a test pinning that at one line. The pre-existing tests also pass unmodified against the refactor commit alone, which is the behaviour-preservation check.

## One more blind spot, found on our side

Reviewing our own diff adversarially turned up a gap you didn't flag, in the code this PR adds.

The abandoned-head loop warns through `continuationMismatchDetails`, which fires **both** detectors — but only the tool half was covered. Skipping every older head carrying reasoning left all 91 tests green. Before this branch only `diagnosticEntry` ever warned, so reasoning-gap-on-an-older-head is a path this PR creates, and it was shipping untested.

Added the reasoning analogue of the multi-head test. It asserts on identical `encrypted_content` and explicitly rejects the `filler-strip rule` text, so it can't pass via the tool detector.

## Mutation results (each a full-file run)

| mutation | result |
|---|---|
| reasoning detector skipped for older heads | fails only the new reasoning test |
| abandoned-head warning disabled | fails **both** older-head tests |
| warn loop gated on a diagnostics listener | fails the tool multi-head test |
| omitted-reasoning alignment reverted | fails only the omitted-reasoning test |

The multi-head clock fix holds at 3/3 full-file runs. Note the second row is intentionally not a one-test mutation any more: disabling the loop kills both detector families, so both older-head tests fail. 92/92 in the file, 1299/1299 across the suite, `typecheck` and `build` clean.

## Not fixed, flagging instead

The warning hard cap is unpinned — disabling the cap guard, or changing 3 to 4, leaves the suite green; only same-signature dedup is covered. It's pre-existing and untouched by this PR, so I left it alone rather than widening the diff. Happy to do it here or in a follow-up, whichever you prefer.

Also noting: `continuationMismatchSummary`'s new `precomputedDetails ?? continuationMismatchDetails(..., true)` fallback would re-warn if a caller ever passed `undefined`. It's unreachable today — one module-private caller, always passing the value — so I left it rather than adding a guard, but it's a small sharp edge if that function ever gets a second caller.

Didn't take you up on the harnesses in the end, but the offer was useful for confirming the target shape — thanks.

bman654 (approved):
Re-reviewed at `0e0ce74`. **Both round-1 findings are resolved**, and I re-reviewed `f36e7f8` as a redesign rather than as a diff. Nothing outstanding beyond three optional nits. Thanks for taking the harder route on the second commit — more on that below.

---

## Resolved: tests are now staged where the fork can occur

The stored `function_call` (and the stored reasoning item) now reach the head through upstream `response.output_item.done` → `expectedAssistantItems()` → `sanitizedCallArguments()` → `entry.expectedAssistant`, which is the only region a forked strip rule can diverge in. I did not take that from the diff — I instrumented `continuationMatch` to print what the head actually held and whether the omitted-reasoning branch was entered:

```
// temporary probe inserted at the omitted-reasoning filter in continuationMatch
PROBE expectedAssistant=["reasoning","function_call"] echoed=1 equal=false
PROBE expectedAssistant=["reasoning","function_call"] echoed=1 equal=false
      Tests  1 passed | 91 skipped (92)
```

So `entry.expectedAssistant` really is `[reasoning, function_call]` (both items arrived from upstream, not from the request body), and `echoedAssistant.length (1) !== entry.expectedAssistant.length (2)`, which is the guard that previously short-circuited the mode. The omitted-reasoning path is genuinely entered on both runs of the test (the diagnostics run and the no-listener run). The divergence pair is then `reasoning` vs `function_call`, which is exactly the shape the alignment block exists for.

The multi-head tests stage the same way — the older head's call is emitted via `response.output_item.done` before the replay carries the filler-bearing echo.

## Resolved: the multi-head test is mutation-sound in the full-file run

The two soundness guards I hoped for are both there: the injected stepping clock (`now: () => (clock += 1)`) makes a `lastUsedAt` tie impossible, and `expect(decision.promptChanges).toEqual(['instructions'])` independently proves the *newer* head became `diagnosticEntry` (the third request's instructions match the older head, so a change is only reported if the diagnostic entry is the newer one). `heads[0].idleMs > heads[1].idleMs` pins the ordering as well.

Mutations were run the way CI runs the suite — the whole file, not `-t` isolation — three times each.

**Mutation A — reasoning-alignment block in `continuationMismatchDetails` replaced with `const gapExpected = expected;`:**

```
$ CLODEX_HOME=$(mktemp -d) pnpm vitest run tests/responses-websocket.test.ts   # x3
   × ... > still catches a forked rule when Claude omits the stored reasoning item
      Tests  1 failed | 91 passed (92)      (identical on runs 1, 2, 3)
```

**Mutation B — the abandoned-head loop deleted:**

```
$ CLODEX_HOME=$(mktemp -d) pnpm vitest run tests/responses-websocket.test.ts   # x3
   × ... > warns for a gap on an older head even when a newer head mismatches ordinarily
   × ... > warns for a reasoning gap on an older head when a newer head mismatches ordinarily
      Tests  2 failed | 90 passed (92)      (identical on runs 1, 2, 3)
```

Each mutation fails exactly the tests that name it, deterministically, in the full-file run. That is the check that was missing last round.

**Mutation C — dropping the new `precomputedDetails` argument so `continuationMismatchSummary` re-evaluates:**

```
   × ... > still catches a forked rule when Claude omits the stored reasoning item
      Tests  1 failed | 91 passed (92)
```

so the `expect(trace.filter(... 'tool argument normalization gap: Grep:filler')).toHaveLength(1)` assertion is load-bearing and the caching itself is covered.

## Independent reproduction

I also re-ran the two standalone reproductions I built during the first round, which drive both scenarios through upstream staging and simply report rather than assert:

| state | HARNESS (omitted reasoning) | HARNESS2 (older abandoned head) |
|---|---|---|
| `0e0ce74` as-is | `warned=true` | `warned=true` |
| alignment block reverted | `warned=false` | `warned=true` |
| abandoned-head loop deleted | `warned=true` | `warned=false` |

with the head-decision record on the first one showing the mechanism end to end:

```
HARNESS decision=history_mismatch_new_head
HARNESS mismatch={"fullItems":3,"expectedPrefixItems":3,"firstMismatch":1,
  "expectedKind":"reasoning","actualKind":"function_call",
  "expectedHash":"d741c87894e28f35","actualHash":"35e5142bbfba013c",
  "toolArgumentNormalizationGap":{"tool":"Grep","equalAfterStrip":true}}
```

Each feature is independently necessary and neither masks the other.

---

## Fresh review of `f36e7f8` (the refactor)

You changed the approach rather than only the test, so I reviewed it as new work.

**Not taking the coupling I suggested was the right call.** My round-1 nit proposed collapsing the double evaluation by passing `warnOnGap` into the `heads` map. Your version keeps the warning on its own explicit path and caches the result for the payload instead, which means the warning cannot become a side effect of building a diagnostic. The tests that run with no `onDiagnostic` listener at all and still assert the stderr warning are a good regression guard for that, even though `emitDiagnostic` currently early-returns after the payload is already constructed.

Checks I ran against the redesign:

- **Can it warn about a head that was not abandoned?** No. The block is the `else if (diagnosticEntry)` arm, reachable only when `matches.length === 0` (so no candidate was continued — `selectedDelta` is an array and always truthy when `matches[0]` exists) and the preceding `else if (candidates.some(entry => entry.inFlight))` did not fire. Every candidate in that arm is idle and abandoned. That also makes dropping the old `candidate.inFlight` skip a no-op rather than a widening — it was already unreachable.
- **Can it double-warn?** No, and it now evaluates strictly less. Per mismatch turn the old code did `1` (summary) + `N-1` (loop) + `N` (heads map) = `2N` evaluations; the new code does `1 + (N-1) = N`. `warnedToolArgumentGaps` / `warnedReasoningGaps` dedup by signature and the `MAX_TOOL_ARGUMENT_GAP_WARNINGS` / `MAX_REASONING_GAP_WARNINGS` caps are untouched.
- **Does the `ws_head_decision` payload change?** No. `warnOnGap` appears at exactly four lines in the file (`:633` declaration, `:648`, `:683`, `:684`) and every one of them gates a side effect — the two warners and one trace line. It never participates in the returned object, so a `mismatch` record computed with `true` is byte-identical to one computed with `false`. No key was added, renamed or removed.
- **Does moving the computation earlier change what it sees?** The `heads[].mismatch` for non-diagnostic candidates is now computed *before* `evictOldestIdleGeneration('nursery', …)` instead of after. I checked the eviction path: `deleteEntry` sets `inFlight = false` and `current = undefined` and unregisters/closes, and touches none of `requestInput` / `expectedAssistant` / `headRequiredToolProps`, which are the only fields `continuationMismatchDetails` reads. Same values either way.
- **New throw on a request path?** None. The additions are a `Map` construction plus `set`/`get` on object identities that are stable for the turn, and one extra optional parameter read. `continuationMismatchDetails` is called with the same arguments it always was, and its existing try/catch around the predicate is unchanged.
- **Scope.** The change is confined to one `else if` arm plus a single optional parameter on `continuationMismatchSummary`; nothing was moved, renamed, or re-ordered elsewhere in the file. That is within what the "do not restructure" rule is asking for.

---

## Nits (all optional)

1. **`CLAUDE.md` overstates the coverage slightly.** "On a turn where no head matches, every abandoned idle candidate passes through the warning path" — when *any* candidate is in flight, the request takes the `parallel_isolated` arm and no candidate warns, even though no head matched there either. "On a turn where no head matches and none is in flight, …" would be exact. Everything else in the rewritten paragraph checks out against the code, including the alignment sentence.
2. **`--trace` line ordering changed for non-diagnostic heads.** Their `tool argument normalization gap: …` lines used to follow the `history mismatch starting an additional chain (…)` summary; they now precede it, because the loop moved above the `debug(...)` call. Observed:
   ```
   TRACE ws: tool argument normalization gap: Grep:filler
   TRACE ws: history mismatch starting an additional chain; retained 2 existing head(s) (full_items=4 …)
   ```
   Purely cosmetic, but a log reader scanning for "which turn was this gap on" now has to look one line down instead of one line up.
3. **`captureDiagnostics?: boolean` only has meaning as `false`** (`options.captureDiagnostics === false`); `true` and omitted are the same. `omitDiagnostics?: true` would say what it does.

---

## Verification

```
$ CLODEX_HOME=$(mktemp -d) pnpm typecheck     # clean
$ CLODEX_HOME=$(mktemp -d) pnpm build         # ESM ⚡️ Build success in 1057ms
$ CLODEX_HOME=$(mktemp -d) pnpm test
   Test Files  83 passed (83)
        Tests  1299 passed (1299)
$ CLODEX_HOME=$(mktemp -d) pnpm vitest run tests/responses-websocket.test.ts
        Tests  92 passed (92)
```

Two full-suite runs on my machine flaked in unrelated timing-sensitive files (`tests/keyring-credentials.test.ts` once, `tests/credential-helper.test.ts` once — a different file each time, both green in isolation and both green on a clean re-run). Nothing in `tests/responses-websocket.test.ts` ever flaked across roughly a dozen runs, so I am treating those as local machine load, not something this PR introduced.

**Recommendation: approve.** Both round-1 findings are closed with evidence, and the refactor is a net improvement on the path it touches.

PR #92, opened 2026-08-08
fix(proxy): honor outbound proxy for raw passthrough

camjac251:
## Summary

- route the native raw passthrough through the existing `HTTP(S)_PROXY` and `NO_PROXY` resolver
- create one CONNECT agent when the local bridge starts and reuse it across requests
- preserve the existing native header, body, and streaming response path without adding configuration or persisted credentials

## Test plan

- [x] `corepack pnpm typecheck`
- [x] `corepack pnpm test` (83 files, 1,298 tests)
- [x] `corepack pnpm build`
- [x] authenticated CONNECT integration test covering destination, credentials, request bytes, and streaming response delivery
- [x] `git diff --check`
- [x] `gitleaks git --staged`


bman654 (changes requested):
Thanks — this is exactly the change we asked for in the #79 discussion, and you implemented it the way we suggested: reuse `outboundProxyUrlForTarget` rather than write a second matcher, and hang the tunnel off the `agent` option of the existing `https.request` instead of swapping in a different HTTP client. That decision is worth calling out, because it is what makes the passthrough integrity hold (details in "Verified clean" below).

The change is small and the tests are real. I have **one blocking issue** and **two smaller ones** before this can land.

Everything below was run in a worktree at `c190eb2`, with `CLODEX_HOME` set to a fresh temp dir and the proxy env vars scrubbed unless a check specifically needed them.

---

## Build and test status

```
$ pnpm typecheck
> tsc --noEmit          # clean

$ pnpm test
 Test Files  83 passed (83)
      Tests  1298 passed (1298)
   Duration  38.95s

$ pnpm build
ESM dist/cli.js  2.79 MB
ESM ⚡️ Build success in 198ms
```

83 files / 1298 tests matches your number exactly. (I saw one-off failures in `tests/credential-helper.test.ts` on an earlier run, but that was my own fault — I had a socket-heavy probe running against the same machine at the time. Clean runs are green.)

---

## BLOCKER — the passthrough can now CONNECT to clodex's own listener and loop

`shouldInterceptConnect()` intercepts exactly `api.anthropic.com:443`, and `forwardRawAnthropicRequest` now CONNECTs to `api.anthropic.com:443` through whatever `HTTPS_PROXY` says. If `HTTPS_PROXY` happens to name clodex's own proxy port, the passthrough tunnels into the listener it is serving, gets intercepted, forwards again, and recurses.

This is not a contrived configuration — it is the configuration this repo prints and tells users to export. `src/http-proxy/index.ts:143-158`:

```
  HTTPS_PROXY=http://127.0.0.1:${handle.port}
  HTTP_PROXY=http://127.0.0.1:${handle.port}
  NODE_EXTRA_CA_CERTS=${handle.caCertPath}
```

plus an adjusted `NO_PROXY` that deliberately *removes* the Anthropic bypass (README:156, `docs/background-agents.md:32`). So: start `clodex server --proxy`, export the printed block into the shell as instructed, then later Ctrl-C and restart the server **from that same terminal** — which is the normal way anyone iterates — and the new server's own `process.env['HTTPS_PROXY']` is `http://127.0.0.1:17645`, its own fixed default port.

Reproduction (single client request; `port` pinned so the env can name the listener in advance):

```ts
const port = await freePort();
process.env['HTTPS_PROXY'] = `http://127.0.0.1:${port}`;   // == our own listener
const proxy = await startHttpProxy({ routes: [], port });  // default origin api.anthropic.com
// ... one ordinary POST /v1/messages through the MITM, then sample accepted sockets on `port`
```

At `c190eb2`:

```
accepted-socket samples on own listener: [ 23, 40, 64, 86, 99, 130, 142, 177 ]
```

177 sockets, still climbing, 2 seconds after **one** request. With `src/http-proxy/server.ts` and `src/outbound-proxy.ts` reverted to `827215a` and nothing else changed, the same script gives a flat count and passes — so this is introduced here, not pre-existing.

Two supporting facts I checked before calling it, because both are the kind of thing that could make it a non-issue:

1. **Does TLS verification stop it?** No — not under the documented setup. The passthrough runs with `rejectUnauthorized: true`, but the CONNECT lands on clodex's own MITM, which serves a cert signed by the CA the same printed block tells the user to export. Against the real `~/.clodex/http-proxy` certs:

   ```
   NODE_EXTRA_CA_CERTS = /Users/…/.clodex/http-proxy/clodex-ca.pem
   RESULT: handshake to clodex MITM cert SUCCEEDED with rejectUnauthorized=true, authorized = true
   ```

   So the handshake completes and the loop keeps going.

2. **Does `NO_PROXY` save the user?** No. Loopback is not implicitly bypassed, and the printed `NO_PROXY` strips the Anthropic entry on purpose:

   ```
   NOPROXY host=127.0.0.1  NO_PROXY="localhost"  bypass=false
   URLRES loopback target, HTTPS_PROXY set -> http://proxy.corp:8080
   ```

Note this is specific to the fixed-port `clodex server --proxy` case. `clodex claude --proxy` binds an ephemeral port, so its env can't accidentally name it — but there it produces a quieter version of the same surprise: the session's Anthropic passthrough now silently transits *another* clodex MITM, and if that other server is down the passthrough fails with `ECONNREFUSED` where it used to work.

**Suggested fix (small).** The listener's bound address is known once `listenTcpServer` resolves, and the agent isn't needed until the first request — so build it after listen and drop it if it names us:

```ts
const proxyUrl = outboundProxyUrlForTarget(anthropicOrigin.href);
// after listen, with host/port known:
if (proxyUrl && isOwnListener(proxyUrl, boundHost, boundPort)) {
  console.error('clodex: HTTPS_PROXY points at this proxy; sending Anthropic passthrough direct');
} else {
  anthropicAgent = await outboundHttpProxyAgent(anthropicOrigin.href);
}
```

`src/cli.ts:997-1008` already computes almost exactly this value (`inheritedProxyPort`, loopback host + port) for the lifecycle log, so the parsing is already written down once.

Whatever shape you pick, please also update the "Self-loop guard" sentence in the CLAUDE.md paragraph you touched. It currently reasons only from "clodex never *sets* proxy vars in its own `process.env`" — true, but no longer sufficient once a transport follows those vars to a host this process intercepts.

---

## MAJOR — a malformed `HTTPS_PROXY` now stops `claude` from launching

`outboundHttpProxyAgent` calls `new HttpsProxyAgent(proxyUrl)` with no guard, and `startHttpProxy` awaits it. A scheme-less proxy URL — a very common way to get this wrong — throws:

```
$ node -e "new (await import('https-proxy-agent')).HttpsProxyAgent('127.0.0.1:8080')"
THROW "127.0.0.1:8080" -> TypeError | msg: Invalid URL
OK    "localhost:3128"  -> localhost:3128      (parses, but as protocol "localhost:")
```

End to end, with `HTTPS_PROXY=127.0.0.1:8080`:

```
HEAD c190eb2 : startHttpProxy result: THREW TypeError: Invalid URL
BASE 827215a : startHttpProxy result: started fine
```

Both callers turn that into a hard stop — `src/cli.ts:989-994` prints `Failed to start proxy: Invalid URL` and returns 1, `src/http-proxy/index.ts:207-218` the same for `clodex server`. So a typo in an unrelated env var now means `clodex claude` never launches Claude Code at all, with an error message that doesn't say which variable is at fault.

This is a regression rather than a pre-existing sharp edge: on `main` the only consumer of that env at startup is `installOutboundProxyDispatcher`, and undici tolerates it —

```
$ HTTPS_PROXY=127.0.0.1:8080 node -e "new (await import('undici')).EnvHttpProxyAgent()"
EnvHttpProxyAgent constructed OK with malformed HTTPS_PROXY
```

— and that call site is wrapped in a catch that explicitly warns and falls back to direct connections ("it must never break the CLI", `src/outbound-proxy.ts:77-97`).

**Suggested fix:** give `outboundHttpProxyAgent` the same contract as its neighbour — try/catch around the constructor, one-line stderr warning, return `undefined`. That also hardens the WebSocket path (`src/oauth/responses-websocket.ts:1838`), which has the same exposure today.

One thing I specifically checked and it is **fine**: no credential leak on this path. Both callers print `err.message` only (`"Invalid URL"`), never the `err.input` property that carries the raw URL. Please keep it that way in the fix — log the error message, not the error object, and never the resolved proxy URL.

---

## MINOR — the new agent has `keepAlive: false`, so every passthrough request re-tunnels

The PR description says the agent is created once and reused across requests. The *object* is reused; the sockets are not:

```
agent.keepAlive = false
request 1 done, CONNECTs so far = 1
request 2 done, CONNECTs so far = 2
request 3 done, CONNECTs so far = 3
TOTAL CONNECT tunnels for 3 sequential requests = 3
```

For comparison, the no-agent path this replaces uses `https.globalAgent`, which does pool:

```
node v24.14.1 | https.globalAgent.keepAlive = true
```

So a proxied user pays a fresh CONNECT plus a full TLS handshake on every Sonnet/Haiku passthrough. `new HttpsProxyAgent(proxyUrl, { keepAlive: true })` gets the behaviour the description already claims. (Non-proxy users are unaffected — `agent` stays `undefined` and `https.globalAgent` is used as before.)

---

## MINOR — an existing test in the same file became environment-dependent

Your new test correctly scrubs `process.env` before it runs. But `logs an Anthropic connection failure as an upstream response failure` now picks up an ambient `HTTPS_PROXY` and stops seeing the `ECONNREFUSED`/`ECONNRESET` it asserts, because the "unreachable origin" is reached through a proxy instead:

```
# in a shell where HTTPS_PROXY is set
BASE 827215a (src+tests): Tests  19 passed (19)
HEAD c190eb2            : × logs an Anthropic connection failure as an upstream response failure
  AssertionError: expected [...] to deep equally contain ObjectContaining{ errorType: /^ECONN(?:REFUSED|RESET)$/, ... }
```

CI is unaffected (no proxy vars there), and with a scrubbed env the whole file passes. But anyone developing while their own shell is bridged through clodex will now hit this, so it is worth folding that one test into the same env-scrubbing helper.

---

## Verified clean — please don't re-litigate these

- **`NO_PROXY` is the existing resolver, not a copy.** `outboundHttpProxyAgent` calls `outboundProxyUrlForTarget`, which calls `noProxyBypasses`; there is no second matcher to drift. I probed the rules and they behave as documented: exact host, bare-domain suffix, `.foo`, `*.foo`, `*`, case-insensitive, whitespace-tolerant list, `https://` prefix stripped, and no false positive on `evilanthropic.com` vs `anthropic.com`. Port-qualified entries match host-only and CIDR is unsupported, but both are unchanged pre-existing behaviour of that helper, not something this PR introduces.
- **Nothing in the diff mutates `process.env`.** `outboundHttpProxyAgent(target, env = process.env)` reads only. `tests/env.test.ts` (`does NOT mutate process.env`, `child-env builders never mutate clodex process.env`) and `tests/outbound-proxy.test.ts` pass together: 38/38. Nothing here reaches child env either.
- **No credential exposure.** The proxy URL (which may carry `user:password`) is passed to `HttpsProxyAgent` and nowhere else — not logged, not persisted, not in child env, not in the printed export block (which prints clodex's own loopback URL). Upstream failures surface as `connect ECONNREFUSED 127.0.0.1:1` with props `["errno","code","syscall","address","port"]` — no userinfo. A proxy returning 407 is relayed as an ordinary response, not an error string.
- **Passthrough integrity holds, including the shapes that broke the undici experiment.** I ran the full matrix through the real MITM, direct vs. CONNECT tunnel:

  ```
  RESULT direct cl              {"connects":0,"seenEncoding":"content-length:23","seenBodyLen":23,"status":"HTTP/1.1 200 OK","gotStop":true,"closed":true}
  RESULT direct chunked         {"connects":0,"seenEncoding":"chunked","seenBodyLen":23,"status":"HTTP/1.1 200 OK","gotStop":true,"closed":true}
  RESULT direct cl-expect       {"connects":0,"seenEncoding":"content-length:23","seenBodyLen":23,"status":"HTTP/1.1 100 Continue","gotStop":true,"closed":true}
  RESULT direct chunked-expect  {"connects":0,"seenEncoding":"chunked","seenBodyLen":23,"status":"HTTP/1.1 100 Continue","gotStop":true,"closed":true}
  RESULT tunnel cl              {"connects":1,...,"gotStop":true,"closed":true}
  RESULT tunnel chunked         {"connects":1,...,"gotStop":true,"closed":true}
  RESULT tunnel cl-expect       {"connects":1,...,"status":"HTTP/1.1 100 Continue","gotStop":true,"closed":true}
  RESULT tunnel chunked-expect  {"connects":1,...,"status":"HTTP/1.1 100 Continue","gotStop":true,"closed":true}
  Tests  8 passed (8)
  ```

  Identical on both sides: chunked framing preserved, `Expect: 100-continue` forwarded and the interim `100 Continue` relayed, exact body bytes, SSE streamed through, socket closed cleanly. (My first attempt at this showed the tunnel case hanging — that was my toy tunnel failing to propagate the origin's close to the client, not your code. Once the tunnel destroys the client socket on target close, like a real proxy does, all eight are identical.) Keeping `https.request` and only swapping the socket source is what buys this.
- **The response-model echo invariant is untouched.** `rawBody` is still forwarded byte-for-byte and `copyResponse` still copies the upstream response through; the diff adds only an `agent` parameter.
- **Cleanup on the failure paths is right.** The agent is destroyed if `startProxyCatalog` throws, if `listenTcpServer` throws, and in `close()`.

## Test quality

I ran the feature-deletion mutations the way CI runs them — whole file, not `-t` isolation:

| mutation | result |
| --- | --- |
| drop `agent,` from the `https.request` options | `tests/http-proxy-server.test.ts` — 1 failed / 19 passed |
| never create the agent (`const anthropicAgent = undefined`) | fails exactly `streams raw passthrough through the configured CONNECT proxy` |
| `outboundHttpProxyAgent` reads `HTTPS_PROXY` directly, skipping the resolver | `tests/outbound-proxy.test.ts` — fails `builds an agent only when the target is not bypassed` |
| remove `anthropicAgent?.destroy()` from `close()` | **survives** — 20/20 still pass |

The first three are what we want to see: deleting the feature fails something, and the `NO_PROXY` coupling is genuinely exercised rather than asserted against a hand-written value. The last one is only a gap in coverage, not a defect — the teardown is correct, just untested.

---

## Summary

- **BLOCKER:** self-CONNECT loop when `HTTPS_PROXY` names clodex's own listener — reachable straight from the export block clodex prints.
- **MAJOR:** malformed `HTTPS_PROXY` now aborts the launch instead of degrading to direct.
- **MINOR:** `keepAlive: false` means a new tunnel per request; the description implies otherwise.
- **MINOR:** one existing test in the touched file became proxy-env-dependent.

The core approach is right and the tests are honest — these are all guard-rails around it rather than rework. Happy to re-review as soon as the first two are addressed.

camjac251:
Updated at `9e348fb` and rebased onto current `main` (`2.3.0`).

The raw passthrough agent is now created after the bridge binds. If the resolved proxy URL names that listener, passthrough warns and connects directly instead of recursively connecting through itself. Agent construction validates the proxy URL, fails open with a credential-safe warning, and enables keep-alive.

Coverage now includes loopback and wildcard listener matching, the self-listener startup path, malformed credentialed and scheme-less proxy values, keep-alive, and an ambient-proxy-independent connection-refusal lifecycle test.

Verification:
- `pnpm typecheck`
- 84 test files / 1,323 tests
- `pnpm build`
- diff check and staged leak scan
- required checks: `test` and `commitlint` both successful

bman654 (changes requested):
Both the blocker and the major from last round are genuinely fixed. One test-hygiene regression is worth fixing before this lands, and it happens to hit the exact users this PR is for.

> **What we are actually requesting: the fix under "Please fix before merge".** That is the only thing blocking a merge.
>
> Everything under "Minors" is optional and listed only in case you would rather sweep some of it while you are in here — minor 1 in particular is cheap and closely related. None of it needs to happen in this PR, and we will not hold the merge on any of it.

## Self-CONNECT loop — fixed

Starting `startHttpProxy` on a fixed port with `HTTPS_PROXY` naming that same listener now warns

```
clodex: HTTP(S)_PROXY points at this proxy; sending Anthropic passthrough direct
```

and creates no agent. A genuine external proxy is not mistaken for the listener. Matrix over `proxyUrlTargetsListener` — exact URL, `localhost` vs `127.0.0.1`, `0.0.0.0` and `::` wildcard binds, `[::1]` vs v4, `127.0.0.0/8`, implicit 80/443, differing port — all classified correctly.

## The scheme-less regression — fixed

`HTTPS_PROXY=127.0.0.1:8080` no longer throws `Invalid URL` out of `startHttpProxy`. Nine hostile values all resolve without throwing and fall back to a direct connection, and the warning leaks nothing: `http://alice:s3cr3t-token@[bad` produces only `(Invalid URL)`. `keepAlive: true` is now set, so the description's reuse claim matches the code.

## Please fix before merge: the PR makes four pre-existing tests env-dependent

`startHttpProxy` now reads proxy environment at startup, so raw-passthrough tests inherit whatever the developer has exported. You correctly wrapped one pre-existing test with `replaceOutboundProxyEnv()` and left the rest unguarded.

```
# head, with an unreachable ambient proxy
HTTPS_PROXY=http://127.0.0.1:9 pnpm vitest run tests/http-proxy-server.test.ts
  -> 4 failed | 17 passed
     forwards first-party request bytes and auth unchanged
     preserves compressed first-party request bytes, encoding, and auth
     logs Haiku passthrough status, error body, and system fallback preview
     logs a partial upstream error body when the origin resets before end

# same command with base src/ (4ee1a49)
  -> 2 failed | 19 passed   (only the two NEW tests, whose feature is absent — expected)
```

So it is zero pre-existing failures at base and four at head. The irony is that this bites exactly the population the PR serves: anyone actually behind a corporate proxy. Neutralising the proxy environment once for the whole describe block, and letting the two new tests opt back in, is the smallest fix.

Worth noting how easy this is to miss — my own first run of the full suite came back 84 files / 1323 tests green *in a shell that has `HTTPS_PROXY` set*, because that proxy happened to point at a live clodex server. With the same variable pointing at a dead port, four tests fail. A reachable ambient proxy hides it.

## Minors

1. **The self-loop guard's test asserts the message, not the behaviour.** Deleting the guard entirely fails only the `console.error` string check; no assertion covers the fact that no agent was created or that the passthrough went direct. If the warning text is ever reworded the test will be rewritten to match, and the behaviour will be unguarded.

2. **A self-reference by non-loopback address is not detected.** `proxyUrlTargetsListener` treats only same-host or loopback-against-loopback/wildcard as self, so `http://0.0.0.0:<port>` or this host's LAN IP against a `0.0.0.0` bind returns false and an agent is created. I want to be precise about the impact, because my first read overstated it: under the production default `rejectUnauthorized: true` this is a **bounded 502** with a peak of one socket, not a runaway — the unbounded loop reproduces only with `anthropicRejectUnauthorized: false`, which is a test-only option no production caller sets (`buildConfiguredHttpProxyOptions`, `src/http-proxy/index.ts:116`). So: worth closing or commenting, not a re-open.

3. **Nit:** `anthropicAgent` is assigned after `listenTcpServer` resolves, and the assignment awaits a dynamic `import()`. With a fixed `--port`, a request arriving in that window forwards direct rather than through the configured proxy. Resolving the agent before the listen removes it.

4. **Nit:** the new `anthropicAgent?.destroy()` in `close()` is uncovered — deleting it leaves all 21 tests in the file green. Same for the changed `outboundWsProxyAgent` behaviour (now keep-alive, now swallowing errors), which has no direct test.

camjac251:
Addressed the remaining merge blocker and the closely related lifecycle coverage at head `fc44400`, rebased onto upstream 2.4.0 (`ad907a5`).

- The selective-proxy test block now scrubs every ambient proxy variable before each test and restores it afterward; proxy-specific cases opt in explicitly. With `HTTPS_PROXY=http://127.0.0.1:9`, the affected file passes 22/22.
- The self-loop regression now proves behavior by spying on agent connection attempts; when the configured proxy targets the listener, no connection attempt is made.
- Wildcard listeners now recognize local interface addresses, with deterministic positive and negative coverage.
- Agent construction is synchronous after the listener is bound, removing the post-listen import window.
- Teardown coverage proves the passthrough agent is destroyed, and the WebSocket path now has direct keep-alive and malformed-configuration coverage.

Verification on this head:

- dead ambient proxy regression: 22/22 passed
- outbound proxy tests: 14/14 passed
- full suite: 85 files, 1,378 tests passed
- typecheck and production build passed
- documentation check and staged secret scan passed
- hosted `test` and `commitlint` checks passed

bman654 (approved):
Verdict: **approve**. The requested test isolation is fixed by the right mechanism, the affected suite is stable under a dead ambient proxy, and nothing blocking remains.

## Requested change

| Ask | Status | Verification |
| --- | --- | --- |
| The PR made four pre-existing tests depend on the developer's proxy environment; scrub that environment for the selective-proxy block and let proxy-specific tests opt back in | **ADDRESSED** | With `HTTPS_PROXY=HTTP_PROXY=http://127.0.0.1:9`, the two affected files pass **36/36** at head. The base affected file passes **19/19** under the identical environment. Removing only the new describe-level hooks makes exactly the four previously reported tests fail again. |

The isolation is at `tests/http-proxy-server.test.ts:206-215`. It uses the helper and complete six-name list at `tests/http-proxy-server.test.ts:18-39`, clearing uppercase and lowercase `HTTPS_PROXY`, `HTTP_PROXY`, and `NO_PROXY`, then restoring each exact prior value after the test. The restore remains exception-safe: an injected test that changed `HTTPS_PROXY` and threw did not poison its successor, and the ambient values were restored after the block.

Removing those hooks under the dead proxy produces exactly:

```text
Tests  4 failed | 18 passed (22)
  × forwards first-party request bytes and auth unchanged
  × preserves compressed first-party request bytes, encoding, and auth
  × logs Haiku passthrough status, error body, and system fallback preview
  × logs a partial upstream error body when the origin resets before end
```

No assertion was weakened or removed to obtain the green run. The positive proxy path remains a real wire-level CONNECT test at `tests/http-proxy-server.test.ts:472-552`: it verifies destination, proxy authorization derived from URL credentials, request path/auth/body, and the streamed response.

## Last round's optional items

| Item | Status |
| --- | --- |
| Self-loop test asserted only warning text | **Fixed behaviorally.** The test now asserts that `HttpsProxyAgent.prototype.connect` is not called (`tests/http-proxy-server.test.ts:562-610`, specifically `:599`). Removing the guard at `src/http-proxy/server.ts:998-1008` fails on that assertion before the warning assertion. |
| Wildcard listener did not recognize a local non-loopback interface | **Covered as defense in depth.** The branch is at `src/outbound-proxy.ts:103-106`, with deterministic injected-address positive and negative cases at `tests/outbound-proxy.test.ts:117-128`. Removing the branch fails the matrix. |
| Agent construction awaited a dynamic import after binding | **Fixed in code.** `HttpsProxyAgent` is statically imported at `src/outbound-proxy.ts:22`, and the assignment at `src/http-proxy/server.ts:1007` is synchronous. There is no dedicated regression test for the absence of an async yield—adding one before the assignment leaves 22/22 green—but this was optional cleanup, not a merge requirement. |
| Agent teardown and WebSocket keep-alive/error paths were uncovered | **Fixed.** Teardown is covered at `tests/http-proxy-server.test.ts:612-628` and implemented at `src/http-proxy/server.ts:1020-1026`; deleting `destroy()` fails that test. Malformed-URL fail-open and HTTP/WebSocket keep-alive are covered at `tests/outbound-proxy.test.ts:79-109`; deleting the URL guard fails one case, and removing `{ keepAlive: true }` at `src/outbound-proxy.ts:150` fails two. |

## Cross-family same-port classifier case

The classifier does have a real false-positive mechanism: on macOS, an IPv4 wildcard listener and an IPv6 loopback listener can coexist on the same numeric port, while `proxyUrlTargetsListener` treats `[::1]:P` as self for a `0.0.0.0:P` bind. I reproduced both coexistence and the `true` classification.

It is not production-reachable in clodex today. The built CLI rejects `server --host 0.0.0.0` as an unknown option. The only production caller of `startHttpProxy` is `startConfiguredHttpProxy` (`src/http-proxy/index.ts:161-176`), and `buildConfiguredHttpProxyOptions` hardcodes `host: '127.0.0.1'` (`src/http-proxy/index.ts:116-141`). `startHttpProxy` also defaults to loopback (`src/http-proxy/server.ts:985-1012`). The `0.0.0.0` network listener at `src/server/index.ts:398` is the endpoint gateway, not this selective MITM. Therefore this is not a blocker or a requested change.

## Optional follow-ups

1. **OPTIONAL:** `outboundWsProxyAgent` is now synchronous (`src/outbound-proxy.ts:161-166`), but its production caller still awaits it (`src/oauth/responses-websocket.ts:1966-1969`), as do the HTTP helper tests at `tests/outbound-proxy.test.ts:63-99`. This is behaviorally harmless; leaving it is reasonable given the instruction not to restructure the WebSocket continuation file.

2. **OPTIONAL:** the dynamic-import delay is gone, but assignment still happens after `listenTcpServer` returns. The listener's callback resolves before the readiness probe at `src/listener-ready.ts:134-151`, so a small bind-to-assignment interval remains. Clodex does not spawn or advertise its own client until startup resolves, making this non-blocking. If desired, construct the candidate agent before binding, then retain it or destroy/clear it after comparing the resolved listener address.

Independent verification at `fc44400`: full suite under a dead ambient proxy, **85 files / 1,378 tests**; typecheck clean; production build clean. Hosted `test` and `commitlint` checks are also green.

PR #93, opened 2026-08-08
fix(sdk): strip the volatile Claude Code billing header on every translated route

zachspartofaday:
## Problem

Claude Code prepends a billing-attribution system block (`x-anthropic-billing-header: …; cch=…;`) whose `cch` value changes on every request. `translateRequest` strips it only when `openAiOAuth` is set, so every API-key route ships a volatile first system line. That caps upstream prompt-prefix caching at the first few hundred tokens: OpenAI-compatible gateways hash the request prefix for implicit caching, and third-party Anthropic-format providers cache up to breakpoints — both diverge at the header on every call.

The explicit-breakpoint path (gpt-5.6+ public API) has the same leak independently: `translateTopLevelSystemForOpenAi` re-emits the raw top-level system blocks as system messages, billing block included.

**Reachability:** every `translateRequest` caller without `openAiOAuth` — API-key `@ai-sdk/openai`, `@ai-sdk/openai-compatible`, third-party `@ai-sdk/anthropic` — receives Claude Code bodies containing the header (it is on every Claude Code request). Observed live before the fix (OpenCode Zen, kimi-k3): cache reads frozen at 256-512 tokens per call while per-call uncached input grew 23k→76k across a session; a direct byte-stable request to the same upstream cached 5888/6104, isolating the divergence to the translated prefix.

## Fix

- `translateRequest`: strip unconditionally (`systemToString(body.system, true)`). OAuth behavior is byte-identical — that route already stripped; only the gate widens. Anthropic passthrough does not go through `translateRequest` and still forwards the header.
- `translateTopLevelSystemForOpenAi`: apply `stripClaudeCodeBillingHeader` per block before emitting system messages.

Side benefit: `openAiPromptCacheKey(baseSystem, …)` now hashes the stripped system, so the API-key fallback cache key stops churning per request too.

## Validation

- Measured after the fix, same churning-`cch` pattern via `clodex server --endpoint`: deepseek-v4-flash second call **58 uncached / 6272 cached** (was all-uncached); kimi-k3 **215 uncached / 5888 cached**; grown-conversation third call keeps the cached prefix.
- `tsc --noEmit` clean; full vitest suite green on this base (90 files / 1296 tests).
- The former OAuth-only strip test now asserts every translated route, promptCacheKey stability under changing `cch` on the API-key route, and billing-block removal from the explicit-breakpoint system messages.

### DIFF
--- a/src/sdk-adapter.ts
+++ b/src/sdk-adapter.ts
@@ -207,10 +207,12 @@ function translateTopLevelSystemForOpenAi(
 ): ModelMessage[] {
   if (!system) return [];
   if (typeof system === 'string') {
-    return system.trim() ? [{ role: 'system', content: system } as ModelMessage] : [];
+    const stripped = stripClaudeCodeBillingHeader(system);
+    return stripped?.trim() ? [{ role: 'system', content: stripped } as ModelMessage] : [];
   }
   return system.flatMap(block => {
-    const text = typeof block === 'string' ? block : block.text ?? '';
+    const raw = typeof block === 'string' ? block : block.text ?? '';
+    const text = stripClaudeCodeBillingHeader(raw) ?? '';
     if (!text.trim()) return [];
     const cacheControl = typeof block === 'string' ? undefined : block.cache_control;
     return [{
@@ -471,9 +473,15 @@ export function translateRequest(
 
   // Claude Code prepends an Anthropic-only billing attribution block whose
   // `cch` value changes every request. It is envelope metadata, not a model
-  // instruction, and forwarding it to OpenAI would invalidate the stable
-  // prompt prefix. Anthropic passthrough and non-OAuth providers are untouched.
-  const baseSystem = systemToString(body.system, options?.openAiOAuth === true);
+  // instruction, and forwarding it invalidates the stable prompt prefix on
+  // EVERY translated provider — OpenAI-compatible gateways hash the request
+  // prefix for implicit caching, and third-party Anthropic-format providers
+  // cache up to breakpoints, so a volatile first system line caps their cache
+  // hits at the first few hundred tokens (observed live: kimi-k3 via OpenCode
+  // Zen froze at 256-512 cached tokens per call). Strip it on every
+  // SDK-translated route; Anthropic passthrough does not go through
+  // translateRequest and still forwards it.
+  const baseSystem = systemToString(body.system, true);
   const systemText = baseSystem?.trim() || (options?.openAiOAuth ? 'You are a coding assistant.' : undefined);
 
   // resolveUpstreamTools uses the shared proxy types; the adapter keeps its own
--- a/tests/sdk-adapter.test.ts
+++ b/tests/sdk-adapter.test.ts
@@ -205,7 +205,7 @@ describe('translateRequest', () => {
     expect(params.maxOutputTokens).toBeUndefined();
   });
 
-  it('strips Claude Code Anthropic billing attribution from OpenAI OAuth instructions only', () => {
+  it('strips Claude Code Anthropic billing attribution on every translated route', () => {
     const body = {
       model: 'gpt-5.6-terra',
       system: [
@@ -216,6 +216,10 @@ describe('translateRequest', () => {
       ],
       messages: [{ role: 'user' as const, content: 'hello' }],
     };
+    const changedCchSystem = [
+      { text: 'x-anthropic-billing-header: cc_version=2.1.207.9bb; cc_entrypoint=cli; cch=cb57d;' },
+      body.system[1]!,
+    ];
 
     const oauth = translateRequest(body, '@ai-sdk/openai', { openAiOAuth: true });
     expect(oauth.providerOptions?.openai?.instructions)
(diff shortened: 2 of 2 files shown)


bman654 (approved):
**Approving this — the change is correct and I'd land it as-is.** The central risk you flagged yourself, that widening the gate might perturb the OAuth route, is empirically clean, and both halves of the edit are independently covered by tests.

Everything below the verdict is for the record rather than a request. There are a couple of places where the mechanism is described differently from what the binary actually does — **including in our own code comment, which is where your wording almost certainly came from** — but none of it changes what the change does or whether it's worth having.

---

## OAuth byte-identity: confirmed

I didn't want to take this from reading, so I ran the real `translateRequest` from `main` (`827215a`) and from this head side by side over a generated corpus: 24 system-block shapes × 6 model ids × 4 npm packages × {session id present, absent} × {oauth, non-oauth} = 2,304 cases. Shapes covered absent/null/empty system, header-as-string with and without trailing newline, header as its own block, header **not** first, two consecutive header blocks, header inline with a newline before real instructions, `cache_control` on the header block / the instruction block / both, and five deliberate near-misses.

```
[oauth] compared 1152 cases; diffs=0
[oauth-breakpoints] offenders=0     # no promptCacheBreakpoint / promptCacheOptions in any OAuth case
```

Two structural facts back it up: `systemToString(body.system, true)` is literally the same expression value as the old gated call when `openAiOAuth` is true, and `translateTopLevelSystemForOpenAi` is only reachable through `supportsExplicitOpenAiCaching`, which is `!options?.openAiOAuth && …` — the OAuth route never enters it.

## The API-key routes do improve

```
[non-oauth] changed=624 unchanged=528
```

Every shape that changed carries a billing block; nothing else moves. On `@ai-sdk/openai-compatible` (your kimi-k3 route):

```
main: instructions = "x-anthropic-billing-header: …\nYou are Claude Code.\nFollow the user instructions."
head: instructions = "You are Claude Code.\nFollow the user instructions."
```

and the fallback cache key stops churning, as you say — `main` moves `relay-7641ee50…` → a different value between calls; `head` holds `relay-fad93b5c…`. The session-derived key is unchanged in both, as expected.

## Anthropic passthrough still forwards the header — confirmed

Traced rather than assumed. `translateRequest` has exactly two callers and both sit behind a gate that excludes the Anthropic format (`proxy.ts:556` under `isSdkMigratedNpm(route.npm)`, and `server/router.ts:399` inside the `modelFormat === 'openai'` branch). The `'anthropic'` branches return first and forward the system array untouched. The opposite direction is also intact: on Anthropic OAuth passthrough clodex still *adds* the block when missing.

---

## For the record — mechanism notes, no action needed

**1. `cch` isn't the volatile field, and our comment is the source of that.** Reading the header builder out of the installed bundles (same code in 2.1.208 / .219 / .221 / .224):

```js
i = getAPIProvider(),
s = i === "firstParty" && isFirstPartyBaseUrl() || i === "vertex" ? " cch=00000;" : "",
u = r && i === "firstParty" && isFirstPartyBaseUrl() ? ` cc_prev_req=${r};` : "",
d = `x-anthropic-billing-header: cc_version=${n}; cc_entrypoint=${o};${s}${l}${c}${u}`
```

`cch` is the literal `00000`. The genuinely per-request field is `cc_prev_req`, the request id of the most recent assistant message, which changes every turn. Your premise — that a volatile component is poisoning the prefix — is right; the field name isn't.

To be clear about where that came from: **the comment on `main` at `sdk-adapter.ts:471` already said "`cch` value changes every request", and that's ours and it's wrong.** You inherited it. We'll fix it on our side; no need to touch it here.

**2. Both fields are gated on the base URL, which makes the effect bridge-mode dependent.** `isFirstPartyBaseUrl()` is `!ANTHROPIC_BASE_URL || host === "api.anthropic.com"`. In **proxy** mode (clodex's default) Claude Code has no override, the gate is open, and the block genuinely churns every request — there your fix does exactly what it says. In **endpoint** mode `ANTHROPIC_BASE_URL` points at `127.0.0.1`, the gate is closed, and the block reduces to `cc_version=<ver>.<3 hex>; cc_entrypoint=cli;` plus optional `cc_workload` / `cc_is_subagent`, where the 3-hex suffix is derived from the first user message and is stable for the life of a conversation.

Since the Validation numbers were taken via `clodex server --endpoint`, the improvement you measured there is most likely a stable prefix *across* conversations plus a smaller prompt, rather than de-churning within one. Same direction, different mechanism — and in proxy mode it's the de-churning story you described. Worth knowing if you ever re-measure, but it argues for the change either way.

**3. The `@ai-sdk/anthropic` rationale describes a route that can't be reached.** `isSdkMigratedNpm` is exactly "not `@ai-sdk/anthropic`", and it guards both `translateRequest` call sites, so `translateRequest(body, '@ai-sdk/anthropic', …)` isn't a production configuration — third-party Anthropic-format providers take raw passthrough and still receive the header. The change remains right for the two routes that *are* reachable (`@ai-sdk/openai` without OAuth, and `@ai-sdk/openai-compatible`). Flagging it only so a future reader doesn't go hunting for that path; the extra test case is harmless as a lock on the function's own contract.

**4. Matcher precision.** `stripClaudeCodeBillingHeader` matches on `startsWith(prefix)` then drops everything to the first newline without validating the line. One corpus case regresses:

```
input : "x-anthropic-billing-header: is a header.\nDocument it carefully."
head  : "Document it carefully."
```

Reachability is genuinely low — the binary always emits the header as its own block, first, with no newline — and the imprecision already existed on the OAuth route; this widens its blast radius to every translated route. If you ever want to tighten it, only dropping a block that is *nothing but* the header line would match what production actually produces. Not worth a round trip on its own.

---

## Verified clean — don't re-derive these

- **Breakpoint placement unaffected in production.** Across all 24 shapes on the public `gpt-5.6` path, the only delta is when `cache_control` sits on the billing block itself — and the binary rules that out, assigning that block `cacheScope: null` so it never carries a breakpoint to lose.
- **Inline system messages keep their positions** — only the top-level billing message disappears; interleaved reminders stay exactly where they were.
- **Cache-key semantics** — the session-derived key is untouched; only the fallback changes, and it changes to the stable value.
- **Under-strip variants** — casing and leading-whitespace variants correctly aren't matched; a header with no `cch` (the shape clodex itself injects on passthrough) is still stripped, since the match is prefix-only.
- **No hidden consumers** — grepping the header *value* across `src/` (not one comparison shape) turns up nothing else reading it.

## Build and test

```
pnpm typecheck   # clean
pnpm test        # 83 files / 1296 tests passed
pnpm build       # success
```

The test count matches your claim exactly; the file count is 83 rather than 90. Both halves of the change fire independently under reverse mutation — reverting the gate fails the strip test, and reverting only the breakpoint path fails it differently — so neither edit is riding on the other's coverage.

One unrelated flake to ignore: `tests/http-proxy-index.test.ts` timed out on two of three full-suite runs here and passes alone; it binds a real listener and is sensitive to local port contention. Nothing to do with this change.

---

Nice find, and the differential evidence in the PR body made this much faster to verify. Approving.

PR #94, opened 2026-08-08
feat(providers): OpenCode Go provider (offered — expands provider scope)

zachspartofaday:
**Scope note up front:** this adds a non-OpenAI provider, which goes beyond clodex's current definition (bridging Claude Code to OpenAI models). We built and run it on our fork and are offering it upstream in case you want it — no hard feelings at all if it's out of scope; feel free to close. It is self-contained (a provider template + its catalog; no changes to existing provider behavior beyond the metadata extensions described below).

## What

Four commits, reviewed on our fork (zachspartofaday/clodex #1, #2, #11 plus a credential-probe fix):

1. **`feat: support mixed-protocol provider metadata`** — template infra the provider needs: per-model `staticModels` metadata (`ProviderTemplateModel`), `staticModelPolicy: 'overlay' | 'allowlist'`, `preserveModelPricing`, and an optional `verifyCredential` probe for providers whose `/models` endpoint answers without auth.
2. **`feat: add OpenCode Go provider`** — `opencode-go` template (API key at `https://opencode.ai/zen/go/v1`). One credential, two wire protocols: Anthropic Messages models pass through; Chat Completions models translate through `@ai-sdk/openai-compatible`. Responses-only entries (Grok, mainline GPT) are excluded. Live `/models` is availability data; the committed catalog supplies per-model protocol, endpoint, context, modalities, pricing, and compatibility as an allowlist.
3. **`fix: probe OpenCode Go credentials with a catalog model`** — the add-time probe rejects only auth-shaped errors, so a model outage can't reject a valid key.
4. **`feat: source the catalog directly from models.dev`** — per-model metadata comes from OpenCode's own catalog service; wire transports and compatibility patches are maintained in the updater script as live-validated knowledge (models.dev publishes no transports, and family-level rules are wrong per model — minimax-m3 is Anthropic-format while minimax-m2.x are Chat Completions). New models.dev ids are reported as unmapped rather than guessed into a routing bug.

Pairs well with #93: with the billing-header strip, these gateways cache properly (measured: kimi-k3 second call 215 uncached / 5888 cached through the translation path).

## Validation

- `tsc --noEmit` clean on this base; vitest 1327/1327 green (adds catalog-contract, template-metadata, and credential-probe suites)
- Run in production on our fork since 2026-08-07 across the 17-model catalog (kimi, deepseek, qwen, minimax, glm families)


bman654 (changes requested):
**On the scope question you raised up front: we'll take it.** No need to split it or keep it on your fork.

You were right to flag it, and I looked hard at the blast radius before answering. Of the +2,174/−110, only ~520 insertions land in existing `src/`, and `sdk-adapter.ts`, `oauth/responses-websocket.ts` and `patch-transforms.ts` are untouched — so the hard-won translation, caching and patcher paths are genuinely not in play. I verified the "no change to existing provider behavior" claim rather than taking it on trust (details below): the two existing providers come out byte-for-byte identical across every shared surface. That's what makes this an easy yes.

The one thing I can't do is **exercise the provider itself** — I don't have an OpenCode Go key and won't be running that service. That doesn't change the answer, but it does shape one request: I'd like the README to say plainly which providers we support ourselves and which depend on their contributor. See the first item below.

Verdict is **changes requested** for the items in §2 — one functional bug, one thing only you can confirm, and a user-visible hang — none of which are structural.

---

## 1. Please add a "Supported providers" section to the README

A short table listing the providers and their support tier, so users know what they're relying on before they configure something, and so we have one obvious place to document providers as they're added:

| Provider | Auth | Support |
|---|---|---|
| OpenAI | API key | Fully supported by the clodex maintainer |
| OpenAI (ChatGPT / Codex plan) | OAuth | Fully supported by the clodex maintainer |
| OpenCode Go | API key | Community-supported — maintained by its contributor |

Wording is yours to pick; what matters is that OpenAI (both auth modes) reads as first-class and maintainer-supported, and OpenCode Go reads as depending on contributor support. That's an honest description of the situation rather than a hedge — we're glad to ship it, we just can't validate it against the live service or debug it when the vendor changes something.

## 2. Requested changes

### 2a. Major — a `compatibility` block that says nothing about reasoning still disables reasoning effort

`effortProviderOptions` takes its new branch on the mere *presence* of `metadata.compatibility`:

```ts
if (npm === '@ai-sdk/openai-compatible' && modelId && metadata?.compatibility) {
  const reasoningEffort = compatibilityReasoningEffort(effort, modelId, metadata.compatibility);
  if (!reasoningEffort) return undefined;   // <-- early return skips every downstream rule
```

When the block carries only wire-shape fields, `compatibilityReasoningEffort` returns `undefined` meaning "no opinion", and that early return swallows the model-name rules that used to handle it. `getReasoningCapabilities` does *not* short-circuit, so the two disagree. `kimi-k2.7-code` in the committed catalog is exactly this shape:

```
kimi-k2.7-code  caps=controllable  patchLevels=-      IF-NO-COMPAT patchLevels=low/medium/high
     wire(with)=(none)
     wire(without)=low={"opencodeGo":{"reasoningEffort":"low"}} medium={...} high={...}
```

So it advertises controllable reasoning, shows zero levels in a patched client, and never puts `reasoning_effort` on the wire. `kimi-k2.6` is the deliberate version of this (it sets `supportsReasoningEffort: false`), which is what makes the contrast clear — the catalog has a way to say "no reasoning effort" and this entry isn't using it.

Suggested shape: only enter the branch when the block actually expresses reasoning intent —

```ts
const hasReasoningOpinion = c && (c.reasoningEffortMap
  || c.supportsReasoningEffort !== undefined || c.thinkingFormat !== undefined);
```

— then either let `kimi-k2.7-code` inherit the Kimi rule, or set `supportsReasoningEffort: false` on it explicitly if suppression is what you want.

### 2b. Major, pending your confirmation — `count_tokens` is forwarded upstream on the new passthrough routes

`proxy.ts` answers `/v1/messages/count_tokens` from a local estimate only when `route.modelFormat !== 'anthropic'`; anthropic-format routes forward it upstream. Before this PR no clodex provider was anthropic-format, so that branch never ran in practice. It does now, for the four Qwen/MiniMax passthrough models:

```
ANTHROPIC-FORMAT count_tokens -> upstreamPath= /v1/messages/count_tokens
    clientStatus= 404  body= {"type":"error","error":{"type":"not_found_error","message":"Not Found"}}
OPENAI-FORMAT  count_tokens -> clientStatus= 200  body= {"input_tokens":12}
```

I can't check whether `https://opencode.ai/zen/go/v1/messages/count_tokens` exists without a key — this is exactly the kind of thing the support-tier note above is about. **Please confirm it against the live service.** If it isn't implemented, letting these routes use the same local estimate the openai-format routes already use is the cheap fix; surfacing a 404 into Claude Code's token accounting is the failure mode to avoid.

### 2c. The credential probe has no timeout and can hang the CLI

`verifyOpenCodeGoCredential` calls bare `fetch` with no `signal`, while every other network call on this path is bounded by `TEST_TIMEOUT_MS = 10_000`. It also runs *first*, so it's the first thing a user hits after pasting a key. Against a stubbed upstream that accepts then stalls:

```
✓ does not abort when the upstream never answers 3015ms
```

Under `spinner.start('Testing connection to OpenCode Go...')` the symptom is a spinner that never stops and no exit but Ctrl-C. A blackholed connection — captive portal, egress filter, a proxy that accepts and stalls — is enough. `signal: AbortSignal.timeout(TEST_TIMEOUT_MS)` is Node 17.3+, fine against the `>=22` floor, and the existing `catch {}` already treats an abort as inconclusive so fail-open semantics don't change.

### 2d. Anchor the auth classification so an entitlement rejection can't read as a bad key

```ts
authRejected = /auth/i.test(errorType) || /api key|unauthoriz|authentication/i.test(message);
```

`/auth/i` is unanchored and matches inside "Unauthorized", so:

```
input : 403 {"error":{"type":"UnauthorizedModelError","message":"Your plan does not include deepseek-v4-flash."}}
VERDICT: "OpenCode Go rejected this API key (authentication failed). Check the key and try again."
```

The probe names one fixed model, so if `deepseek-v4-flash` is ever plan-gated, a valid key on the wrong plan is rejected at add time with a message blaming the key. To be explicit: **I did not verify that OpenCode Go emits such a type** — no live calls were made. This is "the classifier is wider than its own comment claims", not an observed failure. Anchoring it is cheap.

### 2e. `CLAUDE.md` is now stale

It still says "only two providers exist" and describes clodex as bridging to OpenAI models only. Since that file is what the next contributor reads first, this PR should carry the update — including the three cross-cutting concepts it introduces (`verifyCredential`, `staticModelPolicy`/`preserveModelPricing`, and `ModelRuntimeCompatibility`), which are currently documented only in their own doc comments.

## 3. Smaller items — your call, none blocking

- **`preserveModelPricing` doesn't self-heal after a downgrade.** It's read from the provider record, never the template, so an older clodex that parses and re-saves `providers.json` drops it permanently. Consequence is cost display only; a `?? getTemplateById(provider.templateId)?.preserveModelPricing` fallback makes it idempotent.
- **Six models advertise controllable reasoning that never reaches the wire** (`gpt-5.6-luna`, `glm-5.1`, `mimo-v2.5`, `mimo-v2.5-pro`, `minimax-m2.7`, `minimax-m3`). `getPatchReasoningCapabilities` filters the patched client to an empty level list so nothing breaks, but `defaultEffort` is still set and any surface reading `getReasoningCapabilities` reports a control that does nothing. An explicit `supportsReasoningEffort: false` would make the metadata match reality.
- **Register the trace secret before the probe.** `registerTraceSecret` sits inside `fetchTemplateModels`, which runs after the probe — so the first authenticated request with a fresh key happens outside the redaction guard. No leak today (the probe logs nothing and its error string is a fixed literal), but moving it to right after `apiKey.trim()` makes the invariant "registered before first use".
- **A structural invariant test over the whole catalog.** `buildOpenCodeGoModels()` blind-casts the JSON with no runtime validation, and only two entries are spot-checked. A ~10-line loop asserting every entry's host is `opencode.ai` and that `npm`/`apiUrl` match its `modelFormat` would catch a botched regeneration that the current tests would ship.
- **Streaming `responseModelOverride` is untested end to end.** Disabling both branches in `relayAnthropicMessages` only kills the JSON test — `anthropicSseModelRewrite` has good direct tests, but nothing asserts the relay actually pipes through it, so deleting `.pipe(...)` leaves the suite green. Given this is the auto-compaction echo invariant, a streaming counterpart is worth it.
- **Generator ordering:** `writeFile(MODELS_PATH, …)` runs before `updateSourceConstant`, so if that regex ever stops matching you're left with a new catalog and a stale timestamp. Also worth validating `contextWindow` (integer + range) and `name` (printable) at ingest — a bad `contextWindow` flows into the patched context map and breaks auto-compaction.
- **Catalog test literals.** `toHaveLength(17)`, the 4/13 split, and `gpt-5.6-luna`'s exact `cost` read as a deliberate tripwire on regeneration, which I think is the right instinct — but a one-line comment saying so would spare the next person four opaque failures after an upstream price change.
- **Nits:** CRLF loses its `\r` on rewritten SSE lines (cosmetic; Anthropic sends LF); a multi-line `data:` payload falls through unrewritten (fail-open, and Anthropic never emits that shape); `applyTemplateModelMetadata`'s `_baseUrl` is unused; `materializeTemplateModel` is unreachable today; `staticModels: buildOpenCodeGoModels()` runs a 17-entry `structuredClone` at module load on every CLI invocation, including `--help`.
- **Wording:** when the live list is healthy but shares no ids with the catalog, "Connected but no models were returned" points at the wrong cause. And 17 new selectable models against `MAX_MODEL_CATALOG = 20` makes that cap much easier to hit — worth a line in the docs.

---

## 4. Verified clean — please don't re-derive these

**Existing providers are byte-for-byte unchanged.** I built a fixture registry with realistic `openai` and `openai-oauth` caches and dumped every shared surface both branches touch — `materializeRegistry`, the `localProviderToRegistry` round trip, `localProvidersToServerModels`, `localModelToRoute`, `buildHttpProxyRoutes` with a saved alias, all three reasoning surfaces across five effort levels, and `fetchTemplateModels` against a mocked `/models` — then diffed:

```
$ diff -u /tmp/diff-main.json /tmp/diff-head.json && echo IDENTICAL
IDENTICAL
```

The three shared-path edits that look risky on reading are all inert for the existing providers, and the diff proves it rather than arguing it.

**The echo invariant holds end to end, both protocols.** Canonical id, short alias, and the `[1m]` suffix all echo verbatim while upstream still receives the bare id:

```
requested=clodex:opencode-go:qwen3.7-max[1m] -> responseModel="clodex:opencode-go:qwen3.7-max[1m]"
requested=qmax                               -> responseModel="qmax"
requested=clodex:opencode-go:glm-5.1         -> responseModel="clodex:opencode-go:glm-5.1"
```

`anthropicSseModelRewrite` is correct at every chunk boundary (split at every 7th byte), and the non-streaming path recomputes `Content-Length` from the rewritten bytes.

**The `/v1` constraint is right in both consumers**, protocol selection can't mis-fire (`modelPrefersResponsesApi` is never consulted on the openai-compatible branch — `gpt-5.6-luna` was the case that could have bitten, and it correctly goes to Chat Completions), and a deliberately malformed entry fails closed with a 400 rather than building a bad URL.

**Allowlist semantics fail closed both directions** and an error result never clobbers the cache. **Registry downgrade is safe** — an older clodex neither rejects nor corrupts a new-format `providers.json`, and per-model `compatibility`/`modalities` survive because cached models are stored opaquely. **`patch-transforms.ts` is untouched**, so `PATCH_TRANSFORMS_VERSION` correctly stays at 3.

**The supply-chain surface is narrower than it looks, and this is the design decision that made the security review come back clean:** `toClodexModel` hardcodes `apiUrl`, `npm`, `modelFormat` and the whole `compatibility` block locally, and ids are filtered against a local transport map. The feed controls only `name`, `contextWindow`, `cost`, `modalities` and `reasoning`. **A hostile or wrong feed cannot produce a bad base URL, a bad SDK package, or a code-execution path** — worth stating in the script's header comment, because it isn't obvious. The fail-closed `throw` when a transport-mapped id vanishes is the right call too.

**Credentials:** the key is never written to config in plaintext (only `authRef`), never interpolated into an error path, respects the outbound proxy dispatcher, and cannot fire without consent — `verifyCredential` has exactly one call site, downstream of the password prompt. The credential-storage and chunking code is untouched.

**Tests: 11/11 mutations killed**, run as full files rather than `-t` filters — including the round-trip mutations (`convert.ts` / `materialize.ts` dropping `compatibility`/`modalities`), which are the ones that usually slip through.

**Packaging is clean:** catalog JSON inlined by esbuild (nothing needed in `files` for it), `npm pack --dry-run` includes `docs/opencode-go.md`, no `dist/` tracked, no version hardcoded, **no new dependencies at all**, and nothing past the Node 22 floor.

## 5. Checks

```
pnpm typecheck   # clean
pnpm test        # 85 files / 1327 tests
pnpm build       # success, dist/cli.js 2.82 MB
```

Two full runs each reported one failure — but a *different* test each time (`http-proxy-index`, then `credential-helper`), neither file touched by this PR, both passing in isolation, and the proxy-lifecycle one **also fails on `main`**. Bisected across all four commits and green at every point. These are timing flakes on a loaded machine, not yours — flagging so a red CI run on either isn't misread.

---

Thanks for raising the scope question openly instead of just shipping it — that's what made it quick to say yes to. The allowlist-over-live-discovery design and keeping routing constants out of the upstream feed are the two decisions that made this reviewable, and the mutation score speaks for itself.

zachspartofaday:
> [!IMPORTANT]
> **Edited — please don't review this yet.** The §1/§2/§3 work below is pushed and stands. But one premise underneath the reasoning-effort parts was wrong, and I would rather correct it and verify properly than have you spend a review round on it.
>
> **The correction:** I described models.dev as "the closest thing to an authority on what OpenCode's GATEWAY accepts". It is not. It is a community-maintained catalog of model specifications — open TOML in GitHub, created by OpenCode but independent of it — and its own docs describe model metadata, not provider API contracts. OpenCode uses it to source model *names*. I inferred "OpenCode uses this feed" into "this feed describes OpenCode's API", which does not follow.
>
> **The underlying gap is real, though, and worth naming.** As far as I can find, nothing published by OpenCode or by models.dev states which reasoning-effort values *their gateway* accepts per model. Both describe what a MODEL is capable of; neither describes what this deployment in front of it exposes. Those are different questions, and the difference is not theoretical — Z.ai documents seven effort levels for GLM-5.2 and Moonshot three for K3, while the catalog lists `high/max` and `max`, and I have no published source that adjudicates.
>
> So this is knowledge clodex has to establish and maintain itself, per model, as part of catalog updates — the same class of thing the transport map in `update-opencode-go-models.mjs` already is, and what "validated against the live endpoint" in that file is supposed to mean. It is ongoing maintenance for this provider rather than a one-off, and worth knowing before taking the provider on.
>
> **What follows for the code:**
> - The cross-check I added **failed** an update whenever a local map sent a value the catalog did not list. Far too strong for a community catalog: it would block a ladder somebody had validated live. Downgraded to advisory — it reports divergence in either direction, since an unlisted value may equally mean the catalog is incomplete.
> - `glm-5.2` and `kimi-k3` keep their narrow ladders, now marked `UNVERIFIED` and for an honest reason: narrow is the safe side of an unverified guess, because sending fewer levels hides capability while sending a rejected one fails the request.
>
> **What is happening next:** a probe script that calls the live gateway per model and per effort level, comparing each against a no-effort baseline so a value the gateway silently *ignores* is reported inert rather than supported (`qwen3.6-plus` is exactly that — `reasoning_effort` is inert there and `enable_thinking` does the work). Whatever it finds goes into `PATCHES` with the date and method.
>
> I will push that commit and comment again when it is done. Everything below is accurate apart from the models.dev framing in **The thing worth taking away from this PR**, which the above supersedes.

---

Thank you for this — the depth of §4 in particular saved a lot of re-derivation, and the two decisions you singled out (allowlist over live discovery, routing constants out of the feed) turned out to be load-bearing in ways I'll come back to below.

Pushed: **§1, §2a–§2e and all of §3**. It went through three review rounds on my fork first ([#18](https://github.com/zachspartofaday/clodex/pull/18)) rather than arriving here raw.

`pnpm typecheck` clean · **1352 tests / 87 files** · `pnpm build` clean · generator parses.

## Where I departed from your suggestion, and why

**§2b — route-scoped capability, not an anthropic-format rule.** OpenCode Zen documents `/v1/responses`, `/v1/chat/completions` and `/v1/messages` and no token-counting endpoint, so your 404 was real and it is now answered locally. But I first implemented it as "anthropic-format ⇒ no `count_tokens`" and **an existing test caught that**: it would silently downgrade every *custom* Anthropic-compatible endpoint too, not just Zen. It is a `supportsCountTokens` capability on `ModelRuntimeCompatibility` now, stamped on the anthropic-format entries by the generator. Only an explicit `false` diverts; unset keeps forwarding.

**§2a — one shared predicate rather than a local check.** `compatibilityExpressesReasoningIntent` gates *both* the effort path and the capability path, so they cannot drift apart again. On your suggestion for `kimi-k2.7-code`, I took the second option: Moonshot documents *"thinking is always on and cannot be disabled"* and no `reasoning_effort`, so letting it inherit the Kimi rule — which §2a alone does — would have made it start sending a parameter the model does not accept. Suppressed explicitly.

**§3.8 — `static-seed` documented, not deleted.** No template declares it, but removing the arm would make a future one fall through to the api-list path and try to fetch a catalog it does not have.

## Corrections to §3.2

Measuring all 17 models across all seven effort levels rather than applying the list as given: the real count was **nine**, and two of your six were wrong.

- **`kimi-k3` was never broken** — it maps only `max`, which a low/medium/high probe misses.
- **`kimi-k2.7-code`** needed suppression, per above.
- **`qwen3.7-max`, `qwen3.7-plus`, `qwen3.8-max`** were missing from the list. Resolved structurally rather than by vendor claim: effort reaches an upstream only via `effortProviderOptions` and the `thinkingFormat` transform, both of which act on an `OpenAiCompatibleRequestBody`, and a passthrough Messages body is forwarded untouched. That is now a generator default for all anthropic-format entries.
- `glm-5.1`, `minimax-m2.7`, `minimax-m3`, `mimo-v2.5`, `mimo-v2.5-pro` confirmed against their vendors' references — each uses a `thinking` object, none accepts `reasoning_effort`.

**Zero models now advertise a control they cannot operate.**

## The thing worth taking away from this PR

models.dev publishes a per-model `reasoning_options`, and because OpenCode routes from that feed it describes **what the gateway accepts** — which is not the set the upstream vendor documents for its own endpoint. Z.ai documents seven effort levels for GLM-5.2 and Moonshot three for K3; the feed publishes `high/max` and `max`. I widened both to the vendor ladders before understanding that, and had to revert them.

So the updater now cross-checks every local map against the feed. The invariant is a **subset** and the asymmetry is deliberate: sending a value the gateway does not publish **fails the update**; sending fewer is reported and left to a human, because widening a ladder is exactly the change that wants live validation rather than a script's confidence. Verified it reproduces both reverts:

```
glm-5.2: maps to low/medium/minimal/none/xhigh, which models.dev
         does not publish (feed: high/max)
kimi-k3: maps to high/low, which models.dev does not publish (feed: max)
```

This also retires a claim I had written and not honoured — the header previously said `pnpm test` asserted this agreement, and nothing did.

## Two decisions I have left to you

**`deepseek-v4-flash` is narrower than the feed.** models.dev publishes `low/high/max`; the catalog sends `high/max`. Safe direction, so the check reports rather than fails — but adding `low` wants a live call, which I cannot make.

**`qwen3.6-plus` is a coarse control wearing grades.** Qwen has no `reasoning_effort` at all — a boolean `enable_thinking` plus `thinking_budget`. `thinkingFormat: 'qwen'` injects that boolean *only when a `reasoning_effort` is present*, so the effort value is load-bearing as an internal signal even though upstream ignores it. Two consequences:

- The original map dropped `off`, `minimal` **and `max`**, so choosing MAX silently disabled thinking while `low` enabled it. Fixed — every level except `off` now maps.
- Grading it properly means sending `enable_thinking` + `thinking_budget` and dropping `reasoning_effort`, which changes the wire to a gateway I cannot exercise. Tracked as [#19](https://github.com/zachspartofaday/clodex/issues/19), with the underlying `thinkingFormat` conflation as [#20](https://github.com/zachspartofaday/clodex/issues/20).

## One pre-existing gap this surfaced

Not caused by this PR and not fixed by it. `projectNativeEffort` requires a capability to declare all of `low`/`medium`/`high`; most gateway ladders do not. Sweeping all 17:

```
deepseek-v4-flash  [high/max]      → DISCARDED
deepseek-v4-pro    [high/max]      → DISCARDED
glm-5.2            [high/max]      → DISCARDED
hy3                [off/low/high]  → DISCARDED
kimi-k3            [max]           → DISCARDED
gpt-5.6-luna, qwen3.6-plus         → kept
```

Only two of seventeen survive, so patched clients offer no effort control for models the gateway genuinely grades. It is also a hidden constraint on catalog work — it is how an earlier round of mine shipped a single-level capability that looked tidy and left patched clients with nothing. Written up as [#24](https://github.com/zachspartofaday/clodex/issues/24); it touches `patch-transforms.ts` and would bump `PATCH_TRANSFORMS_VERSION`, so it wants its own PR rather than riding along here.

## Everything else in §3

`preserveModelPricing` reads through to the template so an older round trip cannot drop it · `registerTraceSecret` moved ahead of the probe · new whole-catalog invariant test (hosts, `npm`/`apiUrl`/`modelFormat` agreement, context windows, printable names) · streaming relay test with a real `Writable`, since the old object mock meant `.pipe(res)` was never exercised and deleting the SSE rewrite left the suite green · generator validates `contextWindow`/`name` at ingest and resolves the constants content before any write, so the likely failure aborts with both files untouched · CRLF preserved on rewritten SSE lines · unused `_baseUrl` dropped · `staticModels` lazy so `--help` no longer clones 17 entries · "Connected but no models were returned" now distinguishes an empty upstream list from a healthy one sharing no ids with the allowlist · tripwire literals commented · `MAX_MODEL_CATALOG` documented with where it actually binds.

Happy to split any of this out if you would rather review it in pieces.

bman654:
moved PR to draft to prevent claude from trying to review it again before you finish the research

zachspartofaday:
Superseded by the focused OpenCode stack:

1. #99 — mixed-protocol provider foundation
2. #100 — base OpenCode Go provider and hardening
3. #101 — pinned CLI resolver contracts and deterministic catalog
4. #102 — Anthropic Messages route provenance
5. #103 — global exact-effort policy and native representation
6. #104 — same-target alias preservation

The adjacent account, service-tier, and usage-limit work was separated into #98, #97, and #96 respectively.

Each OpenCode slice has focused and complete local validation. Upstream CI is green on every published head. A tree proof was also run at every stack boundary: each reconstructed endpoint exactly matches the conflict-free synthetic merge of the corresponding #94 behavior endpoint with current upstream `main`; the final #104 tree matches #94's complete behavior tree after that baseline merge.

#100–#104 remain draft while their precursors are unmerged because GitHub cannot use a fork branch as an upstream PR base. As each precursor lands, its successor will be rebased, revalidated, and narrowed automatically to the focused delta described in its PR body.

Closing this monolithic PR preserves the existing review history here while moving further review to coherent, dependency-ordered diffs.

PR #96, opened 2026-08-10
fix(transport): classify usage-limit terminals

zachspartofaday:
## Summary

- treat output-less OAuth Responses WebSocket failure terminals as transport errors instead of successful empty responses
- classify bounded terminal fingerprints through the shared upstream-error rules, including numeric error codes
- map usage-limit terminals to HTTP 429 semantics with bounded `retry-after` handling
- keep persisted diagnostics free of upstream prose while retaining a stable message fingerprint

This is the focused usage-limit/terminal slice extracted from #94. It is independent of the OpenCode provider stack and the account-selection work.

## Reachability

`src/oauth/responses-websocket.ts` is the active transport for ChatGPT/Codex OAuth Responses models. The new behavioral cases drive the same terminal-frame handler used by the live WebSocket path. Reverting the source changes makes the output-less failure and usage-limit assertions fail.

## Validation

- `pnpm exec vitest run tests/responses-websocket.test.ts` — 106 passed
- `pnpm typecheck` — passed
- `pnpm test` — 84 files / 1,330 tests passed
- `pnpm build` — passed
- `git diff --check` — passed

Local validation used Node 22.23.1, the package's supported runtime floor. CI exercises the repository's pinned Node 24 target.

## Safety

The change is surgical within the OAuth continuation transport and shared upstream-error classification. Tests use synthetic frames. No prompts, credentials, headers, account identifiers, provider payloads, or raw traces are included.

### DIFF
--- a/src/oauth/responses-websocket.ts
+++ b/src/oauth/responses-websocket.ts
@@ -12,7 +12,7 @@ import type { FetchFunction } from '@ai-sdk/provider-utils';
 import type { RawData, WebSocket as WsWebSocket } from 'ws';
 import { CODEX_RESPONSES_WEBSOCKETS_BETA } from '../constants.js';
 import { outboundWsProxyAgent } from '../outbound-proxy.js';
-import { anthropicErrorType, clampRetryAfterSeconds } from '../upstream-error.js';
+import { anthropicErrorType, clampRetryAfterSeconds, frameStatusCode } from '../upstream-error.js';
 import { sanitizeToolInput } from '../tool-input-sanitize.js';
 
 const RESPONSES_LITE_HEADER = 'x-openai-internal-codex-responses-lite';
@@ -882,7 +882,7 @@ function boundedDiagnosticIdentifier(value: unknown): string | undefined {
 }
 
 function diagnosticTextFingerprint(
-  field: 'errorMessage' | 'closeReason',
+  field: 'errorMessage' | 'closeReason' | 'upstreamMessage',
   value: unknown,
 ): Record<string, unknown> {
   if (typeof value !== 'string' || value.length === 0) return {};
@@ -1598,10 +1598,30 @@ function handleSocketMessage(entry: ConnectionEntry, data: RawData): void {
   const errorStatus = type === 'error' && !ctx.emittedModelData
     ? responseErrorStatus(event)
     : undefined;
+  // A failure terminal with no HTTP status to map (`response.failed`,
+  // `response.incomplete`, or a bare `error` frame that named no status) is the
+  // same defect one layer over: forwarded verbatim it ends the stream having
+  // emitted nothing, so downstream reads a SUCCESSFUL, EMPTY 200 and renders a
+  // turn in which the model said nothing. That is indistinguishable from a
+  // deliberately silent turn, so the actual reason — a usage limit, a rejected
+  // reconstruction, an aborted response — never reaches the operator and the
+  // session simply appears to stop answering, one prompt after another.
+  //
+  // Gated on `!emittedModelData` for the same reason as the branch above: once
+  // model data is downstream the stream is committed and the existing
+  // partial-output path must keep handling it.
+  const emptyFailureTerminal = FAILURE_EVENT_TYPES.has(type ?? '')
+    && errorStatus === undefined
+    && !willRetry
+    && !ctx.emittedModelData;
   // One rejection, one diagnostic record. Without this gate a rejected request
   // emits both this record and `failContext`'s, under different `source` values
   // with disjoint fields, reading as two failures of one request.
-  if (FAILURE_EVENT_TYPES.has(type ?? '') && (errorStatus === undefined || willRetry)) {
+  if (
+    FAILURE_EVENT_TYPES.has(type ?? '')
+    && (errorStatus === undefined || willRetry)
+    && !emptyFailureTerminal
+  ) {
     emitResponseErrorDiagnostic(entry, ctx, {
       source: 'response_event',
       upstreamEventType: type,
@@ -1653,6 +1673,94 @@ function handleSocketMessage(entry: ConnectionEntry, data: RawData): void {
         ...(retryAfterSeconds !== undefined ? { retryAfterSeconds } : {}),
       },
       errorStatus,
+      retryAfterSeconds,
+    );
+    return;
+  }
+
+  if (emptyFailureTerminal) {
+    const details = responseFailureDetails(event);
+    // The bounded identifiers are the whole diagnostic value here: they are what
+    // names `usage_limit_reached` (or whatever else) in the debug log the next
+    // time a session goes silent, without dumping upstream text.
+    const named = [details.errorType, details.errorCode, details.incompleteReason]
+      .filter((value): value is string => typeof value === 'string');
+    const summary = 'OpenAI ended the response with no output'
+      + (named.length ? ` (${named.join(' / ')})` : ` (${type})`);
+    // Classified by the SAME discriminator rules a status-carrying frame goes
+    // through, rather than a local whitelist that would drift from them.
+    //
+    // That matters beyond tidiness: `context_length_exceeded` has to land on
+    // 400. `isContextLengthExceededError` trusts a frame's structured
+    // discriminator over its prose, and the proxy handlers require a 400 before
+    // emitting Claude Code's prompt-too-long response — so as a 502 a
+    // long-context request is retried as a server fault AND loses the
+    // auto-compaction signal that would have made room for it.
+    //
+    // 502 stands in only where the shared classifier has no opinion (its 500
+    // default): an indeterminate failure genuinely is worth retrying, and 502
+    // is the honest "upstream gave us nothing usable" for an output-less
+    // terminal. Every class it does recognise — deterministic 400s, usage
+    // limits, auth, overload — keeps the status the rest of the stack expects.
+    const discriminator = [details.errorType, details.errorCode]
+      .filter((value): value is string => typeof value === 'string')
+      .join(' ')
+      .toLowerCase();
+    // `incomplete_details.reason` is a DIFFERENT vocabulary from a frame's
+    // code/type, so it is classified here rather than fed to frameStatusCode,
+    // which documents itself as reading the latter. These are the reasons that
+    // describe a settled outcome rather than a fault: retrying re-runs a
+    // request upstream has already answered.
+    const settledReason = details.incompleteReason === 'content_filter'
+      || details.incompleteReason === 'max_output_tokens';
+    // errorCode passed as the CODE argument, not folded into the discriminator:
+    // frameStatusCode recognises a numeric HTTP code only through that first
+    // parameter, and clodex's own synthetic frames set `code` to the
+    // stringified status. Fold it in as prose instead and a terminal carrying
+    // `code: '401'` matches no rule, lands on the 500 default, and gets
+    // rewritten as a retryable 502.
+    const numericOrNamed = typeof details.errorCode === 'string' ? details.errorCode : undefined;
+    const classified = numericOrNamed !== undefined || discriminator
+      ? frameStatusCode(numericOrNamed, discriminator)
+      : 500;
+    const statusCode = classified !== 500
+      ? classified
+      : settledReason ? 400 : 502;
+    const usageLimited = statusCode === 429;
+    // Only when upstream stated one, and only baked into the TEXT: the AI SDK's
+    // chunk schema is a closed zod object that strips `retry_after_seconds`, so
+    // a hint carried only in the frame field never reaches the client. This is
+    // the same reason the status-carrying branch above spells it out.
+    const retryAfterSeconds = usageLimited && responseRetryAfterSeconds(event) !== undefined
+      ? clampRetryAfterSeconds(responseRetryAfterSeconds(event))
+      : undefined;
+    // The BOUNDED summary is the message, upstream's prose is not used at all.
+    // A `response.failed` body can echo request content or backend detail, and
+    // it does not stay in one place: the synthetic error is rethrown by the SDK
+    // and `proxy.ts` writes both the formatted message and `errorContent` to the
+    // persistent proxy log, which `redactTraceLine` only scrubs of known
+    // credential shapes. Protecting the immediate `fail:` line while the same
+    // text reaches disk one frame later is not protection. The summary names
+    // the cause — the identifiers are what diagnose a silent session — and the
+    // raw message survives as a length+hash in the diagnostic.
+    failContext(
+      entry,
+      ctx,
+      retryAfterSeconds === undefined ? summary : `${summary}; retry after ${retryAfterSeconds}s`,
+      {
+        source: 'empty_failure_terminal',
+        upstreamEventType: type,
+        ...details,
+        // Under DISTINCT keys. `failContext` fingerprints the message it was
+        // given after spreading these, so an `errorMessage*` pair here is
+        // overwritten by the summary's — which would silently discard the only
+        // content-free evidence of what upstream actually said, and leave two
+        // failures with the same type and code indistinguishable. `errorMessage*`
+        // now means "what the client was told", `upstreamMessage*` means "what
+        // upstream said", and both survive.
+        ...diagnosticTextFingerprint('upstreamMessage', responseErrorMessage(event)),
+      },
+      statusCode,
       retryAfterSeconds,
     );
     return;
--- a/src/upstream-error.ts
+++ b/src/upstream-error.ts
@@ -146,7 +146,7 @@ function errorCodeValue(record: Record<string, unknown>): string | undefined {
  * clodex's own synthetic frames set `code` to the stringified status, so that
  * exact channel is preferred over any discriminator.
  */
-function frameStatusCode(code: string | undefined, discriminator: string): number {
+export function frameStatusCode(code: string | undefined, discriminator: string): number {
   if (code !== undefined && /^\d{3}$/.test(code)) {
     const numeric = Number(code);
     if (numeric >= 400 && numeric <= 599) return numeric;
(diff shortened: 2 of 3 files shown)


bman654 (approved):
# Review — #96 `fix(transport): classify usage-limit terminals`

This is a good change and it is backed by real verification, not test theatre. Reverting both
source files while keeping the new tests fails 17 tests; neutralising *only* the new
`if (emptyFailureTerminal)` block at `src/oauth/responses-websocket.ts:1681` and leaving every
other line in place fails the same 17 — so the tests are pinned to the new behaviour, not to
incidental scaffolding. The new tests drive real frames through `socket.emit('message', …)` into
the exact handler production registers at `src/oauth/responses-websocket.ts:1887`. The change is
also surgical in the way this file demands: one computed const, one `&&` clause on an existing
gate, one new `if`, plus `export` on `frameStatusCode`; no existing branch is reordered,
extracted, or rewritten. And it is reachable — a fresh census of the local production
ws-diagnostics corpus (10 files, ~1.59 GB, 203,578 records) finds 10 `response_event` upstream
failures, **9 of them output-less with a bounded `type`/`code`** (`service_unavailable_error /
server_is_overloaded` ×8, `server_error` ×1). On base those records are emitted from the gate at
`responses-websocket.ts:1604`, which only fires when `errorStatus === undefined` — i.e. exactly the
frames that ended the stream as a silent, successful empty 200. On head that shape classifies
(`frameStatusCode(undefined, 'service_unavailable_error server_is_overloaded')` → `503`, executed).

**Nothing here blocks the merge.** There is one thing worth fixing and it is comment text only —
the rationale written into the shared classifier states a mechanism that does not exist. It needs
no code change, so please feel free to fold it into a later PR rather than respinning this one.
Everything else is optional. We reviewed this PR in two independent passes, and the second
overturned one of the first's two blocking asks and corrected several of its citations; the
"Checked and clean" section below records those outcomes so nobody re-derives them.

`pnpm typecheck` clean. `pnpm test` on head: 84 files, **1330/1330 passing**.

## Worth fixing (not blocking — a later PR is fine)

### The comment justifying the `usage_limit` → 429 rule states a retry-budget mechanism that does not exist

**Files:** `src/upstream-error.ts:158-161` (the rationale), mirrored verbatim in
`tests/responses-websocket.test.ts:1355-1356`; plus the now-overstated doc claim at
`src/upstream-error.ts:135-139`.

The comment says that without the rule the discriminator "falls through to the 500 default — which
`frameIsRetryable` treats as transient, so the client spends its whole retry budget re-asking a
question upstream has already answered." Three independently executed facts say otherwise:

- `frameIsRetryable` (`src/upstream-error.ts:188-192`) returns `true` for **429 as well as ≥500**,
  and so does the SDK's own `APICallError.isRetryable`. Real `@ai-sdk/openai` + `streamText` at
  `maxRetries: 2`, counting fetch invocations: `status429 → 3 calls`, `status502 → 3`,
  `status400 → 1`, `status401 → 1`. 429 and 502 are indistinguishable to the retry loop. (Run
  independently by both reviewers, same numbers.)
- Snapshotting `src/upstream-error.ts`, deleting *only* the `usage_limit` term from the regex at
  `:162`, and re-running the upstream-attempt counter gives an identical count:
  `usage_limit(429)=3  server_error(502)=3  content_filter(400)=1`, before and after. The mapping
  costs and saves **zero** upstream attempts.
- `frameIsRetryable`'s output never drives a retry. `grep -rn isRetryable src/` resolves to
  production at `src/upstream-error.ts:299`/`:324` and consumption at `src/trace-log.ts:204`/`:543`,
  `src/proxy.ts:737`, `src/server/router.ts:161` — all diagnostic log fields. Nothing branches on it.

What the mapping actually buys is real and worth stating plainly: the client sees `429` /
`rate_limit_error` instead of `502` / `api_error`, and gets a `retry-after` hint. Driven end to end
through the real WebSocket path:

```
SSE message : OpenAI ended the response with no output (usage_limit_reached / usage_limit_reached); retry after 60s
client text : … (HTTP 429)
status/hdr  : 429 60
```

In a repo where comments are the durable design record and this file is DO-NOT-RESTRUCTURE, a
wrong mechanism will be used to justify a wrong future edit.

**Ask:** rewrite `src/upstream-error.ts:158-161` and its mirror at
`tests/responses-websocket.test.ts:1355-1356` to say what the mapping delivers — a surfaced
`rate_limit_error` status and a backoff hint, not retry-budget savings — and soften `:135-139`.
With `usage_limit` added, `frameStatusCode` no longer mirrors `@ai-sdk/openai`'s `getStatusCode`
table (`node_modules/@ai-sdk/openai/dist/index.js:155` is `["insufficient_quota", "rate_limit"]`),
so "Deliberately mirrors" should become "mirrors, plus a documented clodex-specific term". No code
change is warranted, and this does not need to hold up the merge — picking it up in a downstream PR
is fine by us.

## Optional

Everything below is explicitly **non-blocking**.

**Optional 1 — normalize a finite number `code` before classification.** `boundedDiagnosticIdentifier`
(`src/oauth/responses-websocket.ts:876-882`) returns `undefined` for anything non-string, so a
JavaScript `code: 429` reaches `frameStatusCode` as `undefined` and the discriminator is just the
frame type. The generic provider path already handles both forms — `errorCodeValue`
(`src/upstream-error.ts:126-129`) stringifies finite numbers, and the SDK's own `getStatusCode`
reads a numeric code at `node_modules/@ai-sdk/openai/dist/index.js:145`. Parity would be cheap.
This is hardening, not a defect: the bundled Responses schemas declare both the bare-error and
`response.failed.response.error` codes as **strings**
(`node_modules/@ai-sdk/openai/dist/index.js:4423` and `:4476`), and the production census above
found zero numeric-code WS frames — all 10 carry named string codes. (The PR's
`routes a numeric error code through the classifier, not past it` test at
`tests/responses-websocket.test.ts:1268-1280` uses "numeric" to mean a digit-only *string* and
exercises `'401'`/`'429'`/`'503'`; it never claimed number support.)

**Optional 2 — one direct case in `tests/upstream-error.test.ts` for the shared rule.** That file
is entirely green (42/42) with `usage_limit` reverted out of the regex; only
`tests/responses-websocket.test.ts` catches it. `frameStatusCode` is now exported and reached from
`parseProviderErrorFrame` → `sdkUpstreamErrorDetails`, called at `src/proxy.ts:698` and
`src/server/router.ts:152` (inside `auditSdkError`, invoked at `:480` and `:671`) plus
`src/server/router.ts:464`/`:655` on the OAuth credential-replacement path — i.e. every SDK error
including plain API-key routes. The behaviour *is* pinned, just transitively. One row asserting
`frameStatusCode('usage_limit_reached', 'usage_limit_reached') === 429`, plus an `isRetryable`
assertion (which the WS-side test omits), makes the coverage independent of the WS branch
continuing to route through it.

**Optional 3 — pin the `!willRetry` conjunct.** `src/oauth/responses-websocket.ts:1615` can be
deleted with 106/106 green in-file and the whole suite green. Downgraded from MAJOR on two grounds
worth knowing: the retry is still logged identically in `--trace` under the mutation
(`entry.debug('previous response unavailable; retrying once with full context')` at `:1634` is
unconditional, and that is the log users paste into bug reports — the mutation costs one structured
record in the narrower `--ws-diagnostics` ledger only), and the triggering shape is a *status-less*
`previous_response_not_found`, which occurs 0 times in the production corpus and which the code's
own comment at `:1596-1597` documents as carrying a 400. For the documented `status: 400` shape the
structured record survives the mutation. Still ~10 lines to close: add a second row without
`status: 400` to the existing "still logs a retried rejection" test.

**Optional 4 — scope note for prose-shaped identifiers.** By design the branch never reads upstream
prose (correct call). The consequence is that "classify usage-limit terminals" is contingent on
upstream naming the reason in a bounded `type`/`code`; `boundedDiagnosticIdentifier` rejects any
identifier containing a space. Executed against the real `frameStatusCode`:

```
disc="usage_limit_reached"                        -> 429
disc="usage limit reached"                        -> 500   (spaces; also rejected upstream by the bounded filter)
disc=""            (prose-only frame)             -> 500 -> 502
```

None of these is a regression — base returned an empty 200 for all of them. Two cheap follow-ups:
say so in the PR body or a comment so nobody later reads the classification as unconditional, and
pin the `` named.length ? … : ` (${type})` `` fallback at `:1689` — rewriting its text today fails
no test (verified: 106/106 green under that mutation).

**Optional 5 — move the `usage_limit` term below the 400 rule.** `src/upstream-error.ts:162` tests
first, ahead of `authentication`/`permission`/`not_found`/`invalid|bad_request|context_length`, and
the discriminator is a substring join, so a cross-field pair resolves to 429. Executed:

```
frameStatusCode(undefined, 'usage_limit_reached context_length_exceeded') -> 429   (400 expected)
frameStatusCode(undefined, 'authentication_error usage_limit_reached')    -> 429   (401 expected)
frameStatusCode(undefined, 'context_length_exceeded')                     -> 400
```

The 429 verdict on the first row loses the auto-compaction signal the PR's own test protects.
**Reachability: hygiene, not a live bug** — no observed OpenAI/Codex frame pairs a `usage_limit*`
field with a conflicting one, and a 7,200-frame + 102-pair sweep found no cross-field accident
(`usage` + `limit` joins with a space, not an underscore, and does not match). Reordering costs
nothing for the real vocabulary and deletes the class.

**Optional 6 — the 60s clamp is honest for rate limits and dishonest for plan limits.**
`MAX_RETRY_AFTER_SECONDS = 60` (`src/upstream-error.ts:31`) was sized against the 120s idle abort;
a weekly plan limit is measured in hours. Two things established under execution: on the **new** WS
path there is no prose/header disagreement (message, client text and header all say 60; upstream's
real number appears nowhere), and on the non-WS path the 60-vs-21600 pair is **pre-existing base
behaviour** for `rate_limit_exceeded` / `insufficient_quota`. So this is a pre-existing policy
question this PR adds a third string to, not a regression — but "exempt plan-level limits from the
clamp, or omit the header when the stated value exceeds the cap" is a good follow-up issue.

**Optional 7 — small test-pinning nits.** Each mutation below was applied to head and re-run;
`tests/responses-websocket.test.ts` stayed 106/106 green in the first three cases:
- `details.incompleteReason` can be dropped from the summary identifiers at `:1686` — the
  `content_filter`/`max_output_tokens` tests assert only status and retryability, so the client
  message silently degrades from `(content_filter)` to `(response.incomplete)`. Adding
  `expect(body).toContain(label)` to the deterministic-outcome `it.each` would close it.
- The `retryAfterSeconds` argument to `failContext` at `:1764` can be passed as `undefined` and the
  test still passes, because `retryAfterFromText` recovers 60 from the message prose. Asserting the
  raw body contains `"retry_after_seconds":60` would pin the argument.
- `usageLimited = statusCode === 429` (`:1729`) can be widened to `|| statusCode === 503` with the
  file green.
- `expect(record!.errorMessageBytes).not.toBe(29)` (`tests/responses-websocket.test.ts:1389`) is
  satisfied by absence: removing both `errorMessage` fingerprints
  (`src/oauth/responses-websocket.ts:917` and `:1311`) leaves that new test green. Worth noting the
  code is not unpinned — a *pre-existing* test ("logs sanitized upstream response failure details
  after partial output") fails under that mutation. Asserting
  `toBe(Buffer.byteLength(expectedSummary))` would make the new test pin what it claims to.

**Optional 8 — dead ternary.** `const classified = numericOrNamed !== undefined || discriminator ?
frameStatusCode(...) : 500;` (`:1723`) — `frameStatusCode(undefined, '')` already returns `500`
(executed). Calling it unconditionally removes a branch no test can distinguish (106/106 green
under that mutation).

**Optional 9 — an HTTP-boundary assertion.** The new tests stop at the WS handler / SDK
classification boundary. One integration assertion through `startProxyCatalog` or `server/router`
proving the synthetic 429 **and** the `retry-after` header as the client actually receives them
would close the last gap. Optional coverage, not evidence of a defect.

**Optional 10 — two latent notes, no action needed now.** (a) The branch feeds chunk discriminators
into `frameStatusCode`, which the shared module's `CHUNK_DISCRIMINATOR_TYPES`
(`src/upstream-error.ts:91` — note the set holds only `'error'` and `'response.failed'`) exists to
keep out on the parse path; `responseFailureDetails` falls back to `record.type`/`record.code` at
`src/oauth/responses-websocket.ts:913-914`. None matches a current rule, but a future rule
containing `fail`/`error` would misfire only here. (b) `responseRetryAfterSeconds` (`:829-842`)
reads only `retry_after_seconds`/`retry_after`; public codex-CLI usage-limit payloads are often
described as carrying `resets_in_seconds`. No usage-limit frame has ever been captured in the local
ledger, so this is unverified either way — and the plumbing is pre-existing, not introduced here.
Worth confirming against a real frame if one ever lands.

Each of these was raised as a concern, then killed by executed evidence. Items 1-4 were
additionally re-verified in a second independent pass.

1. **The 429 mapping adds zero upstream requests.** Measured twice, independently; see Blocking #1.
   The claim that a usage limit now costs "6 attempts over 62s vs base's 1" fails on three counts:
   the count is unchanged with the mapping deleted, the base "1" is base's *empty-200 success* (the
   defect being fixed), and 6 requires an explicit `CLODEX_UPSTREAM_MAX_RETRIES` opt-in — the
   shipped default is 3, since `upstreamMaxRetries()` returns `undefined` when unset.

2. **No prose/header divergence on the new path.** Driven end to end with upstream stating
   `retry_after_seconds: 21600` *and* "retry after 21600s" in prose: SSE message, client text and
   header all say 60; `21600` and the planted prose appear nowhere. The base `error_frame` branch —
   which appends onto `responseErrorMessage(event)` — is the one the `:1643-1652` warning is about;
   the new branch is the one that obeys it.

3. **The `!willRetry` guard is a diagnostics nit, not a correctness bug.** Under the mutation a
   status-less `previous_response_not_found` still retried successfully with
   `retry_prev_id=null`, full context replayed, and the `--trace` line intact; only the narrower
   ws-diagnostics record disappears. See Optional 3.

4. **The pre/post-output split is not production-reachable.** The mechanism is real
   (`PRE-OUTPUT head={"s":500}` vs `POST-OUTPUT head={"s":429}`), but the "OAuth chat models not on
   the WebSocket transport" class is empty: `shouldUseOpenAiResponsesEndpoint`
   (`src/provider-factory.ts:83-85`) is false only for `OPENAI_CHAT_COMPLETIONS_ONLY` =
   `['davinci-002','babbage-002','gpt-3.5-turbo-instruct']` (`:77-81`), which the ChatGPT/Codex
   backend does not serve; the full 55-test provider-factory file confirms every current OpenAI
   model routes to Responses. Every OAuth route that can emit `usage_limit_reached` is a WS route,
   where `failContext` writes `code: String(statusCode)` and the numeric channel wins. That leaves
   api.openai.com API-key routes, whose vocabulary is `insufficient_quota`/`rate_limit_exceeded`.
   Hygiene, addressed by the doc softening in Blocking #1.

5. **The safe retry window is NOT widened, and continuation is unharmed.** The new branch calls
   `failContext` directly — it never enters `retryTransportFailure`/`handleTransportFailure`, never
   stores a chain head, and is gated on `!ctx.emittedModelData` exactly like the branch above it.
   Terminal + socket `close` + `error` produces no replacement socket and exactly one downstream
   frame (`sockets=1, frames=1, status=502`) — identical counts at base.

6. **No head is poisoned and no bad `previous_response_id` is picked next turn.** `failContext` →
   `deleteEntry` unregisters before the head-update block can run: turn 3 after an output-less
   failure opens a fresh socket with `previous_response_id = null` and full input. **Identical at
   base** — this PR touches neither the head lifecycle nor the pool.

7. **`previous_response_not_found` retry-once still works**, including the status-less shape:
   `sockets=2`, `retry_prev_id=null`, full context replayed, same as base. A second prnf on the
   replacement now surfaces as 404 non-retryable instead of base's silent empty 200 — an
   improvement.

8. **Partial output can never be retried, twice over.** Removing `&& !ctx.emittedModelData` fails
   two *pre-existing* tests, and independently the SDK refuses to replay a stream that already
   produced content (`afterOutputCalls: 1` for 429/502/400/401 alike). Partial-output bodies are
   byte-identical to base.

9. **The retry ladder still clears the 120s idle abort.** At `maxRetries=5` six attempts took
   ~62,000 ms — the SDK's `2+4+8+16+32` ladder, not `5 × 60s`, because `getRetryDelayInMs` reads
   `retry-after` only from response *headers* and clodex's synthetic frame never sets one. The clamp
   value is irrelevant to that ceiling.

10. **No upstream prose reaches disk, even with the content opt-in on.** Driven through the real
    `writeWebSocketDiagnosticLog` + `writeInferenceResponseErrorLog` + `formatUpstreamError` into
    real files **with `CLODEX_LOG_REQUEST_PREVIEW=1`**: a planted secret appears in none of the SSE
    body, the `onDebug` lines, the ws-diagnostics JSONL, the inference-error JSONL, or the
    client-facing string. Persisted `errorContent` is the bounded synthetic frame only.

11. **The fingerprints work and are distinct.** Identical prose → identical `upstreamMessageHash`;
    different prose → different hash; both distinct from the client-message fingerprint. The
    "distinct keys" argument in the comment at `:1754-1760` holds.

12. **The response-model echo invariant is untouched** — no line in `src/http-proxy/` changes.
    Correcting an over-broad claim made during review: `src/http-proxy/` does not import
    `upstream-error` directly, but `src/http-proxy/server.ts:9` imports `startProxyCatalog` from
    `../proxy.js`, and `src/proxy.ts:49` imports `sdkUpstreamErrorDetails` — so translated
    MITM-mode requests *do* pass through the classifier indirectly. What is unaffected is the
    request/response echo path; only error classification on those routes is touched, which is the
    intended change. `src/sdk-adapter.ts:819` is a log label only (used solely at `:820`).

13. **429 cannot hijack the OAuth credential-replacement retry** — `resolveOAuthRetryReplacement`
    returns `null` unless `status === 401` (`src/upstream-forward.ts:60`), reached from
    `src/proxy.ts:700-707` and `src/server/router.ts:466-473`/`:657-664`.

14. **`max_output_tokens` / `content_filter` cannot trigger a false auto-compaction** (400 with
    `ctxLen: false`), and a genuinely empty `response.completed` is untouched — it is not in
    `FAILURE_EVENT_TYPES` (`src/oauth/responses-websocket.ts:20`), still establishes the head, still
    retains the socket.

15. **The rewritten pre-existing test did not trade a guard for a weaker one.** Making
    `responseErrorStatus` start reading `record.response.status` still fails "does not adopt a
    numeric response status as the mapped status".

16. **Sibling PRs.** `git merge-tree --write-tree` is clean for #96 against each of #97, #98 and
    #99; none of them touches a file this PR changes, and #99's provider-factory changes do not
    alter OAuth WS routing.

17. **Suite health.** `pnpm typecheck` clean; `pnpm test` 1330/1330 across 84 files on head. An
    intermittent failure seen under parallel load (`tests/credential-helper.test.ts > force-kills a
    helper that exceeds the runtime limit`, and `tests/server-index.test.ts` timeouts) reproduces on
    unmutated head *and* on pristine base `4ee1a49`, and passes in isolation on both — load timing,
    not this PR.

PR #97, opened 2026-08-10
feat(oauth): add service-tier fast mode

zachspartofaday:
## Summary

- add `clodex claude --fast` as launch-scoped Codex fast mode for ChatGPT/Codex OAuth routes
- support `CLODEX_SERVICE_TIER=auto|default|flex|priority|fast`, with `fast` normalized to `priority`
- keep OpenAI API-key and non-OpenAI routes tier-free by resolving the route before applying provider options
- record the resolved/requested tier as pre-dispatch diagnostic intent
- warn once when the AI SDK reports that it omitted the requested tier for a model

## Contract and diagnostics

`--fast` composes into the launch environment before prerequisite and dry-run handling, overriding an ambient tier for that launch. A dry run therefore previews the same launch configuration while still performing no persistence or provider request.

The SDK-omission notice uses the SDK's structured `unsupported` / `serviceTier` warning. It does not duplicate a model-name capability heuristic, inject at the transport layer, or echo the configured value. Streaming observes step-finish warnings and non-streaming observes result warnings, including the `codex-auto-review` case identified during review.

The `serviceTier` request-log field is intentionally pre-dispatch intent: it reports what clodex resolved/requested, not proof that the SDK serialized the field or that a backend received it. When the SDK omits a tier, the bounded user-visible notice is the authoritative observation.

## Review resolution

This update addresses both blocking findings and every still-applicable optional item from the maintainer review:

- one-field-at-a-time OAuth/API-key route negatives
- gateway audit-record coverage
- CLI environment precedence and dry-run composition
- invalid-value and SDK-omission warn-once coverage
- hostile-value handling without echoing the value
- shared `isOpenAiOAuthRoute` use at both server handlers
- public environment-variable documentation
- removal of the vacuous API-key assertion
- corrected request-log and PR coverage wording

The complete reconstructed branch was reviewed in [zachspartofaday/clodex#39](https://github.com/zachspartofaday/clodex/pull/39) before this upstream update. After the OpenAI-format warning gap was fixed and a model-universal SDK-omission claim was disproved with a synthetic differential capture, Codex reported no major issues on exact HEAD `c9e2bb295b`; the fork PR checks passed and the fork PR was merged.

## Reconstruction

This branch is rebuilt directly on current upstream v2.4.0 (`ad907a54`):

- feature commit: `90bdb7a7969502ae209fc028fc74f5078daae09f`
- maintainer-review fix: `122066e1ed2e7089ef53a8dd9151830a56cc5da7`
- fork-review fix / HEAD: `c9e2bb295bee0032b98ea294474e96bd12c43ac7`
- all commits replayed or applied without conflicts
- `git range-diff` maps the original `ac29cb1` feature patch exactly to `90bdb7a`; the other two commits contain only review fixes
- diff limited to 15 service-tier source, documentation, and directly corresponding test files

## Validation

Node 24.19.0 / pnpm 10.34.5 on exact HEAD `c9e2bb295bee0032b98ea294474e96bd12c43ac7`:

- focused service-tier/CLI/proxy/router suite: 7 files / 205 tests passed
- OpenAI-format warning regression subset: 2 files / 84 tests passed
- `pnpm typecheck`: passed
- `pnpm test`: 85 files / 1,382 tests passed
- `pnpm build`: passed
- `git diff --check ad907a5436a13cc807e988ad5b4dff0437483d36..HEAD`: passed

## Safety

- synthetic fixtures and localhost test servers only; no live provider request or real credential mutation
- no provider/model-name heuristic copied into clodex and no transport-level tier injection
- no package, lock, release, generated-catalog, registry/account-selection, pricing, or generated `dist/` content committed
- no raw local logs or sensitive runtime values used as evidence


bman654 (changes requested):
Reviewed at `ac29cb1` against base `4ee1a49`, in three independent passes each attacking the others' conclusions, and then a further pass commissioned specifically to attack the first round's conclusions. That later pass overturned or corrected several of our own findings; what's below is what survived both rounds, with the corrections applied.

The safety property this feature rests on — the public API-key route never carries a billable `priority` tier — holds under every route shape we could construct, and it is **not** unpinned the way an early pass of ours claimed. That claim was wrong and we retract it; details in "Checked and clean" #1. Route discrimination happens after alias/route resolution, a mixed catalog in one process is discriminated per request, the WebSocket transport and responses-lite shaping both carry the tier through untouched, the accepted tier set exactly matches the SDK's zod enum so no accepted value can be rejected downstream, and the new log field is whitelisted, compacted and redacted against 22 hostile env values. Seven findings from the first round were refuted outright and are not repeated as asks. The PR's own validation was real, and its wording is mostly careful — the body says "resolved tier," which is the accurate framing.

Two things are being asked for, and they are the same defect seen from two sides: `--fast` is a silent no-op on one production-materialized OAuth model, and the request-log field that exists to answer "did `--fast` do anything" is resolved intent recorded before dispatch, while its doc comments describe it as wire confirmation.

## Blocking

### 1. `--fast` silently no-ops on `codex-auto-review`, with no user-visible notice

`src/cli.ts:407-408` promises `--fast` will "send `service_tier=priority` on ChatGPT-OAuth models". For the cached OAuth model `codex-auto-review` it does not, and nothing tells the user.

`src/sdk-adapter.ts:527` attaches the tier with no model-capability check:

```ts
const serviceTier = options?.openAiOAuth ? oauthServiceTier() : undefined;
```

`@ai-sdk/openai@4.0.11` then removes it. `dist/index.js:46`, verbatim:

```js
const supportsPriorityProcessing = modelId.startsWith("gpt-4") || modelId.startsWith("gpt-5") && !modelId.startsWith("gpt-5-nano") && !modelId.startsWith("gpt-5-chat") && !modelId.startsWith("gpt-5.4-nano") || modelId.startsWith("o3") || modelId.startsWith("o4-mini");
```

and `dist/index.js:6206-6212` (Responses path, which every OAuth model takes) pushes a warning and runs `delete baseArgs.service_tier` at line **6212** when the model does not support it. Evaluating that predicate directly: `codex-auto-review → false`; the other seven cached OAuth ids (`gpt-5.6-sol/-terra/-luna`, `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex-spark`) → `true`. The SDK's warning never reaches the user: `silenceSdkWarnings()` (`src/proxy-shared.ts:37-41`, sets `AI_SDK_LOG_WARNINGS = false`) is the first statement of both `startProxyCatalog` (`src/proxy.ts:308`) and `startServer` (`src/server/router.ts:182`).

Wire capture through the real `OpenAIResponsesLanguageModel`, reproduced independently in every pass of our review:

```
WIRE service_tier BY OAUTH MODEL
  gpt-5.6-sol / -terra / -luna / gpt-5.5 / gpt-5.4 / gpt-5.4-mini / gpt-5.3-codex-spark  -> "priority"
  codex-auto-review                                                                      -> (ABSENT)
silenced-run: service_tier = undefined, warnings captured = 0
```

Reachability, stated precisely: `codex-auto-review` is present in the local `~/.clodex/providers.json` cache with `upstreamModelId: "codex-auto-review"` (catalog id == wire id, so no rename rescues it), and the real `materializeRegistry` produces it for both the `claude` and `server` agents with `authType:'oauth'`, `npm:'@ai-sdk/openai'`, Responses routing. It survives `shouldHideModel` — nothing in `src/data/model-incompatible.json` or `src/` filters it. It is selectable, favoritable and aliasable. **One correction we owe you:** that cache's `fetchedAt` is `2026-07-24T00:52:24.339Z`. This is production-materialized state from a real local cache, not a catalog queried live during review, and an earlier pass of our review should not have called it that. `createResponsesWebSocketFetch` is installed as the SDK's `fetch` (`src/provider-factory.ts:188-196`) — i.e. *below* `getArgs`, so the delete has already happened before the transport sees the body. `flex` has the same hole against `supportsFlexProcessing` (`dist/index.js:45`), and any future Codex slug not starting with `gpt-4`/`gpt-5`/`o3`/`o4-mini` inherits this on the day it lands in a fetched catalog.

Honest bounds so the ask is sized right: this is 1 of 8 cached OAuth models, there is **no billing leak and no functional break** — the request still succeeds, it just runs at the default tier.

**Ask:** make the user contract truthful — either ensure the tier is actually serialized for this ChatGPT-OAuth backend, or emit a bounded user-visible unsupported/no-op notice for models on which it is dropped.

Explicitly **not** asked for: copying the SDK's model-prefix heuristic into clodex as the preferred fix. That heuristic describes public-API Enterprise support ("requires Enterprise access", per the SDK's own warning text at `dist/index.js:6210`) and is unverified — arguably wrong — for the ChatGPT/Codex backend, where `priority` is exactly the Codex-CLI fast-mode wire value. Mirroring it would hide the mismatch while silently preserving the headline flag's no-op. Equally, no live provider request was made by the PR or by either review pass, so we are not asserting that transport-level injection is safe either; verifying backend acceptance would have to come first. Your call which direction to take.

### 2. The request-log `serviceTier` field records pre-dispatch intent, but its doc comments claim wire truth

This is broader than model #1 and applies on `gpt-5.6-sol` too. The field is written before the adapter is contacted at all. Executed: a real MITM request with a supported OAuth model and an unreachable adapter produced

```
PREWIRE {"status":"502","loggedTier":"priority","failureSource":"adapter_request_error"}
```

No SDK request body ever existed, yet the request record said `priority`. That is not a harness artifact — the PR's own positive test helper hard-codes `adapterHandle: { port: 1, ... }` at `tests/http-proxy-server.test.ts:124-128` and then asserts `serviceTier: "priority"`. So the field is resolved request intent by construction, at all four sites: `src/http-proxy/server.ts:827`, `src/proxy.ts:462`, `src/server/router.ts:394`, `src/server/router.ts:615`.

The comments overstate that. `src/trace-log.ts:192-199` says *"Absent means no tier was sent"* (the sentence is on line 197, the field declaration on 199), and `src/proxy.ts:460-462` says *"its absence in a log is itself the answer to 'did `--fast` do anything on this route'"*. Neither is true of a pre-dispatch field. This matters beyond pedantry because the HTTP-proxy inference log is created unconditionally on the default path — `src/cli.ts:989` is not gated on `parsed.trace` (only `proxyDebugLogPath` on the next line is) — so the record is written on every default proxy-mode `clodex claude --fast` run and is the artifact a user pastes into a bug report. (One correction to an earlier pass of our review: the path is *printed* only inside `if (!agentStdout)` at `src/cli.ts:1034-1037`, so clean-agent-stdout runs create it silently.)

**Ask:** describe the field as the tier clodex resolved/requested — in `src/trace-log.ts:192-199` and `src/proxy.ts:460-461` — rather than as proof of what reached the wire. If you want a field that does prove transmission, that needs post-serialization transport instrumentation; note that "read back what the SDK produced" is *not* local to these four sites, since all four run before the SDK/adapter is invoked.

## Optional

All non-blocking. Several are cheap enough to fold into the same push.

- **One-field-at-a-time negative fixture for `isOpenAiOAuthRoute`.** The new negative test at `tests/http-proxy-server.test.ts:1575-1600` uses `{npm:'@ai-sdk/openai-compatible', authType:'api'}` (lines 1589-1590) against a positive of `{npm:'@ai-sdk/openai', authType:'oauth'}` — it varies both fields, so it is insensitive to the `authType` half, and the shipped `openai` API-key template (`{npm:'@ai-sdk/openai', authType:'api'}`) is never exercised here. This is test design, not an unpinned guarantee (see "Checked and clean" #1). Suggested: change the negative fixture to `{npm:'@ai-sdk/openai', authType:'api'}` and add a third `{npm:'@ai-sdk/openai-compatible', authType:'oauth'}`.
- **Correct the PR body's coverage claim, and optionally pin the gateway/direct-proxy diagnostics.** The body says diagnostic propagation is exercised "through all three launch paths. Reverting the source change breaks those assertions." That holds for the HTTP-proxy path only. Deleting both `src/server/router.ts:394` and `:615` and neutralising `src/proxy.ts:462` kills zero assertions — targeted run of `tests/server-router.test.ts tests/proxy.test.ts tests/proxy-sdk-provider-id.test.ts tests/http-proxy-server.test.ts tests/server-index.test.ts tests/trace-log.test.ts` → 6 files / 109 tests passed, confirmed independently by both passes. Structurally, `grep -rln "serviceTier\|SERVICE_TIER" tests/` returns exactly three files (`sdk-adapter`, `openai-adapter`, `http-proxy-server`), and the gateway audit-sink test asserts with `expect.objectContaining({modelId, effort, provider, route})` (`tests/server-router.test.ts:248-249`) over an `anthropic`/`groq` catalog, so a `serviceTier` key can never move it. Non-blocking because the *default* bridge path is the pinned one, `src/proxy.ts:462` feeds a `plog` string that does not exist without `--trace` (`src/proxy.ts:197`: `if (!debug) return () => {};`), and the gateway has no `--fast` at all. Cheap fix: add an OAuth/non-OAuth pair to the existing audit-sink test.
- **Pin the `--fast` → env edge.** `src/cli.ts:1084` (`if (parsed.fast) process.env.CLODEX_SERVICE_TIER = 'fast';`) is reachable by no test — `runClaudeCommand` has one caller (`src/cli.ts:1560`), and deleting the line leaves `tests/cli.test.ts` green. The code is correct today: it precedes the `bridgeMode === 'proxy'` dispatch at `:1105-1106`, so proxy mode inherits it. A ~20-line probe with `PATH=/nonexistent` pins it without stubbing the wizard — the assignment is the second statement of `runClaudeCommand`, before `findClaudeBinary()` at `:1090`, so the call returns 1 without launching anything. That probe also covers the ambient-`CLODEX_SERVICE_TIER`-is-overridden precedence your comment at `:1080-1084` claims. Verified working: `exit code = 1, tier = fast`.
- **Pin the invalid-value warning and its warn-once cap.** `src/sdk-adapter.ts:586-591`. The PR body says "Invalid values warn once and preserve the backend default"; only the *preserve* half is tested (`tests/sdk-adapter.test.ts:291-294`). Deleting the warning block, and separately replacing `if (!warnedServiceTier)` with `if (true)`, both leave the suite green. Behavior is correct as shipped (measured: 1 stderr line over 50 consecutive calls), but stderr is shared with Claude Code's TUI and `oauthServiceTier()` runs twice per request on the translated OAuth route, so an un-deduped regression would print two lines per turn. Repo convention is the opposite — `tests/responses-websocket.test.ts:2089` spies `process.stderr.write` and asserts both directions.
- **`oauthServiceTier(log?)`'s `log` parameter is dead, and it is the only redacting channel.** `src/sdk-adapter.ts:581,588-591`. `grep -rnE "oauthServiceTier\([^)]" src/ tests/` returns only the declaration; all six call sites (`src/openai-adapter.ts:132`, `src/sdk-adapter.ts:527`, `src/proxy.ts:462`, `src/http-proxy/server.ts:827`, `src/server/router.ts:394,615`) pass nothing, so `try { log?.(message) } catch {}` at `:590` is unreachable. Consequence: the raw env value goes to `console.error` unbounded and unredacted and never reaches the trace log. A probe with a hostile value produced `warn len = 3125 | hasESC = true | hasNewline = true | hasToken = true | redactWouldStrip = true`. **Hygiene, not exfiltration** — the value is the user's own env var and it reaches the terminal, not a `~/.clodex/logs/` file (`grep -rn "process.stderr.write =" src/` → only the temporary mute at `src/launch.ts:120`). One-line fix either way: pass a logger at one site, or drop the parameter and compact the echoed value.
- **Nit — `src/server/router.ts:404` still recomputes the predicate inline.** `const openAiOAuth = model.npm === '@ai-sdk/openai' && model.authType === 'oauth';` sits ten lines below the log site at `:394` that uses `isOpenAiOAuthRoute(model)` specifically so the record mirrors the behavior, while the sibling handler was converted at `:621`. Identical today, so no defect — but it's the drift risk your own doc comment at `:391-393` is written against, on the busiest OAuth route in the product. One line: `const openAiOAuth = isOpenAiOAuthRoute(model);`.
- **Nit — `CLODEX_SERVICE_TIER` is documented only inside `clodex claude --help`.** `grep -rn CLODEX_SERVICE_TIER src/ docs/ README.md` → 7 hits, all in `src/`. It is the only way to select `flex`/`auto`/`default`, the only way to enable the tier on the gateway route this PR instruments (`server --fast` is correctly rejected), and the form that persists across a shell session — where the OAuth-only scoping is least obvious.
- **Nit — `--fast` mutates `process.env` under `--dry-run`, and the value inherits into children.** `src/cli.ts:1084` runs before the dry-run branch; `parseArgs(['claude','--fast','--dry-run'])` → `{fast:true, dryRun:true}`. No disk write, so the documented `--dry-run` invariant is intact. `buildChildEnv()` copies `process.env`, so the launched claude and every `clodex-claude` agent process carries `CLODEX_SERVICE_TIER=fast` — harmless today (nothing in the child reads it), but a `clodex server` started from inside that session would inherit fast mode. Flagged from reading, not exercised.
- **Nit — vacuous assertion.** `tests/openai-adapter.test.ts:137`'s `expect(apiKey.providerOptions).toBeUndefined()` holds identically on `main`: the non-OAuth branch of `translateOpenAiRequest` returns an object with no `providerOptions` key at all, feature present or absent. Confirmed indirectly — dropping the `serviceTier` spread failed exactly one assertion, the OAuth one at `:130`. The real non-OAuth exclusion is pinned at `tests/sdk-adapter.test.ts:278-285`, so nothing is lost; the assertion just doesn't carry the meaning its comment claims.
- **Sibling PRs.** `git apply --3way --check` is clean for #96, #98 and #99 against #97. #99 overlaps `src/cli.ts`, `src/proxy.ts`, `src/server/router.ts`, `src/types.ts`; applied on top of #97 it typechecks and passes 8 tier/routing/registry files / 226 tests, with the wire result unchanged (7 `priority`, `codex-auto-review` absent). #98 preserves provider-level `authType:'oauth'` when selecting an account, so the discriminator is unaffected; #96 does not rewrite the outgoing payload shape. No interaction blocker — just re-run after any sibling merges, since all four are based on main.

Each of these was a live suspicion someone spent real cycles on. Recorded with the correcting facts so nobody re-derives them.

1. **"The public-API-key exclusion is pinned by no test / a billing regression ships silently" — FALSE.** We filed this twice at MAJOR and it was refuted both times, then independently re-verified in a later pass. Mutating `isOpenAiOAuthRoute` (`src/sdk-adapter.ts:573-577`) to npm-only turns the suite **red, deterministically**: `tests/server-router.test.ts > does not force streaming for non-streaming requests on API-key routes` (`:1046`, a test that predates the PR — same line in `4ee1a49`) fails with 1 failed | 31 passed; restored, 32 passed. Separately, the leaf gate that actually decides it — `translateRequest`'s `options?.openAiOAuth` — is pinned by a test **this PR added**: making it unconditional fails `translateRequest > applies CLODEX_SERVICE_TIER on the OAuth route only`. Two independent gates, both red when broken. The residual is test *design* only (the catching test is named after streaming, and the new fixture varies two fields) — that is the first optional item, not a shippable-green billing regression.
2. **The exclusion holds under every shape we could build.** Real `startProxyCatalog` HTTP path with the gate computed by production code, `CLODEX_SERVICE_TIER=fast`: `public-api-key gpt-5.6 → undefined`, `gpt-4o → undefined`, `o3 (responses) → undefined`, `gpt-5-codex → undefined`, `authType UNSET → undefined`, `short alias → api-key route → undefined`, `openai-compatible → undefined`, `OAUTH chatgpt → priority`, `short alias → OAUTH route → priority`. Including the specific worry that an API-key provider whose model prefers the Responses API might slip through — it does not; the gate is on the route, not the endpoint choice. Nine env-value variants (`fast`, `priority`, `PRIORITY`, `" priority "`, `flex`, `auto`, `default`, `warp`, `""`) on the public API-key route: all `undefined`. A second independent pass's wire control agreed — no `service_tier` on the public API-key route for `fast`, `priority`, `flex` or `auto`.
3. **Structurally, no `authType:'oauth'` route can reach the billable public API.** `src/provider-factory.ts:170-173` hard-codes `baseURL: 'https://chatgpt.com/backend-api/codex'` for `authType === 'oauth'` and ignores the registry's stored `api.url` — which matters, because the real `providers.json` records `"api": {"url": "https://api.openai.com/v1"}` on the `openai-oauth` provider. With exactly two shipped templates (`openai` = `authType:'api'`, `openai-oauth` = `authType:'oauth'`, both `npm:'@ai-sdk/openai'`), `authType` is a sound discriminator. Corollary: dropping the *npm* half of the predicate is a behavioral no-op for every reachable route today — hygiene, not a bug.
4. **A mixed catalog in one process is discriminated per request**, despite the env being process-global: one `startProxyCatalog` serving an API-key favorite and an OAuth favorite, alternating → `{"gpt#0":"(none)","sol#1":"priority","gpt#2":"(none)"}`.
5. **Ordering follows the resolved route, never the requested alias.** Two aliases through one real MITM request each (`sol` → `@ai-sdk/openai`/`oauth`, `kimi` → `@ai-sdk/openai-compatible`/`api`): `[{"m":"sol","t":"priority"},{"m":"sol"},{"m":"kimi"},{"m":"kimi"}]` — the aliased OAuth route gets the tier, the non-OpenAI route does not, and the record carries it exactly once (the second record per model is the response-side record, correctly tier-free).
6. **The WebSocket transport is not a blind spot.** `createResponsesWebSocketFetch` parses the SDK-built body (`src/oauth/responses-websocket.ts:1867`) and re-emits it verbatim inside the envelope (`outgoingPayload`, `:1453-1454`); `applyResponsesLiteShape` (`:239`) spreads `{...payload}`, and continuation's `sendPayload = { ...payload, input: selectedDelta, previous_response_id }` (`:1920`) does too. 7 of 8 real OAuth ids put `service_tier: "priority"` in the first frame with `useResponsesLite: true` active. Only the SDK's own capability strip removes it.
7. **No path was missed and nothing double-applies or double-logs.** `startProxy` (`src/proxy.ts:816`) delegates to `startProxyCatalog`; `startHttpProxy` creates its adapter *as* `startProxyCatalog` in-process (`src/http-proxy/server.ts:730`); the gateway exposes exactly two routes (`src/server/router.ts:243,248`). `translateRequest` has exactly two callers (`src/proxy.ts:576`, `src/server/router.ts:405`), `translateOpenAiRequest` exactly one (`:622`), and all three pass `openAiOAuth` computed from the already-resolved route. `writeInferenceRequestLog` has two callers serving disjoint servers; `startProxyCatalog` writes no request record at all.
8. **`isOpenAiOAuthRoute` is not structurally inert on any call site.** `ProxyRoute` carries `npm` (`src/proxy.ts:238`) and `authType` (`:241`), populated by `localModelToRoute` and spread by `buildHttpProxyRoutes`; `ServerModel` carries them too (`src/server/models.ts:40` ← `src/registry/materialize.ts:138`, and `localProvidersToServerModels` at `src/provider-catalog.ts:97` maps both into every `ServerModelInfo`). The PR's hand-built routes match production shape.
9. **Diagnostic leakage is clean.** `serviceTier` is the only addition to `InferenceRequestLogEntry`/`writeInferenceRequestLog`, and its value comes solely from `oauthServiceTier()`. 22 hostile `CLODEX_SERVICE_TIER` values (including `__proto__`, `constructor`, a 5000-char string, ANSI control sequences, an embedded `sk-ant-` token, dotted-I/dotless-ı Unicode lookalikes) all resolve to a whitelisted value or `undefined` — the whitelist is a `Set` (`src/sdk-adapter.ts:579`), not an object literal. A hypothetical hostile direct caller is still covered: `'Bearer sk-ant-api03-…\nmore\tstuff' + 'B'.repeat(200)` lands in the log as `"serviceTier":"Bearer [REDACTED] mor"` via `compactLogValue(entry.serviceTier, 40)` (`src/trace-log.ts:424`) + `redactTraceLine`.
10. **The accepted tier set exactly matches the SDK's**, so no value clodex accepts can produce a zod rejection downstream: `SERVICE_TIERS = new Set(['auto','default','flex','priority'])` (`src/sdk-adapter.ts:579`) vs `z.enum(["auto","flex","priority","default"])` in `@ai-sdk/openai@4.0.11`. The `fast → priority` normalization is the right direction (Codex-CLI spelling in, wire value out).
11. **Invalid values do preserve the backend default, and the key is omitted rather than nulled.** `warp-speed` → `undefined` → the `...(serviceTier ? { serviceTier } : {})` spread omits the key; confirmed in captured wire bodies (no `service_tier` key at all, not `null`).
12. **Warn-once cannot spam the shared terminal as shipped.** `warnedServiceTier` (`src/sdk-adapter.ts:580`) is module-level in an ESM module instantiated once, and the MITM/gateway all run in the single clodex process (background agents `exec` claude via `clodex-claude` and never construct a tier). Measured: 1 line over 50 calls. The optional item above is about the missing *test*, not the behavior.
13. **`--fast`'s parse surface has no surprises.** Executed against `parseArgs`: `claude --fast` → `{fast:true}`; `claude --fast --dry-run` → `{fast:true, dryRun:true}`; `claude -- --fast` → forwarded to claude (correct); `server|models|patch --fast` → `error: "Unknown <cmd> option: --fast"`; `clodex --fast` → `error: "Unknown root option: --fast"`. That last one is **not** a bug despite `--fast` being in `STARTER_CLAUDE_FLAGS` — that set is consulted only inside the `claude` branch (`src/cli.ts:327`) to split clodex flags from passthrough flags, and every peer (`--trace`, `--dry-run`, `--proxy`) errors identically at root. `--fast` is consumed, never forwarded to the child CLI.
14. **The two shipped HTTP-proxy tier tests are real, not theatre**, and so are the two adapter tests: dropping the `serviceTier` line from `writeInferenceRequestLog`, dropping the OAuth guard at `src/http-proxy/server.ts:827`, dropping the `serviceTier` spread in `translateOpenAiRequest`, and dropping the `options?.openAiOAuth` gate in `translateRequest` each fail exactly one intended test. One scoping caveat, since it feeds blocking item #2: they assert against the real log file written by a real MITM request, but the adapter behind it is deliberately unreachable (`tests/http-proxy-server.test.ts:126`, `port: 1`), so they pin what the call site *records*, not what is transmitted.
15. **`resetServiceTierWarningForTests` is not an anomaly** — `src/outbound-proxy.ts:73` and `src/oauth/responses-websocket.ts:197,509,623` establish the `…ForTests` export precedent. **Test isolation is sound** — `vitest.config.ts` uses default per-file isolation, no touched test is `it.concurrent`, and every process-wide `CLODEX_SERVICE_TIER` mutation restores in a `finally`. **`src/patch-transforms.ts` is untouched**, so no "coverage" here is the sha256 pin in disguise.

One methodology note for anyone reproducing the numbers above: full-suite counts on our machines are noisy (`tests/credential-helper.test.ts`, `tests/http-proxy-index.test.ts`, `tests/keyring-credentials.test.ts`, `tests/server-index.test.ts` time out intermittently, and `HTTPS_PROXY=http://127.0.0.1:9` without `NO_PROXY=localhost,127.0.0.1` breaks the loopback-server tests outright via undici's `EnvHttpProxyAgent`). Every conclusion above was settled with targeted per-file mutation runs and structural proof rather than by counting; none of the flakes are attributable to this PR.

zachspartofaday:
Addressed every still-applicable item from the review in the rebuilt branch at `c9e2bb295bee0032b98ea294474e96bd12c43ac7`:

### Blocking

1. **Silent SDK omission:** clodex now observes the AI SDK's structured `{type: "unsupported", feature: "serviceTier"}` warning and emits one bounded process-wide notice when the requested tier was omitted. Both Anthropic-format and OpenAI-format endpoints are covered: streaming observes step-finish warnings, while non-streaming observes result warnings. This covers `codex-auto-review` without copying the SDK's provider/model-prefix heuristic, injecting at the transport, making a live request, or echoing the configured value.
2. **Request-log semantics:** `serviceTier` is now documented consistently as resolved/requested pre-dispatch intent, not wire confirmation. The README also states that an SDK omission notice takes precedence over the recorded intent.

### Optional items folded into the same update

- route negatives now vary `authType` and `npm` one field at a time, including the shipped `@ai-sdk/openai` API-key shape;
- OAuth/non-OAuth gateway audit records are pinned in `tests/server-router.test.ts`; the PR body no longer overstates launch-path coverage;
- `--fast` overriding ambient tier state before dry-run/prerequisite handling is pinned in `tests/cli.test.ts`;
- invalid-value and SDK-omission notices are each pinned as bounded/warn-once behavior, including hostile input that is never echoed;
- the unused logging parameter was removed rather than creating a parallel logging/redaction path;
- both server handlers reuse `isOpenAiOAuthRoute`;
- `CLODEX_SERVICE_TIER` values, normalization, OAuth-only scope, dry-run composition, intent semantics, and SDK omission limits are documented publicly;
- pre-dry-run environment composition is retained intentionally and documented/tested because `--fast` is launch configuration, while dry-run still performs no persistence or live provider request;
- the vacuous API-key assertion was removed while the actual API-key exclusion remains pinned at the SDK and route boundaries;
- the rebuilt range was rechecked on current upstream v2.4.0 (`ad907a54`), with no sibling feature commits or unrelated surfaces present.

Validation on exact HEAD `c9e2bb295bee0032b98ea294474e96bd12c43ac7` under Node 24.19.0 / pnpm 10.34.5:

- focused service-tier/CLI/proxy/router suite: 7 files / 205 tests passed
- OpenAI-format warning regression subset: 2 files / 84 tests passed
- `pnpm typecheck`: passed
- `pnpm test`: 85 files / 1,382 tests passed
- `pnpm build`: passed
- `git diff --check ad907a5436a13cc807e988ad5b4dff0437483d36..HEAD`: passed

The exact reconstructed branch was reviewed in [zachspartofaday/clodex#39](https://github.com/zachspartofaday/clodex/pull/39) before this upstream update. Codex reported no major issues on exact HEAD `c9e2bb295b`; the fork PR checks and final local gate passed, and the fork PR was merged. Synthetic fixtures and localhost test servers only; no live provider call, real credential mutation, raw log publication, package/release change, or committed `dist/` output.

bman654 (approved):
# Review

**Approving.** Both asks from the previous round are addressed. The SDK-omission detector is correct, bounded, and pinned by a loudly-red feature-deletion mutation.

One thing to flag before you read further, because it changes what the feature does in practice and it is **our bug, not yours**: the notice cannot currently reach the terminal on `clodex claude`, because `src/launch.ts` mutes stderr for the lifetime of the Claude child. That file is byte-identical on `main` — this is pre-existing infrastructure that our previous review failed to identify when we asked you for a user-visible notice. **We are fixing it on our side in a follow-up**, which also restores several older warnings of ours that the same mute has been swallowing. Nothing for you to do; details below so the behavior is not a surprise.

## Prior asks

| # | Ask | Status |
|---|---|---|
| 1 | Make the `--fast` contract truthful for models where the SDK drops the tier — serialize it, or emit a bounded user-visible notice | **ADDRESSED as far as this PR can** — detection is correct and pinned; delivery is blocked by a pre-existing mute of ours, which we are fixing separately |
| 2 | Document the request-log `serviceTier` field as resolved intent, not wire truth | **ADDRESSED** |

### Ask 2 — verified addressed

`src/trace-log.ts:192-197` now says the field records intent rather than proof of SDK serialization or wire transmission, and `src/proxy.ts:463-464` describes the value as requested intent that the SDK may omit. `README.md:283-286` states the same publicly and distinguishes the pre-dispatch record from a later SDK-omission notice. Nothing further here.

### Ask 1 — the detector itself is correct and pinned

Verified independently rather than inferred from the PR body:

- `reportUnsupportedServiceTier` (`src/sdk-adapter.ts:597-606`) observes the SDK's structured `{type:'unsupported', feature:'serviceTier'}` warning rather than copying its model-prefix heuristic.
- Through the real `@ai-sdk/openai@4.0.11` Responses provider over a capturing fetch, `codex-auto-review` omitted `service_tier` and produced one notice. The other seven OAuth ids in the local cached catalog serialized `service_tier: "priority"` and stayed silent. `flex` discriminated the same way.
- `silenceSdkWarnings()` does not remove the warning data; it suppresses the SDK's own console output, while the `warnings` value delivered to `StepResult`/`GenerateTextResult` remains available.
- The reporter is bounded across the translated Anthropic- and OpenAI-format generation paths. Its message is constant, so hostile `CLODEX_SERVICE_TIER` input is not echoed.
- Feature-deletion mutation is loudly red: making the reporter a no-op fails five tests across the complete `tests/sdk-adapter.test.ts` and `tests/openai-adapter.test.ts` files.

## The delivery gap — ours to fix, not a change request

### The omission notice never reaches the terminal on `clodex claude`

`src/launch.ts:119-120` replaces both `process.stdout.write` and `process.stderr.write` with `muteWrite` for the lifetime of the Claude child. It restores them only from the child `exit`/`error` handlers (`src/launch.ts:122-125`, called at `:141` and `:146`). Every translated request runs during that interval: the gateway/MITM is in this process, and `launchClaude` is awaited at `src/cli.ts:659`, `:1055`, and `:1444`. `console.error` resolves `process.stderr.write` at call time, so the notice at `src/sdk-adapter.ts:605` is swallowed.

I reproduced this with the production `launchClaude` and `reportUnsupportedServiceTier`, bundled and executed in a plain Node subprocess with a real child kept alive long enough to emit the warning:

```text
# notice fired once while the child was alive and once after launchClaude returned
fd2:
  clodex: requested service tier was not sent for this model; the backend default will be used  # after restore
  __AFTER_MARKER__
```

The during-launch notice and marker produced no terminal bytes. With `--debug-file`, `muteWrite` instead copied them to the debug file (`src/launch.ts:107-114`) as parent lines:

```text
[parent] clodex: requested service tier was not sent for this model; the backend default will be used
[parent] __DURING_MARKER__
```

`--debug-file` is passed only under `--trace` (`src/cli.ts:656`, `:1048`, and `:1441`). The post-run summary does not surface the notice: `printTraceLog` (`src/trace-log.ts:646-660`) filters for error/status/model-resolution text, none of which appears in this message.

Reachability is direct. `--fast` is consumed only in the `claude` branch (`src/cli.ts:334`; `server --fast` is rejected at `:238`) and sets `CLODEX_SERVICE_TIER=fast` at `src/cli.ts:1085`. `codex-auto-review` remains a real materialized OAuth route in the local cached catalog and is the one of its eight current rows on which this SDK drops `priority`. Thus the notice is invisible on the flag's own production path. A standalone `clodex server` with `CLODEX_SERVICE_TIER` exported does not call `launchClaude`, so it can display the notice, but that command has no `--fast` option.

That leaves two new claims false on the launch path: the help text at `src/cli.ts:407-408` says the command warns if the SDK omits the tier, and `README.md:284-286` says clodex warns once.

This is pre-existing infrastructure that our previous ask did not identify. `src/launch.ts` is byte-identical on `main`, so this is not a regression introduced by your implementation. The repository also has older warnings whose comments say they are visible in the terminal — for example `src/oauth/responses-websocket.ts:485-489` — even though this mute has been swallowing them on `clodex claude`. A root fix in `launch.ts` is therefore in scope, but the service-tier feature only needs its own bounded notice delivered.

**What we are doing about it.** We are changing `launch.ts` on our side: preserve the existing debug-file copy, then forward bounded parent notices to the captured original stderr writer rather than swallowing them. I executed that shape and it delivers the notice in both traced and untraced launches while retaining the trace copy. Once it lands, your help text at `src/cli.ts:407-408` and the README line become true as written, with no change needed here.

Recording the constraint we are holding ourselves to, since it is the interesting part: the bypass has to stay narrow. Forwarding every chunk beginning with `clodex:` would be broader than any one feature and is not justified by call-count caps alone — other current writers interpolate diagnostic data, and future writers would silently inherit the bypass.

Worth knowing if you ever test this area: an assertion made only against Vitest's replaced `console` cannot measure this behavior at all. It needs the real `launchClaude` driven in a plain Node subprocess (a Vitest test that spawns the probe is fine). That is how the gap was found.

## Demoted — checked and non-blocking

- **The Anthropic collected-force-stream warning hook is unpinned.** Removing `onStepFinish` from the `forceStream` branch at `src/sdk-adapter.ts:1002` leaves the entire committed suite green. This is a real test gap, but not a shipping defect: the hook is present, the other generation seams continue to report, the affected catalog row is one of eight in the current local cache, and the consequence of this hypothetical regression is one missing advisory line. See OPTIONAL below.
- **The warning discriminator's false-positive direction is unpinned.** Removing the warning-shape check at `src/sdk-adapter.ts:599-603` also leaves the full suite green. HEAD is correct; the hypothetical consequence is one false informational line per process because the tier-request gate and latch remain. See OPTIONAL below.
- **The billing exclusion remains intact.** Mutating `isOpenAiOAuthRoute` (`src/sdk-adapter.ts:573-577`) to npm-only makes `tests/server-router.test.ts` fail at both API-key checks (`:291`, `:294`) and at the force-stream assertion (`:1116`). Making the `translateRequest` gate at `src/sdk-adapter.ts:527` unconditional also turns `tests/sdk-adapter.test.ts` red. `src/provider-factory.ts:164-212` still hard-codes the ChatGPT backend for OAuth and ignores the registry API base URL, so no tier-carrying OAuth route reaches the billable public API.
- **A wholly failed upstream turn produces no omission notice.** The request is still serialized and attempted without the tier, but when all attempts fail there is no completed step/result from which these hooks can surface the warning. That is defensible because no successful default-tier turn occurred; recorded here only so it is not mistaken for an unexamined path.

## Optional

All items below are non-blocking.

- **OPTIONAL — pin the collected-force-stream hook.** `src/sdk-adapter.ts:1002` is the one unpinned `reportUnsupportedServiceTier` wiring site; its OpenAI-format twin (`src/openai-adapter.ts:213`) and its sibling retry option (`src/sdk-adapter.ts:999`) are pinned. Extend the existing force-stream test at `tests/sdk-adapter.test.ts:713-765` so its `streamText` mock invokes `options.onStepFinish` with an `unsupported/serviceTier` warning and asserts the one-time notice.
- **OPTIONAL — add two one-field-at-a-time negative cases for the discriminator.** Current tests prove matching warnings fire, but do not separately prove: (1) a requested tier plus an unrelated `{type:'unsupported', feature:'other'}` warning stays silent, and (2) an `unsupported/serviceTier` warning with no requested tier stays silent. Both are needed: a supported-model run with an empty warning list would not catch removal of either individual guard. Put these beside the real-provider test at `tests/sdk-adapter.test.ts:319-362` or in the existing warning tests.
- **OPTIONAL — pin the OpenAI-format API-key boundary under an ambient tier.** There is already a non-OAuth test at `tests/openai-adapter.test.ts:154-167`, but it runs without `CLODEX_SERVICE_TIER`; therefore its `providerOptions` assertion remains green if the non-OAuth branch starts consulting the env. Adding that leak to `src/openai-adapter.ts:149-156` leaves the five relevant adapter/router/proxy files green. Under the `fast` environment already established by the sibling test at `tests/openai-adapter.test.ts:130-143`, also translate a non-OAuth request and assert `providerOptions?.openai?.serviceTier` is absent.
- **OPTIONAL — rename or strengthen `tests/cli.test.ts:201`.** The test name says it pins composition before the dry-run preview, but the observables only prove the final env value and that a preview printed. The preview at `src/cli.ts:970-987` contains no tier, so moving the assignment after preview generation leaves the test green. Rename it to the behavior it proves, or print/assert the resolved tier in the preview.
- **OPTIONAL — pin the endpoint-mode `tier=` trace diagnostic.** No test contains `tier=`. Removing the suffix at `src/proxy.ts:468` leaves the proxy/router suites green. This is trace-only and the request-shaping path is separately covered.
- **OPTIONAL — identify the model in the notice and consider bounded per-model dedup.** `src/sdk-adapter.ts:605` says "this model", while the single latch at `:581` means the first omitted model suppresses every later one in a long-lived process. The current README promise is only "warns once", so this is UX hardening rather than a contract defect.
- **NIT — the npm half of the translation gate is unpinned but unreachable.** Hoisting the merge at `src/sdk-adapter.ts:526` into a bare `if (options?.openAiOAuth)` leaves the relevant suites green because `tests/sdk-adapter.test.ts:284-288` varies npm and the OAuth option together. Every production caller computes `openAiOAuth` through `isOpenAiOAuthRoute`, which already requires `@ai-sdk/openai`, so this is test hygiene only.

## Checked and clean

- **Rebuild integrity.** `git range-diff 4ee1a49..ac29cb1 ad907a5..90bdb7a` maps the original feature patch exactly. The feature commit is 12 files, +264/−4; the two review-fix commits are 13 files, +434/−68; the complete PR is 15 files, +662/−36, with `git diff --check` clean and no `dist/`. One small non-tier hunk at `src/sdk-adapter.ts:475-484` correctly fixes our pre-existing billing-attribution comment (`cc_prev_req`, not a changing `cch`); it should not be described as part of the service-tier fix.
- **Previous optional items.** One-field route negatives, both gateway audit handlers, `--fast` environment assignment, invalid-value and warn-once behavior, shared route predicates, hostile-value handling, and public `CLODEX_SERVICE_TIER` documentation are present and behaviorally covered.
- **Dry-run invariant.** A built-CLI run with an empty `CLODEX_HOME` and dead ambient proxy — `clodex claude --fast --dry-run --proxy --save-mode` — exited 0, printed the preview, and left the home empty.
- **Validation counts.** `pnpm typecheck` is clean; the committed suite is 85 files / 1,382 tests; the focused seven-file suite is 205/205; the two warning files are 84/84. These match the PR body.
- **OAuth WebSocket transport.** The SDK's warning and `delete baseArgs.service_tier` happen during request argument construction, before any `fetch`, so the custom transport installed at `src/provider-factory.ts:188-201` does not alter this detector mechanism.
- **Modalities not run.** No live upstream/Codex request and no real `claude -p` smoke test were run. Patcher/binary-transform testing was not applicable because this PR does not touch patcher transforms or the Claude Code bundle.

PR #98, opened 2026-08-10
feat(auth): add multi-account OAuth selection

zachspartofaday:
## Summary

- store named ChatGPT OAuth account slots while retaining both the provider-default credential and its isolated catalog
- persist which account a provider launch uses, with explicit environment selection remaining process-scoped
- isolate model catalogs per account and fence credential/catalog transitions across refresh, switching, wrappers, and standalone servers
- contain process-scoped credential overrides to the affected provider without pairing them with a stored catalog or aborting unrelated providers
- materialize selected account storage in a downgrade-safe form so older readers continue to see the active credential
- preserve cleanup-journal, migration, locking, durability, and pricing behavior for every account-backed representation

This is one intentionally coherent auth-subsystem change. Splitting credential slots, selectors, account-scoped catalogs, and downgrade-safe `authRef` materialization would expose intermediate stored representations with missing readers or writers. It consolidates the fork work previously reviewed as zachspartofaday/clodex PRs #5, #17, and #31, adapted directly onto current upstream v2.4.0 (`ad907a54`) without the unrelated OpenCode/session-proxy work in #94.

## Reachability and upgrade path

The provider command, launch environment, wrapper, materializer, refresh paths, registry loader/writer, migration, pricing pass, and cleanup journal all consume the same account representation. Existing single-account providers retain their provider-default credential and catalog. Newer writers park that default while a named account is selected; compatibility materialization keeps the active credential visible to older readers without losing the canonical account slots.

A process-scoped `CLODEX_KEY_*` override is represented as blocked only for the affected provider. Unrelated providers remain launchable and visible to the server catalog; explicitly targeting the blocked provider retains the tailored `no isolated model catalog` diagnostic. The guard still covers slot-free/schema-v1 providers, so no override credential is paired with a catalog discovered under another credential.

## Review resolution

All still-applicable maintainer items are addressed in the rebuilt head:

- corrected downgrade-version wording without adding a sidecar compatibility mechanism
- parked/restored provider-default catalogs independently from named slots
- changed live-server wrapper override fences to bounded warn-and-continue behavior
- documented selectors, process overrides, snapshot behavior, and fail-closed cases
- removed plaintext OAuth account IDs from registry slots
- diagnosed lenient provider/cache drops
- made unchanged ordinary-registry fsync failures non-fatal while materialized selection writes remain fail-closed
- normalized selector case
- added a sequential real cleanup-journal default sign-in → named sign-in → selection test
- composed merged #99 pricing semantics: curated providers preserve top-level, parked-default, and named-slot caches; ordinary providers enrich all three classes

The account-switch test's obsolete assumption that the provider-default catalog disappears was replaced with an exact restoration assertion while retaining its credential/default/selection checks.

## Provenance

- base: `ad907a5436a13cc807e988ad5b4dff0437483d36`
- reconstructed feature commit: `c5d5eb6eeddae2c875a1e50529c877540ba48ed6`
- review-fix commit / HEAD: `b9d6465c5edff807d45c7ba59cc796c0bbcd7b28`
- complete diff: 44 files
- fork review: [zachspartofaday/clodex#41](https://github.com/zachspartofaday/clodex/pull/41), clean on exact HEAD and merged as `44547a4c1c88dfa582e3c46edff34daaefeb80bc`

PR #97 overlap is limited to the existing explicit-target CLI diagnostic branch and its direct test. No second resolver or service-tier behavior is included.

## Validation

Node 24.19.0 / pnpm 10.34.5 on exact HEAD `b9d6465c5edff807d45c7ba59cc796c0bbcd7b28`, rerun after the exact-HEAD clean fork review:

- focused proof/registry/CLI/server/wrapper suite — 10 files / 147 tests passed
- `pnpm typecheck` — passed
- `pnpm test` — 89 files / 1,576 tests passed
- `pnpm build` — passed
- `git diff --check ad907a5436a13cc807e988ad5b4dff0437483d36..HEAD` — passed

Loopback hosts bypassed the ambient outbound proxy during tests. No live provider request or real credential mutation was performed.

## Safety

Tests use isolated synthetic registries and credentials. The PR contains no prompts, credentials, headers, account identifiers, provider payloads, or raw traces. A provider/account switch fails closed when the selected credential or isolated catalog cannot be validated. No package, lock, dependency, release, generated `dist`, service-tier, usage-limit, endpoint, WebSocket, or OpenCode-generated change is included.


zachspartofaday:
RIP my 3 20x plans usage this weekend, thanks Tibo. 

-Zach

bman654:
I heard rumor a reset is coming today b/c everyone complained about the performative weekend reset right after everyone's regular reset.

Do you mind if I change your stacked PR's to target each other's branches instead of all targeting `main`?  It makes the diffs more readable and also makes it procedurally impossible to merge out of order.

bman654 (changes requested):
# Review — multi-account OAuth selection (#98)

We reviewed this in several independent passes, each taking a different angle (credential security; registry durability/locking/migration/downgrade; account-selection correctness and races; test soundness and mutation resistance; blast radius and wrapper/launch surface), with a separate pass set against every blocker or major raised, and then a final pass run specifically to attack our own conclusions. Everything below was executed against the PR head (`ff2821e`) and, where the claim is a regression, differentially against `4ee1a49`.

The core of this PR holds up unusually well. We attacked credential handling, concurrent registry writes, migration, downgrade, the selection state machine, and the tests themselves. No way was found to leak, lose, orphan, or cross a credential; the v5 "top-level cache == selected slot cache" invariant is genuinely maintained by every writer; 8-process concurrent mutation with a tight-loop reader produced zero torn state; and the tests are behavioural, not decorative — every load-bearing guard we deleted was killed by a whole-file run. Several scary-sounding concerns were run down and demoted with facts; they are listed at the bottom so nobody re-derives them. The comment quality on the new branches (especially `materialize.ts:131-144` explaining why the gate is keyed on a *stored* selection rather than on which selector won) is doing real work.

One thing is blocking: a pre-existing credential path (`CLODEX_KEY_<PROVIDER>`) that base honours becomes a hard failure of the entire catalog load, with cross-provider reach. Everything else here is optional.

A later pass overturned part of our own framing of that ask, corrected a misquoted transcript, and found one place where we had dismissed a concern too readily. Those corrections are applied below; our original wording for the blocking ask is not what is being asked for.

---

## Blocking

### 1. `src/registry/load.ts:24-44` — one `CLODEX_KEY_*` override aborts `loadRegistryProviders()` for the whole process

The `Promise.all` over every provider starts at `src/registry/load.ts:24`; the new guard at lines 38-44 throws out of it rather than being applied to the affected provider. The result is that an override for a provider the user is not launching takes down the command.

Executed differentially — both trees freshly built, `env -i`, dead ambient proxy, isolated `CLODEX_HOME`, a `schemaVersion: 1` registry with the two supported providers (`openai` via `env:OPENAI_API_KEY`, `openai-oauth` via `env:OTHER_KEY`), launching **openai** explicitly:

```
$ node <tree>/dist/cli.js claude --endpoint --dry-run --provider openai --model gpt-5.6

[base4ee  no override]                 exit=0  DRY RUN — would execute: …
[base4ee  CLODEX_KEY_OPENAI_OAUTH set] exit=0  DRY RUN — would execute: …
[base4ee  CLODEX_KEY_OPENAI set]       exit=0  DRY RUN — would execute: …
[head     no override]                 exit=0  DRY RUN — would execute: …
[head     CLODEX_KEY_OPENAI_OAUTH set] exit=1  … no isolated model catalog for provider "openai-oauth"
[head     CLODEX_KEY_OPENAI set]       exit=1  … no isolated model catalog for provider "openai"
```

Same registry, `clodex server --endpoint --no-discovery`:

```
[base4ee  no override / either override] Loaded 2 models (2 from registry providers)   # server runs
[head     no override]                   Loaded 2 models (2 from registry providers)
[head     CLODEX_KEY_OPENAI_OAUTH set]   exit=1  Unexpected error: … for provider "openai-oauth"
[head     CLODEX_KEY_OPENAI set]         exit=1  Unexpected error: … for provider "openai"
```

Two things this shows beyond the cross-provider collateral: at head `CLODEX_KEY_*` also hard-fails the provider it *names*, so it ceases to be a usable launch override for any enabled provider (that total removal may be intended, but the PR body only says a switch "fails closed"); and `clodex server` surfaces a deliberate user-facing error as an unhandled `Unexpected error:` crash with a stack trace, where `clodex claude` catches it and prints one red line (`src/cli.ts:1138-1153` vs the top-level handler at `src/cli.ts:1571`).

Reachability, stated honestly: `CLODEX_KEY_*` is undocumented — clodex never writes it, and it appears in no README, changelog, docs page, or help text. But it is implemented in `src/env.ts:284` (`clodexKeyEnvVar`) and resolved ahead of `authRef` in `resolveProviderCredentialWithSource` (`src/env.ts:2314-2319`), and it is pinned by base tests this PR keeps green — `tests/env.test.ts:111` ("resolves CLODEX_KEY_\* env before authRef"). The two halves of the tree now disagree: the refresh path deliberately *skips* on the same state (`tests/provider-auth.test.ts:600` pins `reason: 'CLODEX_KEY_OPENAI_OAUTH is a process-scoped provider credential override.'`), while `load.ts` treats it as fatal. And this PR's own `tests/claude-wrapper.test.ts:49-53` now strips every `CLODEX_KEY_*` from the spawned test env — an implicit acknowledgement that developer machines carry these.

**The ask** — and note this is deliberately behaviour-level, not a prescribed diff:

> Keep the safety property: never pair a `CLODEX_KEY_*` credential with a catalog discovered for a different credential, including schema-v1 / API-key providers. Scope the failure to the affected provider so an override cannot abort unrelated providers or whole-catalog commands. Preserve a structured blocked reason (or perform a target-aware preflight) so selecting that provider still reports the existing tailored "no isolated model catalog" error. Do not restore base's unsafe override-plus-stored-catalog pairing.

Two constraints on the fix that a later pass established, and that an earlier pass of our review got wrong:

- **A slot-free `schemaVersion: 1` provider is not exempt.** Its single stored cache is still a catalog discovered under one credential, and a process override can represent a different project or entitlement set. The PR tests exactly this at `tests/registry-load-credential-source.test.ts:93-118` ("does not pair an API-key provider override with its persisted catalog"), and the schema-v1 executions above reach that guard. Whatever the fix, do not go back to pairing the override with the stored catalog.
- **"Hard-fail only when the overridden provider is the one being launched" is mislocated as a purely local `load.ts` change.** The loader runs before endpoint-mode provider selection and loads the complete catalog; in interactive mode there is no launch target yet. Simply dropping the provider also turns a precise safety diagnostic into a generic "provider/model not found" unless the blocked reason is carried to the selection layer — you already render exactly the right message there (`src/provider-catalog.ts:338-344`, `launch blocked — no isolated model catalog; stored auth: …`), so representing blocked entries rather than deleting them is one clean shape. A target-aware preflight is another. Your call.

---

## Optional

Everything in this section is **non-blocking**.

**1. `src/registry/types.ts:7-16` states a downgrade guarantee that is false for every release before 1.3.0.** The comment says *"Older builds fail closed on an unknown schema version in every MUTATING path (parseRegistryStrict throws)"*, and the PR body inherits it. `parseRegistryStrict` first appears in **v1.3.0**: `git show <tag>:src/registry/io.ts | grep parseRegistryStrict` is empty for every tag from `v0.1.0` through `v1.2.2` and first matches at `v1.3.0` (tag commit dates: `v0.1.0` 2026-07-20T01:12-05:00, `v1.2.2` 2026-07-23T09:37-05:00, `v1.3.0` 2026-07-23T21:58-05:00). A later pass independently bundled the v1.2.2 registry code from tag source and ran its real `toggleProviderEnabled` against a head-shaped v5 registry: `before slots=["work","personal"] active=work` → `after slots=[] enabled=false`. Suggested wording: *"builds ≥ 1.3.0 fail closed …; releases before 1.3.0 (0.1.0–1.2.2) have no strict loader and will silently strip slot state."* See "Checked and clean" #1 for why the severity is low and why the sidecar-mirror idea should be rejected.

**2. `src/registry/crud.ts:147-166` — the provider-default catalog is not parked on the way out, so only the default identity cannot be returned to offline.** The parking branch at line 153 requires `previous` (the outgoing `activeAuthAccount`), which is `undefined` when the current identity is the provider default; lines 163-165 then `delete provider.modelsCache`. Executed against a real on-disk v5 registry:

```
start        | top: ['default-model'] | work slot: ['work-model'] | default-model anywhere: true
after ->work | top: ['work-model']    | work slot: ['work-model'] | default-model anywhere: false
after ->dflt | top: NONE              | work slot: ['work-model'] | catalog after round trip: (EMPTY)
at work -> at alt -> back at work     | catalog back at work: openai-oauth(1)   # named slots round-trip fine
```

Exposure is switch-only: `upsertOAuthAccountSlot` (`src/registry/provider-auth.ts:136-142`) drops the top cache only when re-authing the already-selected slot, so merely adding an account is safe. Suggested fix: a `defaultModelsCache` alongside `defaultAuthRef`, mirroring the parked-default idea you already have for the credential. Severity is low because the same `(EMPTY)` terminal state is reached by the design's ordinary intended path — first switch to a freshly added account — and `if (models.length === 0) return null` plus "No providers available." are pre-existing base behaviour (`base4ee/src/registry/materialize.ts:129`, `base4ee/src/cli.ts:1159`).

**3. `src/claude-wrapper.ts:53-61` matches a strictly broader set than the guard it mirrors.** `activeProviderCredentialOverride` accepts any `/^CLODEX_KEY_[A-Z0-9_]+$/`, while the load-path guard fires only for `clodexKeyEnvVar(<enabled provider id>)`. Real built `dist/claude-wrapper.js`, real loopback listener, real `server-runtime.json` record, fake claude that prints on exec:

```
[base] CLODEX_KEY_TOTALLY_UNRELATED  exit=0 launched=true
[head] CLODEX_KEY_TOTALLY_UNRELATED  exit=1 launched=false
[head] CLODEX_KEY_ blank ("   ")     exit=0 launched=true
```

`CLODEX_KEY_TOTALLY_UNRELATED` maps to no provider in any registry, yet claude never launches. Same site: `Object.keys(env).sort()` runs on every wrapper launch (hot path for every spawned agent) and then reports whichever key sorts first, which need not be the relevant one.

**4. Related wrapper corollary, found on a later pass: the `CLODEX_OAUTH_ACCOUNT` fence has a reproducible no-op false positive.** `src/claude-wrapper.ts:146` fires on any non-empty `CLODEX_OAUTH_ACCOUNT` whenever a server is live, without consulting the registry. But `applySelectedOAuthAccount` deliberately **ignores** the selector when the provider has no slots (`src/registry/materialize.ts:145-152`). So on a slot-free registry the variable has no semantic effect and there is no alternate account the running server could have snapshotted — yet the wrapper still kills the process:

```
slotless CLI selector:     exit=0 dry-run=true error=none
slotless wrapper selector: exit=1 launched=false
  error=clodex-claude: CLODEX_OAUTH_ACCOUNT cannot override the credential snapshot …
```

This is not a compatibility regression (the variable is new in this PR), which is why it is optional. A warn-and-exec policy at both fences would resolve this and item 3 at once, without the wrapper having to import registry code — which matters given the "keep the wrapper tiny, minimal imports" constraint.

**5. Docs and PR body don't mention the new failure modes.** `git diff --name-only 4ee1a49 HEAD` contains no `docs/` file. `docs/background-agents.md` gains failure modes it does not describe (zero occurrences of `CLODEX_OAUTH_ACCOUNT` or `CLODEX_KEY` in that file, yet both wrapper fences can now exit non-zero), and the PR body does not state that `CLODEX_KEY_*` becomes fatal on the load path. A short note in both is worth more than the code change in item 3.

**6. `src/registry/types.ts:98` / `src/registry/provider-auth.ts:132` — the ChatGPT account UUID moves from keychain-only to plaintext `providers.json`.** It is an identifier, not a bearer credential, and the file is mode `0600` inside a `0700` dir — but `providers.json` is what users back up, sync, and paste into bug reports. Cost to remove looks low: the on-disk copy is carried into the credential snapshot at `src/registry/refresh-credentials.ts:84-90` and consumed by exactly one thing, the `selectedAccount` generation fence at `src/registry/refresh-models.ts:408`; `selectedAccount` already carries `addedAt`, re-stamped on every `upsertOAuthAccountSlot` (`provider-auth.ts:131`), so that fence already detects a slot re-sign-in without the UUID. (`credentialInstanceAuthRef` is deterministic — `src/credential-helper.ts:66` — so `authRef` alone cannot fence a re-auth; `addedAt` can.) Drop the field or hash it.

**7. `src/registry/io.ts:277-284` — a v5 provider whose top-level cache drifts from the selected slot's is dropped *silently* by the lenient loader**, while the strict path names the problem:

```
lenient providers: []
strict THREW: Provider registry contains an invalid provider entry.
```

`clodex claude` then says "no providers" rather than "your registry is inconsistent". The asymmetry is the issue, not the strictness — we traced every cache writer and could not construct this state from clodex's own writes (see "Checked and clean" #6), so it needs external corruption or a hand edit. Consider having the lenient path emit the same diagnostic when it drops a provider for selection-storage reasons.

**8. `src/registry/io.ts:429-449` — `loadRegistry` can now throw on a pure read path; base never could.** Head fsyncs the parent directory on every load and rethrows anything not `EINVAL`/`ENOTSUP`/`EPERM`, so an `EMFILE` under agent fan-out or an `EACCES` on `~/.clodex` makes `clodex claude` die at load instead of degrading. The behaviour is deliberate and the comments explain why; measured cost is irrelevant (0.076 → 0.661 ms/call on APFS). Narrow suggestion: on the **no-migration** early-return path (`io.ts:435-447`), consider treating an fsync failure as non-fatal unless `hasMaterializedActiveAccount(registry)` — that branch computes the predicate one line later anyway (`io.ts:438`), and the identity-splitting risk that justifies throwing exists only for materialized registries.

**9. `CLODEX_OAUTH_ACCOUNT` ergonomics, two small things.** (a) `src/registry/materialize.ts:153-165`: the selector is process-global but validated per provider, so with two slot-bearing OAuth providers no value can satisfy both (`CLODEX_OAUTH_ACCOUNT=work: provider "openai-oauth" has no account named "work" (available: personal)`). Not constructible today — `src/provider-templates.ts` ships exactly one OAuth-capable template — so this is a forward hazard worth a line in the variable's doc comment, or a per-provider spelling before a second OAuth template lands. (b) `validateOAuthAccountName` (`src/registry/provider-auth.ts:56-57`) lowercases, while `applySelectedOAuthAccount` only `trim()`s, so `--account "Work"` stores `work` and `CLODEX_OAUTH_ACCOUNT=Work` then throws. The error does list available names and the success message echoes the normalized slot, so it is discoverable — but `toLowerCase()` here costs nothing.

**10. Test notes — no action required.**
- Strongest coverage gap found: `isManagedCredentialAccount`'s slot-account branch is pinned by a single direct unit test; neither `tests/provider-auth.test.ts` nor `tests/registry-crud.test.ts` notices a mutation there because both mock `credential-lifecycle`. One end-to-end test running the real journal through an `--account` sign-in would close it.
- `src/registry/add-template.ts:155` and `:186` — the new OAuth-state branch of `existingProviderError` is unexercised at the post-probe and commit-lock call sites (selectively disabling the branch at those two sites leaves `tests/registry-add-template.test.ts` 20/20 green; the control, deleting the *call*, is killed). Non-actionable: the guard is unreachable in production at all three sites, because the sole production caller `src/providers-command.ts:581` passes no `opts` and `replaceExisting` is test-only.
- `src/registry/io.ts:412-415` — the `schemaVersion === 5` conjunct in `hasMaterializedActiveAccount` is an equivalent mutant (removing it leaves the suite green). Recorded so nobody later reads it as a tested invariant.
- `tests/claude-wrapper.test.ts` gains three more spawn+loopback tests (`:319`, `:334`, `:350`) on the suite's most load-sensitive mechanism. Green here; recorded so a future red CI on those names is not chased as a regression in the account code.
- Harness caution for this repo: do not write a vitest test issuing **concurrent** `provisionProviderCredential` calls — `vi.mock('@napi-rs/keyring')` is not reliably applied to a concurrently-resolving dynamic `import()`, and the losing call falls through to the real macOS keychain. In production these are serialized by `withProviderMutationLock`, so the interleaving is not reachable anyway.

**11. Cross-PR composition (a heads-up, not a finding against this PR).** This branch was reviewed on its own, not in an integration context. PR #99 overlaps in `src/provider-catalog.ts`, `src/registry/{add-template,io,materialize,pricing,refresh-models,types}.ts`, and `tests/pricing.test.ts`, and a real three-way probe is not clean:

```
git merge-tree --write-tree HEAD refs/review/pr99
exit=1
CONFLICT (content): Merge conflict in src/registry/pricing.ts
CONFLICT (content): Merge conflict in tests/pricing.test.ts
```

Whichever lands second, the resolution wants a composed behavioural test proving that #99's `preserveModelPricing` skips both the selected top-level cache and parked slot caches while ordinary providers still enrich both. PRs #96 and #97 have no file overlap.

---

**1. "A downgrade silently destroys every OAuth account slot" — mechanism real, severity is not.** Reproduced twice independently (transcript in optional item 1). But the reachability argument does not survive: the entire pre-fence era is the package's first four days (`v0.1.0` 2026-07-20T01:12-05:00 → `v1.3.0` 2026-07-23T21:58-05:00), and `@latest`, `@1`, `@^1.0.0`, `@1.x` and `npx` all resolve to a fenced build today. Harm is bounded and recoverable: keychain entries are not deleted (the cleanup journal only fires on provider removal), post-strip `authRef` still names the slot the user had selected — so the next launch uses the *same identity as before the downgrade*, no wrong-account request, no billing surprise — and recovery is `providers auth --account <name>`. No code change in this PR can fix it. **The sidecar mirror (`~/.clodex/oauth-accounts.json` + repair logic) is not proportionate — reject it.** The only genuinely wrong thing is the comment, which is optional item 1.

**2. "An orphaned v3/v4 selector makes `toggleProviderEnabled` throw uncaught" — real, but pre-existing on base main and needs bytes no shipped build can write.** Executed on base 2.3.0 with a lenient-loadable / strict-rejectable registry, toggling a *healthy* provider: `BASE toggle THREW Error Provider registry contains an invalid provider entry.` Same call chain, same missing try/catch, on code this PR does not touch. Also: `git log --all -S "activeAuthAccount" -- src` returns only `ff2821e`, and base `types.ts` defines `REGISTRY_SCHEMA_VERSION = 1` — no released `@bman654/clodex` has ever written a registry above schemaVersion 1, so the state requires an unmerged fork build *plus* a hand edit. And the menu is not wedged: `removeProviderFromRegistry` succeeds on the orphan, as do Browse models and Back, and `setActiveOAuthAccount(id, undefined)` repairs it. `hasValidSelectionStorage` runs before any byte is written, so the failed toggle leaves the registry byte-identical. If tracked at all, it belongs against base main.

**3. The wrapper's documented fail-open contract is intact.** `src/claude-wrapper.ts:19` and `docs/background-agents.md:12,34` both scope their promise to *no live server* ("no live server: env untouched — `claude` always launches normally"); both new fences are guarded by `state &&` (`claude-wrapper.ts:146` and `:157`) and can only fire when a server **is** live. Confirmed by executing both built wrappers with no runtime record:

```
base4ee no-server + key override: exit=0 output=WRAPPER-LAUNCHED
head    no-server + key override: exit=0 output=WRAPPER-LAUNCHED
base4ee no-server CLODEX_REQUIRE_SERVER: exit=1  (base already lets a clodex variable stop the wrapper)
```

`--check` also still exits at `claude-wrapper.ts:141` before both fences. One qualification, from a later pass: this does **not** mean every live-server `CLODEX_OAUTH_ACCOUNT` refusal is semantically necessary — the slot-free case is a reproducible no-op false positive, which is optional item 4.

**4. No credential leak, loss, orphan, or cross-account mix.** Canary access-token / refresh-token / account-id values planted, then the real `resolveActiveAccount`, `applySelectedOAuthAccount`, `projectSelectedOAuthAccount`, `refreshCredentialSnapshot`, `formatRegistryAuthLabel`, `resolveProvidersForDisplay`, `accountSwitchOutcome`/`Hint`/`ServerRestartWarning` were driven and a full walk of `CLODEX_HOME` performed: zero secret/refresh hits on every surface; the only account-id hits are optional item 6. The new `ProviderCredentialOverrideState.fingerprint` (sha256) is compared with `isDeepStrictEqual` and never rendered — only `.variable` reaches user-visible text. Chunking/tombstones behave for slot accounts (4000-char credential → 7 chunks, exact round-trip, zero dangling chunks on delete; sibling slots byte-intact; an injected mid-chunk write failure fails closed; delete-then-re-provision clears the tombstone). Real `removeProviderFromRegistry` over a parked default + two chunked slots: `readable credential residue = 0 | dangling chunks = 0`, journal empty. The cleanup journal can neither delete a live credential nor orphan a new one (`journalCredentialWrite` precedes the write; both crash windows covered). And no switch can cross credentials: real launch materialization returned the correct distinct token in every case, with a vanished slot throwing rather than silently launching as the provider default.

**5. Concurrency, locking, and migration hold under real multi-process load.** 8 processes × 30 iterations of `setActiveOAuthAccount` (cycling work/personal/cleared) + `toggleProviderEnabled`, with a 9th process doing raw read + parse + invariant checks in a tight loop: `workers failed: 0`, `reader: {"reads":231828,"problems":[]}`, no stray `.lock`/`.tmp`/`.reap`. Zero torn JSON, `authAccounts` never dropped below `{work, personal}`, all four v5 invariants held on every observation. The live-PID lock rule survives the new writers (a contender that back-dated the lock mtime by an hour still got `Timed out … held by clodex process (pid …)`), and `lock.ts` is byte-unchanged from base. v3/v4 → v5 migration is idempotent and concurrency-safe: 10 processes × 40 loads → 400 identical materialized results, final file a single well-formed v5 with the unprovenanced top cache correctly dropped. Nested `withProviderMutationLock → withRegistryWriteLock` chains are consistently ordered and do not self-deadlock.

**6. The v5 "top-level cache == selected slot cache" invariant is genuinely maintained by every writer.** `updateProviderCache` assigns the *same object* to both and an env-override refresh writes only the slot; `applyPricingToRegistryProviders` enriches both through the identical deterministic path; `setActiveOAuthAccount` parks/adopts by reference; `upsertOAuthAccountSlot`/`upsertOAuthProvider` drop the top cache exactly when they replace the credential that owned it. This is the analogue of the `preserveModelPricing` defect and it is handled. The one route that could have produced an unsaveable state — a provider holding `defaultAuthRef` whose `authType` flips to `api` — is explicitly closed by `src/registry/add-template.ts:40-66`.

**7. Selection correctness under races.** A stale `refreshCredentialSnapshot` cannot commit across an account switch (`{"ok":false,"reason":"Provider account selection changed while models were refreshing."}`, top cache unchanged). Two concurrent refreshes of different accounts of one provider do not cross-contaminate (`top: [new-work-model], work: [new-work-model], alt: [new-alt-model]`). The `:account:` collapse in `oauthProviderIdFromAccount` does not merge credential caches — the cache key is `oauthCredentialStateKey(providerId, authRef)` and `authRef` carries the full account string; the collapse only picks the refresh endpoint. No `Date.now()` recency hazard: `addedAt` is compared for equality as a generation marker, never ordered, and the same-millisecond re-auth gap is closed by the provider mutation lock held across both writers.

**8. Zero migration friction for the single-account majority, and no forced schema bump.** A pre-PR `schemaVersion: 1` registry loads identically on base and head, and `clodex claude --dry-run` output is byte-identical between worktrees. `loadRegistry(); saveRegistry(reg)` on a v1 OAuth registry writes back `schemaVersion: 1` with the same key set on both trees — the version ladder only escalates when slot/selector/parked-default state is actually present. `oauthProviderIdFromAccount` is a no-op for slot-less providers, and `readEnvCredential`'s new `env` parameter defaults to `process.env` with no production caller passing anything else.

**9. The launch/proxy invariants this repo cares about are untouched.** Exhaustive scan of every changed `src/` file for `process.env` writes: zero — `buildChildEnv`/`buildHttpProxyChildEnv` are not in the diff, so the self-loop guard is intact. `execIntoClaude` is byte-identical to base (still `process.execve`, still the pre-syscall re-check, still the spawn fallback). Neither `src/http-proxy/` nor `src/proxy.ts` appears in `git diff --name-only 4ee1a49 HEAD`, so the unrewritten-body / response-model-echo invariant is out of this PR's reach. The server genuinely snapshots one account identity at startup and `src/server/router.ts:123` resolves the per-request credential from the already-account-projected `model.authRef`, so a request cannot be served with one account's credential and another's catalog.

**10. The tests are not theatre.** Across all four passes, every load-bearing guard we deleted was killed by whole-file runs (never `-t`), with `src/` restored from pre-mutation snapshots: fail-closed switching, downgrade materialization, the orphaned-selector refusals, cleanup-journal integrity, the credential-reference set, the refresh generation fences, the crash-recovery durability re-syncs, the v5 storage invariants, and the wrapper fences. One pass completed 22 mutations with 21 killed and the survivor an equivalent mutant. The 288-case `authType × enabled × slots × stored × CLODEX_OAUTH_ACCOUNT × CLODEX_KEY_*` grid really executes both the runtime and display resolvers and requires them to agree *including when the runtime throws*; a separate 480-cell differential found 0 drift between `resolveActiveAccount` and `applySelectedOAuthAccount`. No mutation was caught by a source-digest pin. Also checked and absent: the "asserts a console string instead of behaviour" shape, the "hand-written expected value" shape, and the staging trap.

**11. Suite state.** `88 files / 1512 tests passed` at head under a dead ambient proxy and an isolated `CLODEX_HOME` — reproduced here for this write-up and independently in three passes, matching the PR body's claim. The one full-suite failure a fourth pass saw (`external credential helper > force-kills a helper that exceeds the runtime limit`) is the sole failure on base under identical env — pre-existing load sensitivity, not a regression.

**12. Residual validation gaps (not findings).** Nobody executed the supported Node 22 floor, Windows credential-manager behaviour, or a live OpenAI device-code sign-in followed by an account switch and a real request; the PR body is explicit that no live provider request or real credential mutation was performed. The synthetic credential and chunking evidence is otherwise strong, so these are noted rather than asked for — but a manual `clodex claude -p` smoke on both accounts before merge would be cheap insurance.

zachspartofaday:
Addressed the blocking item and every still-applicable optional suggestion in the current-upstream reconstruction. I first rebuilt the feature directly on v2.4.0 (`ad907a54`) so merged #99's pricing behavior composes rather than losing either side of the conflict.

### Blocking provider-override isolation

The loader now returns a blocked entry only for the affected provider and carries its tailored reason through selection. Unrelated providers remain available to explicit CLI launch and the server catalog; explicitly targeting the blocked provider still reports `no isolated model catalog` rather than a generic not-found error. The schema-v1/API-key case remains fail-closed—no process override is paired with a stored catalog.

Differential coverage now spans loader, explicit launch, and server catalog, including one blocked plus one healthy provider.

### Optional items

1. **Downgrade wording:** corrected the contract to distinguish releases ≥1.3.0 from 0.1.0–1.2.2; no disproportionate sidecar was added.
2. **Provider-default catalog parking:** added `defaultModelsCache`; default and named identities now round-trip their own catalogs independently.
3–4. **Wrapper override breadth / slot-free selector:** live-server wrapper fences now emit a bounded warning and continue instead of refusing launch; no registry import was added to the wrapper.
5. **Docs:** documented `CLODEX_OAUTH_ACCOUNT`, process-scoped provider overrides, snapshot behavior, warning behavior, and fail-closed cases in `docs/background-agents.md`.
6. **Plaintext account UUID:** registry slots no longer persist the OAuth account identifier; runtime credential payloads retain what the provider protocol requires, while generation fencing uses retained non-secret state.
7. **Lenient drops:** invalid providers/caches dropped by the lenient loader now emit bounded diagnostics while strict loading remains strict.
8. **Read-path fsync:** an unchanged ordinary registry diagnoses and tolerates parent-directory fsync failure; migrated/materialized account state remains fail-closed.
9. **Selector ergonomics:** selector matching is consistently trim/lowercase normalized. The documented process-global limitation remains explicit before any second OAuth template exists.
10. **End-to-end journal coverage:** added a sequential real-journal default sign-in → named sign-in → named selection test without concurrent credential provisioning. The unreachable/test-only add-template notes remain unchanged.
11. **#99 composition:** `preserveModelPricing` now protects curated top-level, parked-default, and every named-slot cache; ordinary providers enrich all three cache classes. A single composed behavioral test fails if either conflict side wins wholesale.

The account-switch test's obsolete assumption that the default catalog disappears was replaced with an exact parked-default restoration assertion; its credential, default-selection, and account-selection assertions remain intact.

Fork review: [zachspartofaday/clodex#41](https://github.com/zachspartofaday/clodex/pull/41) reached a clean result on exact `b9d6465c5edff807d45c7ba59cc796c0bbcd7b28` and merged as `44547a4c1c88dfa582e3c46edff34daaefeb80bc`.

Validation on Node 24.19.0 / pnpm 10.34.5 at `b9d6465c5edff807d45c7ba59cc796c0bbcd7b28`, rerun after that exact-HEAD clean result:

- focused proof/registry/CLI/server/wrapper suite: 10 files / 147 tests passed
- full suite: 89 files / 1,576 tests passed
- typecheck, build, and diff-check passed
- synthetic fixtures and isolated homes only; no live provider request or real credential mutation

The upstream branch and PR body now reflect this exact reviewed head.

bman654 (approved):
# Review — multi-account OAuth selection (#98), re-review at `b9d6465`

**Approving.** The one blocking ask from the previous round is addressed, and it is addressed at the right level — the failure is scoped to the affected provider, the safety property is preserved rather than traded away, and the tailored diagnostic still reaches an explicit launch targeting that provider. Rebuilding the feature directly on `ad907a5` was the right call; the #99 pricing composition is real and is pinned by a test that fails if either side of the old conflict wins wholesale.

Any follow-up request below is optional. Nothing blocks merge.

## Prior asks

| Ask (2026-08-10 round) | Status | How it was verified |
|---|---|---|
| **BLOCKING** — scope the `CLODEX_KEY_*` failure to the affected provider, preserve a structured blocked reason, do not restore base's override-plus-stored-catalog pairing | **ADDRESSED** | Executed differentially at head and base; loader/consumer both mutation-killed; a six-shape pairing attack finds no route |
| Optional 1 — downgrade wording ≥1.3.0 vs 0.1.0–1.2.2 | Addressed | Wording corrected; no sidecar added |
| Optional 2 — park the provider-default catalog | Addressed | `defaultModelsCache` (`src/registry/oauth-account-storage.ts:36-41`); deleting the parking assignment at `:40` fails 3 tests across `tests/registry-crud.test.ts` and `tests/account-switch-refresh.test.ts:173` |
| Optional 3–4 — wrapper override breadth / slot-free selector | Addressed | Real built `dist/claude-wrapper.js` spawned against a live loopback listener and a real `server-runtime.json`: both fences now warn and launch (exit 0, claude executed), neither prints a credential value |
| Optional 5 — docs | Addressed | `docs/background-agents.md:16,102` |
| Optional 6 — plaintext account UUID | Addressed | `oauthAccountId` gone from `src/registry/types.ts`; parser strips a hand-injected one and a save round trip removes it from disk; the credential *payload* still carries `accountId` (`tests/provider-auth.test.ts:223-227`) |
| Optional 7 — lenient-drop diagnostics | Addressed (see O2) | Strings pinned; delivery wiring is not — coverage note only |
| Optional 8 — read-path fsync | Addressed (see O3, O4) | Tolerate half pinned; fail-closed half is real but unpinned — coverage note only |
| Optional 9 — selector normalization | Addressed | `src/registry/materialize.ts:122`; removing `.toLowerCase()` fails `tests/oauth-account-selection.test.ts` twice |
| Optional 10 — end-to-end journal coverage | Addressed | `tests/credential-cleanup-journal.test.ts:115` does not mock `credential-lifecycle.js`; deleting the journal write in `src/registry/provider-auth.ts` fails 6 tests, so its `toEqual([])` assertions are not vacuous |
| Optional 11 — #99 composition | Addressed | `src/registry/pricing.ts:224-255`; reconstructing either wholesale side of the old conflict is killed by `tests/pricing.test.ts:98` |

## The blocking ask — verified by execution

Re-checked here on a fresh `pnpm build`, `env -i`, isolated `CLODEX_HOME`, dead ambient proxy, `schemaVersion: 1` registry with `openai` (`env:OPENAI_API_KEY`) and `openai-oauth` (`env:OTHER_KEY`), each provider carrying its own stored `modelsCache`:

```
$ CLODEX_KEY_OPENAI_OAUTH=override-token clodex claude --endpoint --dry-run \
      --provider openai --model gpt-5.6
exit=0   (dry run complete — Claude Code was NOT launched)

$ CLODEX_KEY_OPENAI_OAUTH=override-token clodex claude --endpoint --dry-run \
      --provider openai-oauth --model gpt-5.6-codex
exit=1   ■ CLODEX_KEY_OPENAI_OAUTH is a process-scoped credential with no isolated model
           catalog for provider "openai-oauth". Save that credential as a provider or
           account and refresh its models, or unset the variable.
```

Both halves of the ask hold: the unrelated provider is untouched, and the blocked provider still gets the tailored reason rather than a generic not-found. `src/registry/load.ts:47-54` now records the reason and returns instead of throwing out of the `Promise.all`, and `src/cli.ts:1158-1160` carries it to *both* exits — the all-blocked branch at `:1161-1165` and the target-missing branch at `:1189-1197`. `clodex server` no longer dies with `Unexpected error:` + stack trace.

The safety property was attacked, not assumed. A blocked provider never reaches `keys`, so `src/registry/materialize.ts:239` (`if (!apiKey.trim() && !anonymous) return null;`) drops it from the catalog, and every launch-time credential resolver (`resolveLocalProviderApiKey` at `src/provider-catalog.ts:36-50`, `localModelToRoute`'s refresh, `resolveModelApiKey` in `src/server/router.ts`) keys off a provider already in the catalog. A six-shape harness — schema-v1 API key, oauth, explicit anonymous, `authType:'none'` with a real authRef, legacy `custom-openai` with `authType` undefined, and oauth with `CLODEX_OAUTH_ACCOUNT` — produced **zero** override/stored-catalog pairings at head against **four** at `ad907a5`. Over a real socket with `openai-oauth` blocked and aliased as `codex`: `GET /openai/v1/models` omits it, both `codex` and its canonical id 400, and the healthy provider's alias reaches upstream. Proxy bridge mode drops the blocked provider's favorite and alias from the route table.

Mutation checks on the fix itself: deleting `blockedProviders.set(...)`, restoring the original `throw`, deleting the `materializeOne` keyless gate, forcing `blockedLaunchReason = undefined`, and neutering the wrapper fence are each killed by whole-file runs (4, 4, 6, 1 and 2 failures respectively). Both seams are individually pinned; there is no single end-to-end test spanning loader → catalog → CLI (`tests/cli.test.ts:280` and `tests/server-index.test.ts:214` hand-construct the map), which is hygiene rather than a gap given both halves die separately.

## Raised and then cleared — stated so nobody re-derives them

- **"Both new diagnostic sinks are entirely unpinned" was raised as MAJOR and demoted.** The mutation survivors are real (see O2), but the framing was wrong: both *producers* are directly pinned — `tests/registry.test.ts:143` drives `loadRegistry` against a real malformed v5 registry and asserts the drop diagnostic, `tests/oauth-account-selection.test.ts:108` asserts the slot-free warning. What is unpinned is only whether `fetchProviderCatalog` forwards the callbacks. Under the mutant, the lenient loader still drops the inconsistent provider and the slot-free selector is still a no-op; only the explanation is lost.
- **"The read-path fail-closed guarantee is untested" was raised as MAJOR and demoted.** Deleting `src/registry/io.ts:475-477` does survive the full 1,576-test suite (see O3), but it is not an identity-safety guard on that branch: `loadRegistry` reaches `:467-481` only when `migration.changed === false`, a v5 provider is admitted only if `hasValidSelectionStorage` (`src/registry/io.ts:265-303`) already proved top-level `authRef`/`modelsCache` equal the selected slot's, and publication is a whole-file temp+rename. The mutant continues with bytes that are already credential/catalog coherent. The actual write and locked-migration paths remain fail-closed (`src/registry/io.ts:533-568,627-700`), with the durability cases beginning at `tests/registry-io-durability.test.ts:279,325`. Coverage note, not a defect.
- **The account-switch assertion change is stronger, not weaker.** Replacing `expect(persisted.modelsCache).toBeUndefined()` with an exact `['default-only-model']` restoration plus `defaultModelsCache` deletion additionally proves the parked field was consumed; the credential, default-selection and account-selection assertions are intact (`tests/account-switch-refresh.test.ts:173-186`).
- **Two one-off full-suite failures were not attributed to #98.** `tests/credential-helper.test.ts:284` (`ENOENT … credentials.json.helper-pid`) was independently observed on base under the identical dead-proxy/isolated-home environment and passed in isolation and on repeats. `tests/http-proxy-index.test.ts:39` timed out once under load, then passed in isolation and on repeats; the file is outside this PR's diff, but the timeout was not reproduced on base, so it is not labeled proven pre-existing. Neither failure recurred in this audit's full head or base runs.
- Nothing overturned the previous round's cleared items: no credential leak on any new surface (canary scan over the blocked-reason map, the loader diag/warn channels, `ProviderCredentialOverrideState`, the lenient-drop diagnostic, and the wrapper warnings — all name variables, never values), and the refresh path still refuses to persist a catalog discovered under an override (`src/registry/refresh-models.ts` skip + `{ ignoreProviderOverride: true }` on the sign-in refresh).
- **Not executed:** the Node 22 floor, Windows credential-manager behavior, a live two-account OpenAI device-code sign-in followed by a real request, and the manual two-model `clodex claude -p` smoke. The synthetic credential, loopback, mutation, differential, typecheck, build, and Node 24 coverage are otherwise strong; these unrun modalities do not change the verdict.

## Optional

All non-blocking. Take any, all, or none.

**O1 (MINOR) — `clodex server` has no consumer for the blocked reason, and the all-blocked message names the wrong repair.** `blockedProviders` is read in exactly one place in `src/` (`src/cli.ts:1159`). `resolveServerUpstreamApiKey` (`src/server/index.ts:352-359`) has the loaded catalog in hand, discards the map, and returns `null`; `runServerCommand` then prints `No providers configured. Run \`clodex providers\` to add OpenAI.` (`src/server/index.ts:377`). Reproduced with both overrides set:

```
$ CLODEX_KEY_OPENAI=o1 CLODEX_KEY_OPENAI_OAUTH=o2 clodex server --endpoint --quick --no-discovery
exit=1   ■ No providers configured. Run `clodex providers` to add OpenAI.
```

Providers *are* configured; the cause is a process-scoped override. In the partial case the blocked provider is simply absent with no message (the closest signal, when a saved alias targets it, is `1 saved model alias inactive … target unavailable`). The behavior remains fail-closed; the all-blocked case is loud but misdiagnosed, while the partial omission is silent. `docs/background-agents.md:102` promises the tailored repair message when a blocked provider is selected, and `clodex server` has no selection step. A `p.log.warn` per blocked provider plus a blocked-aware branch on the `!apiKey` path would close both observability gaps.

**O2 (MINOR, coverage) — nothing pins that `fetchProviderCatalog` supplies the diagnostic sinks.** `src/provider-catalog.ts:19-20` supplies both callbacks, and `src/registry/load.ts:22` forwards the registry diagnostic. Reverting either to base's shape (`loadRegistryProviders(undefined, { ...opts })`, `loadRegistry()`) leaves the full suite at 1,576/1,576 green, i.e. a byte-level revert of the optional-7 delivery ships green. One test driving the real `fetchProviderCatalog` against an on-disk registry with `vi.spyOn(console, 'warn')` kills both. Note this pins the common catalog seam only — `first-run.ts`, `patcher.ts` and `providers-command.ts` call `loadRegistry` without a sink, so it is not a universal lenient-loader guarantee.

**O3 (MINOR, coverage) — the unchanged-v5 read-path fail-closed branch is unpinned.** Deleting `src/registry/io.ts:475-477` survives the full suite; the two durability tests that assert a throw (`tests/registry-io-durability.test.ts:279,325`) both enter the `migration.changed === true` path rather than the early unchanged-registry branch. Nothing constructs an already-materialized v5 registry that needs no migration and whose parent fsync fails. The opposite-direction control (always fatal) *is* killed, so the branch is not dead from the suite's point of view — one case in the existing durability harness closes it.

**O4 (MINOR, recorded only) — the fsync tolerance also covers the selection-clearing downgrade.** `hasMaterializedActiveAccount` (`src/registry/io.ts:444-447`) requires `schemaVersion === 5` and some `defaultAuthRef`, so a registry whose v5 selection was just cleared back to v1/v2 takes the tolerated branch at `:478-479` — even though the comment at `:467-471` names that transition as one the re-sync is meant to repair. The re-sync is still attempted; only the escalation is skipped. This is the change the previous round asked for, and reaching a bad outcome needs an `EIO`/`ENOSPC`-class directory fsync failure, an ignored loud write-path throw, and an unclean shutdown; the worst case is reverting to the user's *own* prior slot. No change requested — recorded so it is not rediscovered as a defect.

**O5 (NIT) — `blockedProviders` is a non-enumerable property on the array.** `src/registry/load.ts:75-78` attaches it via `Object.defineProperty`, and `LoadedRegistryProviders` (`:13-15`) does not survive `map`/`filter`/spread. Today's single consumer reads it at `src/cli.ts:1159` *before* `providersForTarget(providersForPicker(catalog), 'claude')` at `:1157` narrows the type, so nothing is broken; a future consumer that filters first would silently lose the reasons.

**O6 (NIT) — the new diagnostics land inside the clack frame.** `src/provider-catalog.ts:19-20` routes both channels to `console.warn`, so `Provider registry dropped invalid provider "…"` prints between UI frames rather than through `p.log.warn`. Content is bounded and leaks nothing; ordinary-path output is byte-identical to base.

**O7 (NIT, no action) — `src/registry/refresh-models.ts:383` (`current.addedAt !== routing.addedAt`) is an unpinned conjunct.** Deleting it survives the full suite; it is almost certainly an equivalent mutant (`addedAt` is the provider registration timestamp and does not change for an existing provider, and the `selectedAccount` deep-equal at `:408` carries the weight — removing *that* is killed). Noted only so it is not later read as a tested invariant.

PR #99, opened 2026-08-10
feat(providers): support mixed-protocol metadata

zachspartofaday:
## Summary

- make provider/model metadata carry request and response protocol independently instead of inferring both from one package id
- propagate the representation through registry conversion, materialization, model catalogs, runtime selection, server routes, and upstream forwarding
- validate runtime compatibility before a mixed-protocol model can launch
- preserve the client-requested model identity when an Anthropic passthrough route forwards a different upstream id

This is **OpenCode stack 1/6**, the protocol foundation extracted from #94. It is intentionally provider-generic; the OpenCode Go provider itself is the next stacked PR.

## Reachability

The changed metadata flows from provider templates and registry records into every active launch surface: direct proxy, HTTP proxy, endpoint server, catalog display, and passthrough forwarding. Behavioral tests cover conversion, materialization, compatibility rejection, route selection, model-id echo, chunk-split SSE, and malformed upstream events.

## Validation

- focused protocol suite — 8 files / 126 tests passed
- `pnpm typecheck` — passed
- `pnpm test` — 85 files / 1,332 tests passed
- `pnpm build` — passed
- `git diff --check` — passed

Local validation used Node 22.23.1, the supported runtime floor. CI exercises the repository's pinned Node 24 target. No live provider request was made.

## Stack

1. **This PR:** mixed-protocol provider foundation
2. Base OpenCode Go provider and hardening — targets this branch
3. Pinned CLI resolver contracts and deterministic catalog
4. Anthropic Messages route provenance
5. Global exact-effort policy and native representation
6. Same-target alias preservation

Each successor targets its predecessor so every review diff remains coherent. The final stack will be checked against #94's behavior tree before #94 is superseded.

### DIFF
--- a/src/catalog.ts
+++ b/src/catalog.ts
@@ -54,6 +54,7 @@ export function localModelToRoute(lp: LocalProvider, model: LocalProviderModel):
     interleavedReasoningField: model.interleavedReasoningField,
     useResponsesLite: model.useResponsesLite,
     preferWebSockets: model.preferWebSockets,
+    compatibility: model.compatibility,
   };
 }
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1400,6 +1400,7 @@ export async function runClaudeCommand(parsed: ParsedArgs): Promise<number> {
           interleavedReasoningField: selectedModel.interleavedReasoningField,
           useResponsesLite: selectedModel.useResponsesLite,
           preferWebSockets: selectedModel.preferWebSockets,
+          compatibility: selectedModel.compatibility,
           headers: activeProvider.headers,
         },
         launchApiKey ?? '',
--- a/src/http-proxy/routes.ts
+++ b/src/http-proxy/routes.ts
@@ -45,7 +45,7 @@ export interface ResolvedHttpProxyAlias {
   sourceNames?: string[];
 }
 
-/** Build a positive allowlist: only favorite AI-SDK routes can leave Anthropic's path. */
+/** Build a positive allowlist: only explicitly configured favorite routes can leave Anthropic's path. */
 export function buildHttpProxyRoutes(
   providers: LocalProvider[],
   favorites: FavoriteModel[],
@@ -66,7 +66,10 @@ export function buildHttpProxyRoutes(
       unavailable.push(favorite);
       continue;
     }
-    if (model.modelFormat !== 'openai' || !isSdkMigratedNpm(model.npm)) {
+    const supported = model.modelFormat === 'anthropic'
+      ? Boolean(model.baseUrl)
+      : isSdkMigratedNpm(model.npm);
+    if (!supported) {
       unsupported.push(favorite);
       continue;
     }
--- a/src/model-runtime-compatibility.ts
+++ b/src/model-runtime-compatibility.ts
@@ -1,0 +1,110 @@
+/**
+ * Provider-neutral per-model compatibility hints.
+ *
+ * These describe Chat Completions wire quirks that cannot be inferred safely
+ * from a provider or model family. Runtime adapters consume only the fields
+ * they understand; unknown fields remain inert metadata.
+ */
+export interface ModelRuntimeCompatibility {
+  /** Claude/Codex effort -> upstream reasoning_effort map. null disables that level. */
+  reasoningEffortMap?: Record<string, string | null>;
+  /** Explicitly enable or disable reasoning_effort for this model. */
+  supportsReasoningEffort?: boolean;
+  /** Additional request shape required when reasoning is enabled. */
+  thinkingFormat?: 'deepseek' | 'qwen';
+  /** Replay an empty reasoning_content field when prior assistant reasoning is absent. */
+  requiresReasoningContentOnAssistantMessages?: boolean;
+  /** Whether the upstream accepts the OpenAI `store` request field. */
+  supportsStore?: boolean;
+  /** Whether the upstream accepts Chat Completions `developer` messages. */
+  supportsDeveloperRole?: boolean;
+  /** Output-token field accepted by the upstream Chat Completions endpoint. */
+  maxTokensField?: 'max_tokens' | 'max_completion_tokens';
+  /** Whether the upstream accepts long prompt-cache retention controls. */
+  supportsLongCacheRetention?: boolean;
+}
+
+export type OpenAiCompatibleRequestBody = Record<string, unknown>;
+
+type ChatMessage = Record<string, unknown> & { role?: unknown };
+
+function remapMaxTokensField(
+  body: OpenAiCompatibleRequestBody,
+  field: ModelRuntimeCompatibility['maxTokensField'],
+): void {
+  if (field === 'max_tokens') {
+    if (body.max_tokens === undefined && body.max_completion_tokens !== undefined) {
+      body.max_tokens = body.max_completion_tokens;
+    }
+    delete body.max_completion_tokens;
+    return;
+  }
+  if (field === 'max_completion_tokens') {
+    if (body.max_completion_tokens === undefined && body.max_tokens !== undefined) {
+      body.max_completion_tokens = body.max_tokens;
+    }
+    delete body.max_tokens;
+  }
+}
+
+function transformMessages(
+  messages: unknown,
+  compatibility: ModelRuntimeCompatibility,
+): unknown {
+  if (!Array.isArray(messages)) return messages;
+  const rewriteDeveloper = compatibility.supportsDeveloperRole === false;
+  const replayReasoning = compatibility.requiresReasoningContentOnAssistantMessages === true;
+  if (!rewriteDeveloper && !replayReasoning) return messages;
+
+  let changed = false;
+  const transformed = messages.map(message => {
+    if (!message || typeof message !== 'object' || Array.isArray(message)) return message;
+    const source = message as ChatMessage;
+    const role = source.role;
+    const needsRoleRewrite = rewriteDeveloper && role === 'developer';
+    const needsReasoningReplay = replayReasoning
+      && role === 'assistant'
+      && !Object.prototype.hasOwnProperty.call(source, 'reasoning_content');
+    if (!needsRoleRewrite && !needsReasoningReplay) return message;
+
+    changed = true;
+    return {
+      ...source,
+      ...(needsRoleRewrite ? { role: 'system' } : {}),
+      ...(needsReasoningReplay ? { reasoning_content: '' } : {}),
+    };
+  });
+
+  return changed ? transformed : messages;
+}
+
+/**
+ * Apply model-specific OpenAI-compatible request quirks after the AI SDK has
+ * assembled its Chat Completions payload.
+ */
+export function transformOpenAiCompatibleRequestBody(
+  body: OpenAiCompatibleRequestBody,
+  compatibility: ModelRuntimeCompatibility,
+): OpenAiCompatibleRequestBody {
+  const transformed = { ...body };
+
+  if (compatibility.supportsStore === false) delete transformed.store;
+  if (compatibility.supportsLongCacheRetention === false) {
+    delete transformed.prompt_cache_retention;
+    delete transformed.promptCacheRetention;
+  }
+  remapMaxTokensField(transformed, compatibility.maxTokensField);
+
+  const messages = transformMessages(transformed.messages, compatibility);
+  if (messages !== transformed.messages) transformed.messages = messages;
+
+  const hasReasoningEffort = typeof transformed.reasoning_effort === 'string'
+    && transformed.reasoning_effort.trim().length > 0;
+  if (hasReasoningEffort && compatibility.thinkingFormat === 'deepseek') {
+    if (transformed.thinking === undefined) transformed.thinking = { type: 'enabled' };
+  } else if (hasReasoningEffort && compatibility.thinkingFormat === 'qwen') {
+    if (transformed.enable_thinking === undefined) transformed.enable_thinking = true;
+  }
+
+  return transformed;
+}
--- a/src/patcher.ts
+++ b/src/patcher.ts
@@ -278,6 +278,7 @@ export function buildDesiredPatchConfig(): DesiredPatchConfig {
         reasoning: model.reasoning ?? modelsDev?.reasoning,
         interleavedReasoningField:
           model.interleavedReasoningField ?? modelsDev?.interleaved?.field,
+        compatibility: model.compatibility,
         upstreamModelId,
       });
       meta.set(`${provider.id}:${model.id}`, {
--- a/src/provider-catalog.ts
+++ b/src/provider-catalog.ts
@@ -122,6 +122,7 @@ export function localProvidersToServerModels(localProviders: LocalProvider[]): S
       interleavedReasoningField: model.interleavedReasoningField,
       useResponsesLite: model.useResponsesLite,
       preferWebSockets: model.preferWebSockets,
+      compatibility: model.compatibility,
       headers: provider.headers,
       providerData: provider.providerData,
     }))
--- a/src/provider-factory.ts
+++ b/src/provider-factory.ts
@@ -14,6 +14,10 @@ import {
   injectClaudeIdentity,
 } from './oauth/claude-identity.js';
 import { isCredentialBearingHeader } from './credential-headers.js';
+import {
+  transformOpenAiCompatibleRequestBody,
+  type ModelRuntimeCompatibility,
+} from './model-runtime-compatibility.js';
 
 /** Models that must use /v1/responses instead of /v1/chat/completions. */
 const RESPONSES_ONLY_PREFIXES = [
@@ -110,6 +114,8 @@ export interface ProviderModelSpec {
   useResponsesLite?: boolean;
   /** Backend capability: model must use the WebSocket Responses transport instead of HTTP. */
   preferWebSockets?: boolean;
+  /** Provider-neutral per-model wire quirks. */
+  compatibility?: ModelRuntimeCompatibility;
   /** Optional debug logger (wired to the proxy trace log) for transport-level diagnostics. */
   onDebug?: (msg: string) => void;
   /** Optional privacy-safe structured WebSocket diagnostics. */
@@ -249,6 +255,12 @@ export async function createLanguageModel(spec: ProviderModelSpec): Promise<Lang
       ...(spec.authType !== 'none' && apiKey.trim() ? { apiKey } : {}),
       ...(spec.authType === 'none' ? { fetch: fetchWithoutCredentialHeaders } : {}),
       ...(spec.headers ? { headers: spec.headers } : {}),
+      ...(spec.compatibility
+        ? {
+            transformRequestBody: (body: Record<string, unknown>) =>
+              transformOpenAiCompatibleRequestBody(body, spec.compatibility!),
+          }
+        : {}),
     };
     model = createOpenAICompatible({
       ...options,
@@ -292,6 +304,8 @@ export interface ReasoningMetadata {
   supportedParameters?: string[];
   reasoning?: boolean;
   interleavedReasoningField?: string;
+  /** Provider-neutral per-model wire quirks. */
+  compatibility?: ModelRuntimeCompatibility;
   /**
    * Bare upstream model id (e.g. 'grok-4.5'), distinct from the request's `model`
    * field which may be a gateway alias or catalog slug (e.g. 'xai-oauth__grok-4.5').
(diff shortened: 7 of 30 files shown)


bman654 (approved):
# Review — #99 `feat(providers): support mixed-protocol metadata`

This landed with real verification behind it, and our own review had to be walked back in places, so the short version first: **there is nothing blocking here.** Everything below is optional.

What holds up under execution: the two shipped providers are provably unaffected — a 1,386-row capability/effort differential over base-vs-head implementations comes out `0 rows differ`, and a 200-case real `sdkTranslateRequest` corpus is byte-identical (30,982 bytes both sides, `diff` empty). The echo mechanism is correct end to end over a real socket for canonical / bare-canonical / alias / masked / `[1m]` ids, streaming and non-streaming, including a `message_start` deliberately split mid-`"model":"` across two TCP writes; `Content-Length` is recomputed correctly and `count_tokens` correctly omits the override. The `[1m]` case is a genuine fix — base echoed the bare upstream id, so a patched Claude Code resolved 200k for a 1M-context model. `transformRequestBody` really reaches the wire on `@ai-sdk/openai-compatible` 3.0.7 in both generate and stream modes. Responses-API selection and the OAuth WebSocket transport are untouched, with and without `compatibility` and even with a foreign `baseURL` present. Leaving `PATCH_TRANSFORMS_VERSION` alone is correct.

We did file one blocking ask internally — the `effortProviderOptions` presence-gate. A later pass overturned it on reachability: we asserted the triggering shape was "what stack PR 2 will author," and that is false. We re-checked the real heads (#100 `c72890d`, #104 `3c82e66`) and every OpenAI-compatible row in `src/data/opencode-go-models.json` carries `supportsReasoningEffort`, `reasoningEffortMap`, or `thinkingFormat` — zero triggering rows in either. It survives below as optional hardening only. Apologies for the churn; we'd rather correct it here than have you spend a cycle on it.

The two items most worth acting on before #100 lands are (1) and (2).

## Optional

All non-blocking. Ordered by how much they matter to the rest of the stack.

**1. Per-model `apiUrl` survival through add/refresh is unpinned.** Mutating all three preservation expressions — `src/registry/add-template.ts:121`, `src/registry/refresh-models.ts:254`, `src/registry/refresh-models.ts:277` — from `m.apiUrl ?? fetched.baseUrl` back to `fetched.baseUrl` recreates the old one-URL-per-provider behavior and destroys mixed-protocol routing after add/refresh, with the whole shipped suite green (`85 files / 1332 tests`). We exercised `fetchTemplateModels`, conversion and materialization, but not the two commands that persist refreshed models. Dormant in this PR; live the moment #100 supplies the mixed catalog. Worth an add/refresh persistence test in the successor.

**2. The echo decision logic is unpinned at both call sites and on the streaming branch.** Forcing `responseModelOverride: undefined` at `src/proxy.ts:540` and `src/server/router.ts:358`, plus deleting the whole SSE rewrite pipe from `src/upstream-forward.ts`, leaves `pnpm vitest run` at **85 files / 1332 tests, zero failures** (reproduced independently three times). The two shipped `relayAnthropicMessages` tests both pass `clientWantsStream=false` (`tests/upstream-forward.test.ts:217`, `:237`), so the streaming branch is never entered. The remedy is small and lives in files that already exist:

- **1 line** in `tests/proxy.test.ts:133` (`'forwards anonymous Anthropic routes without authentication headers'`), which already drives the exact override branch and simply never asserts the echo: `expect(JSON.parse(res.body).model).toBe(route.aliasId)` fails under the mutation with `expected 'anonymous-model' to be 'anthropic-local__anonymous-model'`.
- **~25 lines** as a sibling of `tests/server-router.test.ts:1254` (`'resolves a bare saved alias and echoes it back verbatim…'`), reusing that file's own `startUpstream` / `startTestServer` / `createGatewayModelCatalog` helpers with a `modelFormat:'anthropic'` passthrough.
- **~40 lines** driving `relayAnthropicMessages` with `clientWantsStream=true` and a stubbed `fetch` returning SSE. A probe of that shape fails under the mutation with `Received: "model":"claude-sonnet-4-5"` / `Expected: "model":"clodex:acme:sonnet[200k]"`. Prefer awaiting `'finish'` on the stub response over a fixed settle timeout so it can't flake on a loaded CI box.

This is coverage debt, not a defect, and PR #99 alone has no producer for an Anthropic catalog model. #100 adds four (`minimax-m3`, `qwen3.7-max`, `qwen3.7-plus`, `qwen3.8-max`), which is when it goes live.

**3. `compatibility` propagation past materialization is unpinned at 10 sites.** `src/catalog.ts:57`, `src/provider-catalog.ts:125`, `src/server/router.ts:415`, `:735`, `src/server/index.ts:184`, `src/cli.ts:1403`, `src/proxy.ts:590`, `:610`, `:867`, `src/patcher.ts:281` — all ten deleted at once (leaving `materialize.ts` and `convert.ts` intact so they can't mask anything) → **85 files / 1332 tests green**. Both *ends* are genuinely pinned; what's unpinned is the middle plumbing on a currently-dormant value. One route-level test carrying a `compatibility` object from a fake registry entry through `startProxyCatalog` and `getOrInitLanguageModel` would close most of the ten. This is the PR's headline deliverable, so it's the gap most worth tracking, and it becomes material as soon as the OpenCode Go provider actually emits `compatibility`.

**4. Effort predicate asymmetry (the de-escalated ask).** `effortProviderOptions` treats any OpenAI-compatible `compatibility` object as reasoning-authoritative (`src/provider-factory.ts:955` gates on presence of the object), while the capability classifier only does so when a reasoning field is present (`compatibilityReasoningCapabilities`, `src/provider-factory.ts:643`, consumed at `:734` and `:909-910`). A hypothetical future record carrying only a wire-shape quirk would suppress otherwise-supported effort — the UI would keep advertising `controllable` while nothing reaches the wire. Gate both sides on the same reasoning-field predicate (`supportsReasoningEffort !== undefined || reasoningEffortMap || thinkingFormat !== undefined`) and add a non-reasoning case if you want this internal API to fail safely. **No current shipped or stacked catalog row triggers it.** For reference, the fix was applied and verified twice: `pnpm vitest run tests/provider-factory.test.ts tests/model-runtime-compatibility.test.ts tests/patcher.test.ts` → 146 passed, with the `maxTokensField`-only row regaining `patch.levels=["low","medium","high"]` and wire `reasoning_effort:"high"`.

**5. Heads-up, not a change here: #98 conflicts with this PR's pricing policy.** `git merge-tree --write-tree f2126c3 <pr98>` reports `CONFLICT (content)` in exactly `src/registry/pricing.ts` and `tests/pricing.test.ts`. The semantic resolution has to keep **both** behaviors — `preserveModelPricing` must skip enrichment for the opted-out provider, while #98's account-specific loop must still enrich both top-level and named-account caches for providers that don't opt out — and both pricing tests must survive. Taking either side wholesale loses one guarantee. #96 and #97 merge clean against this head (exit 0, no markers).

**6. `npm: '@ai-sdk/openai'` paired with a per-model `apiUrl` silently ignores the URL.** `src/provider-factory.ts:167` onward never passes `baseURL` on the `@ai-sdk/openai` branch. Real `generateText` capture with a third-party key: `{"url":"https://api.openai.com/v1/responses","authorization":"Bearer THIRD-PARTY-GATEWAY-KEY"}` — the credential leaves the intended domain. On base this pairing was structurally impossible (`add-template.ts:120` and `refresh-models.ts` forced `apiUrl: fetched.baseUrl`, and one npm per provider); head makes both per-model overridable (`src/registry/fetch-template-models.ts:149`, `:167`). Optional because the trigger is a curator authoring a self-inconsistent record in reviewed source and the failure is loud (OpenAI 401 on the first request). If you want a guard, `resolveEndpoint` already drops the anthropic-without-baseUrl case and is the natural home.

**7. Consider normalizing `compatibility` at the registry boundary before stack PR 3.** `src/registry/io.ts:116` accepts model records on object-ness alone, so `compatibility: { reasoningEffortMap: 'not-an-object' }` in `providers.json` survives `loadRegistry()` verbatim and `Object.entries` over the string yields `levels:["0"…"12"], defaultLevel:"0"`, ending at `{"opencodeGo":{"reasoningEffort":"n"}}` on the wire via the endpoint server's `defaultEffort`. Two things shrink this: the patched-binary half does *not* happen (`projectNativeEffort`, `src/patch-transforms.ts:58-68`, requires low+medium+high, so `BAKED_INTO_BINARY = undefined`), and "never validated" is the pre-existing posture of the whole model cache — base is in fact worse, throwing an uncaught `TypeError` out of `getReasoningCapabilities` for a string `supportedParameters`. So: hygiene today, one live path, and a ~15-line normalizer is cheaper now than after three more PRs stack on it, since PR 3 sources model metadata from an external CLI.

**8. Three of the eight `ModelRuntimeCompatibility` knobs are inert, and a fourth is a no-op in one direction.** `src/model-runtime-compatibility.ts:8-25`. `transformRequestBody` is installed only on the `@ai-sdk/openai-compatible` branch (`src/provider-factory.ts:258`), and that SDK never emits `store`, `prompt_cache_retention`, or a `developer` role. Real wire capture with the full blob applied contains none of the three. `maxTokensField:'max_tokens'` is likewise a no-op (the SDK already emits it); only the `'max_completion_tokens'` direction does work. The one route that *does* emit `store:false` (`src/provider-factory.ts:1089-1092`, `@ai-sdk/openai`) never receives the transform. Either drop them until a path emits those bytes, or note in the interface doc that they're forward-looking — the risk is a curator marking a `developer`-hostile upstream "handled" when clodex is doing nothing.

**9. The proxy-side override also fires on the default-route fallback.** `src/proxy.ts:540-543` evaluates `originalModel !== route.realModelId` *after* `const route = resolvedRoute ?? defaultRoute` (`src/proxy.ts:434`), so a request for a model with no catalog route lands on the default route and gets its own unresolved id echoed back while a different model served the turn (executed: echo `claude-3-5-haiku-20241022`, upstream actually served `claude-sonnet-4-5-20250929`). Reachability is narrow — the MITM sends unresolved ids straight to real Anthropic (`src/http-proxy/server.ts:874`) and the endpoint server 400s them — but this is the one *behavior* delta from base in the echo area; everything else is coverage. Suggested: `resolvedRoute && typeof originalModel === 'string' && originalModel !== route.realModelId ? originalModel : undefined`.

**10. `responseModelOverride` is format-agnostic on the JSON path but Anthropic-only on the SSE path.** `src/upstream-forward.ts:126-154` vs `:230-237`. `relayAnthropicMessages` also serves the OpenAI-format passthrough (`src/server/router.ts:579`), which rewrites the *request* model but sets no override. If someone later sets the option there, the JSON branch rewrites any top-level `model` while the SSE branch only matches `message_start`, so stream and non-stream would disagree silently. Either narrow the JSON rewrite to `type === 'message'`, or rename the option (`anthropicResponseModelOverride`) so the contract can't be misread.

**11. The route-support gate inverted from allowlist to deny-one/allow-the-rest.** `src/http-proxy/routes.ts:69-71`. Base was `if (model.modelFormat !== 'openai' || !isSdkMigratedNpm(model.npm))`; head special-cases `'anthropic'` and lets every other value through on `isSdkMigratedNpm(npm)`. A 768-row route corpus shows 288 rows differing, of which only 96 are the intended anthropic admission — the other 192 flip from `unsupported` to a live route. Unreachable today (`LocalProviderModel.modelFormat` is typed `'anthropic' | 'openai'` at `src/types.ts:23` and `src/registry/materialize.ts:67` coerces), but it inverts the function's own docstring right as the stack starts adding protocols. Suggested: `model.modelFormat === 'openai' ? isSdkMigratedNpm(model.npm) : model.modelFormat === 'anthropic' ? Boolean(model.baseUrl) : false`.

**12. Cheap single-assertion coverage gaps**, each verified by a surviving mutation under a full-file run:

- `reasoningEffortMap` **allowlist** semantics (`src/provider-factory.ts:717-719`) — the one surviving mutation in an otherwise-clean set (95/95 green when "effort not in the map" falls back instead of being refused). One assertion: `effortProviderOptions('@ai-sdk/openai-compatible','medium','custom-reasoning-model', metadata)` is `undefined`.
- Six branches of the new compatibility→reasoning mapping (`src/provider-factory.ts:643-726`) survive deletion at 143/143. Two have real behavioral consequences: the added `&& !metadata?.compatibility?.reasoningEffortMap` guard (deleting it turns `{reasoning:false, compatibility:{reasoningEffortMap:{high:'max'}}}` from `levels:["high"]` into `EMPTY_REASONING`), and `|| compatibility.thinkingFormat !== undefined` (`{thinkingFormat:'deepseek'}` alone currently yields `levels:["low","medium","high"], wireFormat:{kind:"deepseek-thinking"}`, asserted nowhere). The `wireFormat` ones are cosmetic — `grep -rn "\.wireFormat" src/` is empty, it has no reader.
- The deepseek `thinking` gate (`src/model-runtime-compatibility.ts:103-104`): dropping `hasReasoningEffort &&` is a free mutation (62/62 green) while its qwen twin *is* pinned. Add the deepseek half of the existing qwen test.
- The rewritten `static-seed` template path (`src/registry/fetch-template-models.ts:144-168`) has no tests at all — `grep -rn "static-seed" tests/` is empty; dropping curated fields, dropping the `apiUrl` fallback, and ignoring per-model npm all stay green 32/32.
- Curated-overlay guards (`src/registry/fetch-template-models.ts:170-205`) are all green 32/32. The one that matters is `hasFreeMetadata` (`:177-179`) — it exists so a curated entry with no `cost`/`isFree`/`freeStatus` doesn't clobber the free/paid classification from the live listing, but the fixture's live rows are all paid, so its presence and absence are indistinguishable. Add one live row with zero pricing plus an overlay omitting free metadata, assert `freeStatus` stays `verified_free`.
- `src/upstream-forward.ts:149-152` `flush()` is unpinned because every fixture ends in `\n\n`; one fixture ending mid-line pins it. Same for strict `preserveModelPricing` validation (`src/registry/io.ts:135-137`).

**13. Small stuff.**

- `preserveModelPricing` is provider-level, and base's `parseProvider` has no case for it (`grep -n preserveModelPricing src/registry/io.ts` in the base worktree is empty), so an older clodex loads a head-written registry, drops the flag, writes it back without it, and — having no opt-out — overwrites the curated `cost` values the flag protects. Bounded (cost display, restored on the next refresh); either persist it under `api`, which is copied wholesale, or accept the loss in a comment. Model-level `compatibility`/`modalities` survive intact.
- `applyTemplateModelMetadata`'s third parameter `_baseUrl` (`src/registry/fetch-template-models.ts:217`) is genuinely dead — drop it.
- `src/upstream-forward.ts:144-147` splits on `\n` and rejoins with `\n`, so a CRLF upstream loses the `\r` on the rewritten `message_start` line only. Spec-legal and Anthropic uses LF; recording it so it isn't rediscovered. Likewise SSE-legal multi-line `data:` fields pass through unrewritten — already in the code comment, but the *new* `modelFormat:'anthropic'` admission is exactly the kind of endpoint that might emit them.
- `delete transformed.promptCacheRetention` (`src/model-runtime-compatibility.ts:94`) has no producer anywhere in `src/` or in the SDK dist; either comment it as defensive or drop it. Same for the whitespace-only `reasoning_effort` guard (`:101-102`) — `effortProviderOptions` returns `undefined`, never `''`.
- `supportsReasoningEffort:false` isn't enforced at the last mile: `transformOpenAiCompatibleRequestBody` never deletes `reasoning_effort`. Currently sufficient (`effortProviderOptions` is the only producer on this path, `src/sdk-adapter.ts:501-503`), but the transform is the natural belt-and-braces spot and would also cover the value-remap divergence in the next bullet.
- Second axis of the #4 asymmetry: `getReasoningCapabilities` / `getPatchReasoningCapabilities` consult `compatibility` for **every** npm, while `effortProviderOptions` and `createLanguageModel` consult it only under `@ai-sdk/openai-compatible`. What's left after the scary version was refuted (below) is a value-remap divergence — with `reasoningEffortMap:{low:null,high:'max'}` on `@ai-sdk/openai`, the UI promises `high→max` and the wire gets `high`, and `max` stays selectable even though the map hides it. Note the "make every npm honor it" direction isn't implementable: `transformRequestBody` exists only on `@ai-sdk/openai-compatible` (`grep -c transformRequestBody node_modules/@ai-sdk/openai/dist/index.d.ts` → 0). The other direction is free — scoping `compatibilityReasoningCapabilities` to `@ai-sdk/openai-compatible` keeps the four relevant suites at 95/95.

Established by execution, mostly corrections to scary-sounding claims we raised ourselves and then killed. Recording them so nobody re-derives them.

1. **The two shipped providers are byte-identical to base.** 1,386-case corpus (9 npm × 14 model ids × 11 metadata shapes × three entry points × 9 efforts) → `BASE_HEAD_IDENTICAL_ROWS 1386`, independently reproduced by importing both implementations into one Vitest process. An independent 945-row / 791 KB dump → `IDENTICAL`. A 200-case real `sdkTranslateRequest` corpus → `diff` empty, 30,982 bytes both sides. The new branches are genuinely inert without `compatibility`, including the added `&& !metadata?.compatibility?.reasoningEffortMap` guard and the `const candidate = effort` refactor (a pure rename).

2. **Leaving `PATCH_TRANSFORMS_VERSION` alone is correct**, and the `patcher.ts` one-liner cannot churn any existing user's config hash. `git diff 4ee1a49 HEAD --name-only | grep -c patch-transforms` → `0`; the source-digest pin passes for the right reason. No spurious `stale-config` re-patch.

3. **`supportsReasoningEffort:false` does NOT leak `reasoning_effort` on any npm.** Verified twice by modelling the real chain rather than hand-feeding `effortProviderOptions`: `buildDesiredPatchConfig` bakes effort only when `mode === 'controllable'` (`src/patcher.ts:288`), `enrichServerModelReasoning` sets `defaultEffort` only on a truthy `caps.defaultLevel` (`src/server/index.ts:186`), and `anthropicEffortFromRequest` reads only `body.output_config.effort` (`src/sdk-adapter.ts:126-130`). For `@ai-sdk/openai`, `@ai-sdk/anthropic`, `@ai-sdk/google` and `@ai-sdk/openai-compatible`: `patchMode=internal-only baked=undefined serverDefaultEffort=undefined resolvedEffort=undefined WIRE=undefined`, including for an empty/all-null map. `@ai-sdk/anthropic` is passthrough anyway — it never calls `translateRequest`.

4. **Garbage `compatibility` cannot reach the patched binary.** `projectNativeEffort` (`src/patch-transforms.ts:58-68`) requires low+medium+high and rejects a `defaultLevel` outside `levels`, so junk levels produce no `effort` key: `PATCH_CAPS levels=["0","1","2",…]` → `BAKED_INTO_BINARY=undefined`. The same guard means empty effort levels can never throw `PatchApplyError: effort … must include low, medium, and high`. The patch-level `max` drop under `{low:null,medium:'medium',high:'high',max:'high'}` is the documented de-duplication (identical provider options collapse), not a bug.

5. **`transformRequestBody` is not theatre.** `@ai-sdk/openai-compatible` 3.0.7 forwards the provider-settings hook into the chat model in both `doGenerate` and `doStream`; real captures show the transform's output on the wire in both modes. `{opencodeGo:{reasoningEffort:'max'}}` becomes `reasoning_effort:"max"`, not a camelCase field — the SDK resolves both raw and camelCased provider names and snake_cases the arg, which matters because the deepseek/qwen envelope keys on `body.reasoning_effort`.

6. **Responses-API selection and the OAuth WebSocket transport are untouched by the new metadata**, with and without `compatibility` and even with a foreign `baseURL` present (`authType:'oauth'` still resolves `https://chatgpt.com/backend-api/codex/responses` with `customFetch=true`). Worth knowing for future review: in this fork `modelPrefersResponsesApi` is *not* the endpoint selector — `shouldUseOpenAiResponsesEndpoint` is. The OAuth model-refresh path never goes through `fetchTemplateModels`, so a template overlay cannot touch `preferWebSockets` / `useResponsesLite`.

7. **The echo mechanism itself is correct and could not be broken.** Verbatim echo for canonical `clodex:` id, bare canonical, short alias, masked gateway id and `[1m]`, streaming and non-streaming, through a real `startProxyCatalog` in front of a real loopback upstream that splits `message_start` mid-`"model":"` across two TCP writes. The forwarded *request* carries `realModelId` and the upstream URL is exactly `/v1/messages`. Byte-by-byte 1-byte chunking preserves both rewrite and framing; a UTF-8 character split inside its 3 bytes survives; a `message_start` with no trailing newline at EOF is still rewritten by `flush`; a `content_block_delta` whose *text* contains `message_start` and a `"model"` key comes back byte-identical; `callback(null,'')` on a newline-free chunk does not truncate a real chunked response; first-byte latency is unchanged (~1 ms) against a 250 ms mid-line stall.

8. **The Anthropic `/v1` constraint is not violated.** 120-combination matrix (5 npm × 8 apiUrl × 3 modelFormat) through the real `cachedModelToLocal` → `localModelToRoute`: 16 produce an Anthropic route, and in all 16 `route.upstreamUrl` fails `/\/v1\/?$/` while `${upstreamUrl}/v1/messages` ends in exactly `/v1/messages`. Structural — `baseUrl` has one producer (`src/registry/materialize.ts:21-26`) and it strips a trailing `/v1`; the only other `/v1`-bearing path (`src/registry/custom-endpoint.ts:64`) strips it too. `src/registry/convert.ts:20`'s `apiUrl: model.apiBaseUrl ?? model.baseUrl` round-trips stably in both the present and absent cases.

9. **A format/npm disagreement fails closed.** `modelFormat:'anthropic'` on a non-anthropic npm gets no `baseUrl`, `localModelToRoute` returns `null`, and the new gate classifies it `unsupported`; `src/server/router.ts:298` 400s the same shape. The widened allowlist does not reach a missing handler — `startProxyCatalog` already has the Anthropic passthrough branch (`src/proxy.ts:505`) and `count_tokens` already keys off `route.modelFormat !== 'anthropic'` (`src/proxy.ts:435`).

10. **No overclaim in the PR body.** `src/http-proxy/routes.ts:69-71` *is* a new runtime-compatibility precondition — an anthropic-format model with no `baseUrl` cannot launch — so "validate runtime compatibility" reads as protocol compatibility, not as validating the `ModelRuntimeCompatibility` blob. The body never claims call-site or end-to-end echo coverage; "behavioral tests cover … model-id echo, chunk-split SSE" is literally true at the helper level (`tests/upstream-forward.test.ts:153`, `:162`, `:173`, `:206`).

11. **Registry compatibility is fine in both directions.** An older clodex's *strict* loader does not reject a head-written registry, so a downgrade cannot brick destructive operations; model-level `compatibility`/`modalities` survive a downgrade round-trip because `modelsCache.models` are passed through raw. This head parses older registries — every new field is optional and the pre-existing fixtures pass unchanged. `apiUrl: m.apiUrl ?? fetched.baseUrl` cannot change the shipped providers because `parseModelList` never writes `apiUrl`. Both shipped templates fetch identically base-vs-head.

12. **Do not "fix" these surviving mutations — they're semantically equivalent, not coverage gaps:** the `'"message_start"'` substring pre-filter and the `parsed.type` check in `anthropicSseModelRewrite` (perf guards; only `message_start` carries `message.model`); `map[effort] ?? undefined` (`src/provider-factory.ts:718-719` — the `hasOwnProperty` check only rejects prototype keys, and `effortProviderOptions(npm,'constructor',…)` correctly returns `undefined` today); the curated-overlay id normalization (normalizes to the same value); the `?? model.baseUrl` fallback in `convert.ts` (round-trips correctly and `resolveEndpoint` still strips `/v1`).

13. **The red suites on our machines are ambient, not this PR.** `credential-helper`, `http-proxy-index`, `server-index`, `keyring-*`, `oauth-credential-store` fail intermittently on head **and identically on base `4ee1a49`** under the same env, same test names, across repeated runs. A clean head run is `Test Files 85 passed (85) / Tests 1332 passed (1332)`, so the PR body's claim reproduces. `tsc --noEmit` is clean at HEAD.

Issue #95, opened 2026-08-09
Claude Code asks for authorization after the patch


artembaechka:
After applying the latest Claude Code patch, Claude Code asks for authentication again.

The terminal shows:
Not logged in · Please run /login

This happens even though the Claude Code proxy starts successfully and reports:
Proxy started on port 49653; Claude Code's Anthropic auth remains active.

## Reproduction
1. Start Claude Code with clodex.
2. Enter a simple message, for example `Hi`.
3. Claude Code responds with `Not logged in · Please run /login`.

The attached screenshot shows the proxy startup output and the subsequent `Not logged in` message.

<img width="3819" height="1461" alt="Image" src="https://github.com/user-attachments/assets/1cdf104b-d718-4e05-af99-8c74e6a7fe88" />

bman654:
I cannot repro this.  What clodex version?  What happens after you login?  Note even when you have model set to sol, Claude Code still makes calls to Anthropic models for some stuff.

artembaechka:
Clodex version - 2.30
Login redirects me to anthropic login page. I dont have account active subscription, so i cant login to claude code

bman654:
Thanks for the details — that explains it, and the patch isn't what's causing this.

Proxy mode (the default) deliberately keeps Claude Code's own Anthropic auth: it only reroutes requests for your OpenAI models and passes everything else through to Anthropic untouched. That's why it says "Claude Code's Anthropic auth remains active" — so without an Anthropic login, proxy mode can't work for you.

What you want is **endpoint mode**, where no Anthropic account is involved at all. clodex runs a local Anthropic-format gateway and launches Claude Code pointed at it, so every request (including the small background ones Claude Code makes for things like session titles and tool-use summaries) goes to your OpenAI model instead.

Try:

```bash
clodex claude --endpoint
```

Claude Code should start and ask you to approve a custom API key — that's clodex's local gateway token, not an Anthropic key. Approve it, and you should get a working session with no Anthropic login. You'll still need your OpenAI credentials (an OpenAI API key or a ChatGPT/Codex plan), but nothing from Anthropic.

If that works, please let us know — and specifically tell us **whether it asks you to approve that API key every time you restart Claude Code**. We think it might, because clodex currently generates a new gateway token on each run. If it does, that's something we can fix on our side by making the token stable, so it would be very helpful to hear.

A couple of extras:

- To make endpoint mode your default: `clodex claude --endpoint --save-mode`
[...]

PR #106, opened 2026-08-10
fix(env): bypass the proxy for the loopback gateway in endpoint mode

bman654:
## Summary

Endpoint mode launches Claude Code with `ANTHROPIC_BASE_URL` pointed at a local gateway, but `buildChildEnv` passes the user's `HTTP_PROXY`/`HTTPS_PROXY` straight through — `CONFLICTING_ENV_VARS` contains no proxy variables. Claude Code honours `NO_PROXY` but has no implicit loopback bypass, so it asked the corporate proxy to fetch `127.0.0.1` and a proxy that refuses answered with a bodyless 503.

This appends the loopback spellings, plus the gateway's own address, to the child's `NO_PROXY`/`no_proxy` whenever a proxy is configured.

Reported in #95, where the reporter confirmed a manual `no_proxy` resolved it.

## Why not just delete the proxy variables

The wrapper's endpoint branch (`computeWrapperEnv`) does exactly that, but it is the blunter option: the child still needs the proxy for its **own** outbound traffic — WebFetch, MCP servers, telemetry. On a network where direct egress is blocked, stripping the proxy trades one broken thing for another. This change is additive instead: existing entries are preserved and the proxy stays usable for everything that is not the gateway.

(The resulting inconsistency between the two launch paths is pre-existing and left alone here.)

## Behaviour is keyed to Claude Code's actual matcher, not to intuition

Three details were taken from the matcher in the 2.1.226 bundle rather than assumed:

- It reads `no_proxy || NO_PROXY`, so a non-empty **lowercase** value wins outright. The two casings are therefore *not* unioned — unioning would silently un-proxy hosts the user had bypassed in only the losing variable. Both casings are written to the same resolved value so neither can win stale.
- It treats `*` as bypass-everything **only when that is the entire value**; a `*` appearing as one entry in a list matches nothing. So `NO_PROXY=internal.corp,*` must not suppress this fix — an earlier revision of this change got that wrong and would have silently no-opped for those users.
- Its host comparison treats `localhost` and every `127.0.0.0/8` address as equivalent, so the explicit gateway entry is not needed *for Claude Code*. It is kept for exact-matching consumers such as `proxy-from-env`.

## Scope: loopback only

`buildChildEnv` is also called with a **remote** Anthropic-passthrough `baseUrl`. Bypassing the proxy for that host would point the child straight at a server the proxy may be the only route to, so the bypass only applies when the gateway is loopback. There is a regression test for this.

## Validation

- `pnpm test` — 85 files / 1370 tests passed
- `pnpm typecheck`, `pnpm build` — passed
- Mutation-tested: removing the call site, no-oping the helper, dropping the lowercase write, weakening the loopback predicate, dropping the wildcard guard, dropping dedupe, dropping the existing-entry preservation, dropping the proxy-presence guard, and reverting each of the two matcher-semantics fixes above are all caught by full-file runs.
- Tests manipulate real proxy environment variables, so they save and restore what they touch and pass both in a clean shell and one that already has `HTTPS_PROXY`/`NO_PROXY` set.

No credentials, prompts, or provider payloads are involved.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -60,6 +60,100 @@ export function applyClaudeCodeThirdPartyCompat(env: NodeJS.ProcessEnv): void {
   env['CLAUDE_CODE_SIMPLE_SYSTEM_PROMPT'] = '0';
 }
 
+/** Loopback spellings the child must reach directly even behind a proxy. */
+const LOOPBACK_NO_PROXY_ENTRIES = ['localhost', '127.0.0.1', '::1'] as const;
+
+/** Hostname of a base URL, with IPv6 brackets stripped; '' when unparseable. */
+function gatewayHostname(gatewayUrl: string): string {
+  try {
+    return new URL(gatewayUrl).hostname.replace(/^\[|\]$/g, '').toLowerCase();
+  } catch {
+    return '';
+  }
+}
+
+/** Loopback per RFC 5735/4291: `localhost`, all of 127.0.0.0/8, and `::1`. */
+function isLoopbackHost(host: string): boolean {
+  if (host === 'localhost' || host.endsWith('.localhost')) return true;
+  if (host === '::1' || host === '0:0:0:0:0:0:0:1') return true;
+  const v4 = /^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$/.exec(host);
+  if (!v4) return false;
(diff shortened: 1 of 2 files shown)


bman654:
adresses some pain uncovered in #95 .

PR #110, opened 2026-08-11
fix(models): separate curation and exposure caps

### LINKED ISSUE #107: Separate persisted favorites capacity from Claude-facing exposure
## Problem

The persisted favorites limit currently also acts as the limit for models exposed to Claude-facing catalogs. That couples two different concerns: users may want to curate more models than should be injected into any one Claude session, while raising the shared cap would enlarge every patched/server catalog.

## Proposed contract

- Allow a larger persisted curation set (the downstream implementation uses 100 entries) while keeping Claude-facing exposure capped at 20.
- Apply one ordered exposure projection across endpoint launches, proxy/server catalog routes, favorites-only mode, and patch configuration.
- Choose the 20 positions before resolving model availability, so an unavailable saved entry does not silently let a later favorite jump the user's order.
- Preserve the selected launch model in the exposed set without double-counting it when it is already saved.
- Keep saved configuration intact; report capacity-omitted entries explicitly and leave aliases to omitted entries inactive.

This should not change credentials, request translation, or inference protocol behavior.

## Evidence / readiness

A downstream implementation is already isolated in [zachspartofaday/clodex#28](https://github.com/zachspartofaday/clodex/pull/28). Its current head received a clean Codex review and passed the full Node 24 gate (`pnpm typecheck && pnpm test && pnpm build`: 92 test files / 1,462 tests), including focused coverage across catalog, favorites, HTTP routes, server filtering, and patch configuration.

The reproductions use synthetic catalogs only; no live provider request, credential mutation, or raw local log data is involved. If this contract fits upstream direction, I can reconstruct the change directly on current upstream `main` as a focused PR.

zachspartofaday:
Closes #107.

## Summary

- raise persisted favorites curation capacity to 100 while keeping Claude-facing exposure capped at 20;
- centralize saved-order exposure projection across endpoint launches, HTTP routes, favorites-only servers, and patch configuration;
- preserve a selected launch model within the 20 exposed slots without double-counting it;
- keep saved configuration untruncated, report exact capacity-omitted favorites, and leave aliases to omitted entries inactive;
- warn after a successful alias save when its target is outside the default active exposure window;
- make every capacity warning accurately describe selected-order behavior when unavailable entries leave fewer active models.

Capacity is selected before model resolution. An unavailable entry inside the first 20 therefore keeps its saved position instead of allowing a later favorite to silently jump the user's order. Removing or reordering unavailable entries reclaims positions.

## Stack topology

This source branch is rebuilt directly on the expected reviewed #100 head `a2deaa60ad652b998fcc3d083f27acb3048b1001`, not independently on `main`. Its focused #110 delta is exactly 24 files, 710 additions, and 288 deletions. Once #100 lands, GitHub will remove the ancestor commits from this PR comparison and expose only this focused delta.

The six logical #110 commits were reconstructed in order. Commits 2–6 are patch-equivalent to the prior fork-reviewed series; commit 1 differs only where current v2.5/PR #117/O2 documentation and provider-neutral CLI wording required reconciliation.

## Scope boundary

The cross-consumer distinction between an available alias target outside the exposure window and an unavailable/unsupported target is intentionally separate in #113. This PR preserves alias persistence, selected-order/no-backfill projection, and bounded warnings without partially implementing that diagnostic taxonomy.

No O3 behavior is included. The reviewed O2 catalog, credential-destination pinning, count-token routing, response identity/SSE behavior, and pricing are conserved. No package, lockfile, dependency, configuration, schema, or migration change is included.

## Review and validation

Exact rebuilt head `59158490462d5c7d186822b9b1e8436c4fdab51f` received a clean serialized fork review in [zachspartofaday/clodex#50](https://github.com/zachspartofaday/clodex/pull/50#issuecomment-5261048717). Its one first-round P2 was explicitly deferred to linked #113 by declared scope; round 2 reviewed the unchanged exact head clean.

Node 24.19.0 / pnpm 10.34.5:

- focused owning suites: 21 files / 443 tests passed;
- full suite: 93 files / 1,730 tests passed;
- `pnpm typecheck`: passed;
- `pnpm build`: passed;
- `git diff --check`: passed;
- clean/untracked check: passed.

The complete gate was repeated after the exact-head clean review. No live provider request or credential mutation was made.

## Safety

- synthetic fixtures and localhost test servers only;
- no credential, registry/account-selection, request-translation, adapter, OAuth/WebSocket, or trace/logging change;
- no favorites manager/reordering/truncation feature and no endpoint policy change;
- no package, lock, dependency, release, or generated `dist/` content committed;
- no raw local logs or sensitive runtime values used as evidence.


bman654 (changes requested):
# Review

Verdict: **CHANGES_REQUESTED** — one blocking test gap. The curation/exposure implementation itself held on the production-reachable edge cases I executed; the remaining non-blocking items are coverage, diagnostics, and follow-up UX issues.

## What I verified

- All four production consumers use the shared exposure projection: endpoint catalogs at `src/catalog.ts:139-143`, HTTP-proxy routes at `src/http-proxy/routes.ts:57-64`, favorites-only servers at `src/server/catalog-filter.ts:44-45`, and patch configuration at `src/patcher.ts:178-180`. `src/server/index.ts` no longer applies a separate slice.
- The 100/20 split is behaviorally pinned. Changing `MAX_MODEL_CATALOG` to `MAX_FAVORITES` at `src/constants.ts:49` produces nine failures across six focused test files; this does not rely on the patch-transform digest pin.
- Patch configuration is capped before the unchanged transforms consume it. `src/patch-transforms.ts:152` remains the single `MODEL_CONFIG` pass, and `src/patch-transforms.ts` is byte-identical to `ad907a5`, so `PATCH_TRANSFORMS_VERSION` correctly does not change. I also exercised 25 favorites across all 13 extracted Claude Code bundles (2.1.208 through 2.1.226): the first 20 identities were patched consistently and favorites 21–25 did not leak into the output.
- `computePatchConfigHash` follows the exposed set: changing which favorites are inside the first 20 changes the hash, while appending a 21st favorite does not. Existing configurations with at most 20 favorites therefore do not incur a gratuitous repatch.
- Saved configuration is not truncated. A 100-favorite configuration survived `models --list`, `server`, and `claude --dry-run`; an unrelated preference write under 2.4.0 also preserved all 100 entries.
- Aliases to omitted entries remain inactive on every production surface: HTTP proxy (`src/http-proxy/routes.ts:102-105`), patcher (`src/patcher.ts:189-198`), endpoint catalog (`src/catalog.ts:87-106`), and favorites-only server (`src/server/index.ts:478-505`).
- Endpoint launch-model reservation behaves correctly: a launch model in the first 20 is not double-counted; a later or unsaved launch model consumes one slot and leaves 19 favorite routes.
- Choosing positions before availability resolution is a deliberate behavior change, not a defect. Issue #107 and the PR body both require an unavailable entry inside the window to retain its saved position rather than backfilling from later favorites.
- The focused suite reproduces at 9 files / 171 tests, the committed full suite at 85 files / 1,382 tests, and typecheck and build are clean.

## Blocking ask

**Pin “capacity is selected before model resolution” in the HTTP-proxy consumer.**

`src/http-proxy/routes.ts:57` computes the projection and line 64 iterates `projection.exposedFavorites`, replacing the base loop that continued through saved favorites until it had 20 successful routes. That ordering is a stated acceptance criterion, but the new case at `tests/http-proxy-routes.test.ts:184-219` uses 25 resolvable favorites. Pre-resolution selection and post-resolution backfilling therefore produce the same output for its fixture.

I restored the base loop shape while retaining the new omission projection and ran the complete committed suite under the same dead-proxy environment as the clean head. All **85 files / 1,382 tests passed**. A discriminating harness then failed under that mutation and passed on this head.

This is production-reachable. With more than 20 saved favorites and favorite #1 unavailable, the mutated loop backfills favorite #21 into `routes` while `capacitySkippedFavorites` also reports favorite #21 as omitted. `loadHttpProxyRoutes()` reaches this through proxy startup, `clodex models --list`, and proxy-mode dry runs; proxy is the default bridge mode. The server consumer already pins the corresponding ordering at `tests/server-catalog-filter.test.ts:62-75` by placing an unavailable favorite inside a three-position window.

**Minimum change to unblock:** add a case to `tests/http-proxy-routes.test.ts` with more than `max` favorites and one unavailable entry inside the projected window. Assert that the later favorite is absent from `routes`, appears only in `capacitySkippedFavorites`, and is never both routed and reported omitted. For example, with `max: 3` and saved order `[gone, one, two, three]`, the result must contain routes `[one, two]`, `unavailable: [gone]`, and `capacitySkippedFavorites: [three]`.

## Optional, non-blocking follow-ups

1. **OPTIONAL — pin the launch-model reservation handoff.** `src/catalog.ts:141` passes `reservedFavorite: startingFavorite`; replacing it with `undefined` leaves the complete committed suite green. `tests/catalog.test.ts:43-65` uses a starting favorite absent from the saved list, while `tests/favorites.test.ts:109-118` tests de-duplication only at the helper boundary. Add a `buildCatalogRoutes` case where the starting model is one of the first saved favorites, asserting 20 distinct routes and that omission begins with favorite #21. The reachable regression is limited to interactive endpoint switch-menu mode with the launch model among favorites #1–#19 and enough resolvable favorites to fill the catalog.

2. **OPTIONAL — pin the patch-command capacity warning.** Deleting the new warning at `src/patcher.ts:926-935` leaves both full patcher test files green (119/119). `tests/patcher-command.test.ts` already executes `runPatchCommand`, captures `p.log.*`, and has a six-line `saveFavorites()` helper at `tests/patcher-command.test.ts:107-113`, so a 21-favorite assertion is cheap. The endpoint warning lives in the interactive `runClaudeCommand` path, which this repository explicitly verifies manually; no automated test currently calls that command. The PR did correctly pin the real server command at `tests/server-index.test.ts:255-282` and the HTTP reporting helper at `tests/http-proxy-index.test.ts:280-307`.

3. **OPTIONAL — use `target-not-exposed` consistently.** The PR adds that reason at `src/model-aliases.ts:30` and `src/model-aliases.ts:100-101`, but only the patcher emits it (`src/patcher.ts:189-198`). Endpoint and server catalog paths still report the generic `target unavailable` (`src/catalog.ts:106`, `src/server/index.ts:500-502`), and the HTTP-proxy report does the same. A missing provider and a saved target outside the active window have different remedies.

4. **OPTIONAL — warn when saving an alias outside the patch/proxy window.** `src/cli.ts:687-704` checks only that the target is a saved favorite and then reports success. An alias beyond the first 20 remains saved but inactive until enough earlier favorites are removed; the save command gives no indication of that state. Reuse `describeModelAliasRejection('target-not-exposed')` as a warning, not an error.

5. **OPTIONAL — make under-filled-window diagnostics accurate.** The capacity warnings at `src/http-proxy/index.ts:112-121`, `src/cli.ts:1288-1297`, `src/server/index.ts:428-436`, and `src/patcher.ts:926-935` say the first saved favorites “remain active.” That is false when unavailable entries occupy positions inside the selected window. The cap still explains why later entries are omitted, but the message should also say when earlier unavailable entries leave active slots unused and that removing or reordering those entries is the remedy.

6. **OPTIONAL / follow-up — expose ordering and the boundary in the favorites manager.** `src/cli.ts:781-803` displays `Favorite Models (n/100)` with no indication of which entries are inside the first 20, and the available operations only append (`src/favorites.ts:57-65`) or remove. Promoting favorite #37 while preserving the rest requires removing earlier entries and re-adding them. `models --list` and every **favorites-only** server start can also print an 80-entry warning for a full configuration; mark the exposure boundary and truncate long warning lists with a remaining count.

7. **OPTIONAL / follow-up — document or reconcile the pre-existing endpoint/patch window mismatch.** The patcher exposes favorites #1–#20 with no reservation (`src/patcher.ts:178`), while an endpoint switch-menu launch with a separate starting model exposes that model plus favorites #1–#19 (`src/catalog.ts:139-143`). Consequently, an alias for favorite #20 can be baked into Claude Code but be unavailable from that endpoint catalog. This already occurred with a full 20-favorite configuration on base, so it is not introduced by this PR; it is reachable when a user combines `clodex patch` with endpoint switch-menu mode.

8. **NIT — duplicate saved favorites consume exposure positions before surface-specific de-duplication.** `src/patcher.ts:178` and `src/http-proxy/routes.ts:57` project first, while the later config/route builders de-duplicate. The UI prevents duplicates at `src/favorites.ts:62`, so this requires a hand-edited configuration and is hygiene only.

9. **NIT — `buildFavoritesList` has no production caller.** The PR changes `src/favorites-resolver.ts:73-127`, but references in `src/` are limited to its definition and the explanatory comment at `src/catalog.ts:123`. If it is not going to become a fifth production consumer, it is a deletion candidate rather than another projection path to maintain.

Once the HTTP-proxy ordering test lands, this is ready to merge.

zachspartofaday:
Resolved fork-first in [zachspartofaday/clodex#45](https://github.com/zachspartofaday/clodex/pull/45), which received a clean serialized review at exact source head `0a0c2a4f91adde0295270f65b6d132418f07c93d` before the fork merge advanced this PR to `1998875aab169e3fba0b08c8acb2b390ed4d4727`.

Blocking ask:

1. **HTTP pre-resolution capacity ordering:** added the discriminating `max: 3`, `[gone, one, two, three]` row. It yields routes `[one, two]`, unavailable `[gone]`, capacity-skipped `[three]`, and proves no model is both routed and reported omitted.

Optional and nit items:

1. **Launch-model reservation handoff:** added the first-saved-favorite case; it produces 20 distinct routes with omission beginning at #21.
2. **Patch-command warning:** added a 21-favorite `runPatchCommand` row and pinned the corrected selected-order semantics.
3. **`target-not-exposed` consistency:** routed to [#113](https://github.com/bman654/clodex/issues/113) because unavailable targets and available targets beyond the exposure window need distinct reasons across all consumers; this PR continues to reuse the existing taxonomy rather than introduce a partial distinction.
4. **Alias save outside the window:** save/persistence still succeeds, followed by an existing-taxonomy warning; command coverage pins exit success, persisted alias, and warning.
5. **Under-filled-window accuracy:** CLI, patch, HTTP-proxy, and server warnings now explain that capacity is selected from saved order before availability/support checks, unavailable entries can leave fewer active models, and removing/reordering them reclaims positions.
6. **Favorites manager boundary/order UX:** routed to [#114](https://github.com/bman654/clodex/issues/114) for exposure-boundary display, intentional reorder behavior, and bounded long omission warnings.
7. **Endpoint/static patch mismatch:** routed to [#115](https://github.com/bman654/clodex/issues/115) as a pre-existing cross-consumer policy decision.
8. **Duplicate saved favorites:** no code change. The supported production writer rejects duplicates; the premise requires hand-edited configuration and does not justify normalization in this PR.
9. **Unused `buildFavoritesList`:** deleted the zero-caller helper, associated option type/tests, stale imports, and explanatory references; zero source/test references remain.

Validation with ambient CLodex/proxy variables removed: focused `7 files / 98 tests passed`; full `85 files / 1,381 tests passed`; typecheck, build, and diff-check passed. Targeted falsifiers cover the ordering, alias-warning, and warning-wording branches. No live provider request, credential mutation, or raw-log publication occurred.

zachspartofaday:
#110 has been transplanted onto current `main` after #100 merged and the v2.6.0 release landed.

- current head: `c59ea880eba78807216b66e824fa2d780c068a39`
- base: v2.6.0 `e05e202fdc55ec8bec1fbf0fd3df44487b3fb12b`
- all seven commits remain patch-equivalent to the previously reviewed series according to `git range-diff`
- Node 24.19.0 / pnpm 10.34.5 validation passed: typecheck, 94 test files / 1,809 tests, and build
- no release metadata, dependency, lockfile, CI, or generated-data changes

This PR is now ready for maintainer review. Per the dependency order recorded on #100, #113 remains draft and will be transplanted onto the resulting `main` only after #110 merges; #101 follows #113.

bman654 (changes requested):
Round-2 review at `c59ea88`, against `main` @ `e05e202` (v2.6.0).

Thanks for the transplant note — the head/base/commit facts in it all check out, and your
`94 test files / 1,809 tests` matches what I measured independently, three times.

## The round-1 ask is closed, and it is closed the right way

`tests/http-proxy-routes.test.ts:222-259` is the requested fixture exactly, including the third
assertion that no entry is both routed and reported omitted. I did not take that on faith — I
restored the pre-PR resolve-then-cap loop in `src/http-proxy/routes.ts:64` and ran the full suite:

```
 Test Files  1 failed | 93 passed (94)
      Tests  2 failed | 1807 passed (1809)
```

Only the tests the ask was filed for. That is discrimination, not assertion. The reordering work in
this round is also genuinely load-bearing: reverting the retention-filter move in
`buildDesiredPatchConfig` reds exactly the two patcher tests that own it and nothing else, and it is
what makes `clodex patch` and the running gateway expose the same set — the "the alias is the model
identity" invariant in CLAUDE.md. Good change.

## The one blocking ask

**B1 — pin "capacity is selected before route resolution" in the endpoint consumer,
`buildCatalogRoutes`.**

This is the same shape as the round-1 ask, at the consumer this round newly rewired. The guarantee is
stated in your own new doc comment at `src/catalog.ts:120-127` — *"a favorite that fails resolution
still consumes its selected window slot — the window is never backfilled with later favorites"* — and
in the CLAUDE.md invariant this PR edits. Nothing holds it.

One token reverts it to `main`'s semantics, and the full suite does not notice:

```
src/catalog.ts:146   const tail = projection.exposedFavorites  →  const tail = favorites
 Test Files  94 passed (94)
      Tests  1809 passed (1809)          ← zero red
```

Run side by side, the contrast is the whole argument: that identical mutation shape applied to
`src/http-proxy/routes.ts` reds 2 tests, because you pinned it there.

Both existing cases in the `buildCatalogRoutes` block supply a resolver that returns a route for
*every* favorite (`tests/catalog.test.ts:49-54` is unconditional; `:76-85` never returns `undefined`),
so `projection.exposedFavorites` and `favorites` cannot produce different output. That is the same
all-resolvable-fixture defect round 1 flagged on the HTTP-proxy side.

Two things stop this being hygiene. `main` really does resolve-then-cap
(`const tail = favorites`, verified by reading `main`'s builder), so this PR **changes** endpoint
behavior rather than inheriting posture. And it is reachable through supported commands only: save
favorites with `clodex models`, remove a provider with `clodex providers remove` — which does not
rewrite `favoriteModels` — then launch interactively in endpoint mode. Under the reverted semantics a
later favorite is routed into `/model` while the launch warning still names it as skipped.

What would close it: one behavioral case in `tests/catalog.test.ts` with a favorite that fails
resolution *inside* the selected window — e.g. favorites `[{missing,gone}, {main,one}, {main,two},
{main,three}]`, `max: 3`, an explicit fifth-argument `startingFavorite` that is not itself saved, and
a resolver returning `undefined` for `missing:gone`. Assert exact ordered `routes` (so `main:two` is
visibly not backfilled), `droppedFavorites`, `capacitySkippedFavorites`, and no overlap between the
last two and `routes`. A 21-favorite case at the real `MAX_MODEL_CATALOG` is stronger if you prefer
it. Acceptance criterion: with that case present, `const tail = favorites` must go red in a full run.

This is a test-only change. `tests/catalog.test.ts` already owns this pure builder, so it is not the
interactive `runClaudeCommand` seam this repo verifies manually — that exclusion does not apply here.

## Worth doing in the same commit

**Every one of the five new warnings is unpinned in the over-scope direction.** Widening a guard
ships green — a regression printing `0 saved favorites not exposed...` on every proxy startup would
pass CI:

```
src/http-proxy/index.ts:112   if (true || loaded.capacitySkippedFavorites.length > 0)
      Tests  1809 passed (1809)     ← green
src/http-proxy/index.ts:112   if (false && ...)
      Tests  1 failed | 1808 passed  ← the under-scope direction IS pinned
```

`tests/http-proxy-index.test.ts` already has three fixtures with `capacitySkippedFavorites: []`, and
`toHaveBeenCalledWith` does not fail on an extra call. One
`expect(warn).not.toHaveBeenCalledWith(expect.stringContaining(...))` per zero-skipped fixture covers
the HTTP, server, patcher and alias-save guards; the endpoint guard at `src/cli.ts:1317` needs the
switch-menu seam and is fine to leave.

## Non-blocking notes

1. **A correction we owe you, not a request.** All four capacity warnings advise *"Removing or
   reordering those entries reclaims positions"*, and clodex offers no reorder affordance —
   `src/favorites.ts` exports only `addFavorite` (appends) and `removeFavorite`. That wording came
   from **our own round-1 review**, so it is ours to fix, not a defect in your submission. If you are
   touching the strings anyway: *"Remove entries you no longer need, or edit the order of
   `favoriteModels` in `~/.clodex/config.json`."*

2. **`clodex patch` is the only Claude-facing surface that never names a favorite it dropped as
   unavailable.** `src/patcher.ts:183` filters and discards, while the server, HTTP-proxy and launch
   paths all warn by name. What changed this round is that such an entry now also *burns a catalog
   slot*, and your new remediation sentence is written about exactly those entries. A single
   `unavailableFavorites` field out of `buildPatchModelConfig` would close it. Not blocking:
   reachability needs a saved favorite on a retained OpenCode Go identity whose id has left the
   committed allowlist, and all 17 committed ids were still live when checked on 2026-08-13.

3. **A fully-unavailable window dead-ends `clodex patch` with advice that cannot help.** The
   `Object.keys(desired.config).length === 0` early return at `src/patcher.ts:962-966` sits *above*
   the capacity warning, so a user with 25 favorites whose first 20 are unavailable is told to save
   favorites — and new favorites append past position 20. Hoisting the warning above the early return
   is two lines.

4. **`filterServerModelsByFavorites` was silently re-ordered.** Catalog-order became favorites-order
   (`src/server/catalog-filter.ts:15-32`), which changes the order of `GET /anthropic/v1/models` and
   flips which provider wins a bare-model-id collision in `createGatewayModelCatalog`. It affects
   users with 20 or fewer favorites — the population otherwise untouched by this PR. Reverting just
   that function body reds 2 tests, so it is severable. It looks like an improvement and I would keep
   it; please just say so in the PR body.

5. **The alias-save warning measures a 20-slot window; an endpoint launch exposes 19.**
   `src/cli.ts:711` passes no `reservedFavorite`/`reservedSlots`, while `src/catalog.ts:140-144`
   always reserves one. Favorite #20 therefore saves an alias with no warning yet is inactive in an
   endpoint launch whose starting model is not itself a favorite. The user is still told at launch,
   and erring generous is right for proxy mode and `clodex patch` — so this is wording, not routing.

6. **`target-not-exposed` reaches 2 of the 5 code paths.** `src/cli.ts:717` and `src/patcher.ts:200`
   emit it; `src/catalog.ts:106`, `src/server/index.ts:503` and `src/http-proxy/index.ts:136` still
   say `target unavailable`. For an alias on a resolvable favorite that merely sits past position 20,
   `clodex models --alias` says "outside the active catalog" while a launch says "unavailable", which
   reads as "your model is broken". Deferring the full taxonomy to #113 is defensible; this is just
   the seam a user meets first.

7. **At 100 favorites the capacity warning is an unbounded wall.** All four sites `.join('\n')` the
   full skipped list uncapped; `clodex claude --dry-run` at 100 favorites prints 116 lines, 101 of
   them model ids. `src/parent-notice.ts` bounds its notices for the same reason — naming ~10 and
   summarising the rest would match.

8. **The curation UI gives no signal of the exposure boundary it now spans.** The header reads
   `Favorite Models (100/100)` and rows 1-20 render identically to 21-100, though the manager already
   marks a different failure mode (`★ <id> — provider gone`). #114 covers this; noting it because the
   cap raise is what makes it bite.

9. **Two small ones.** `.claude/docs/patcher.md` gained no line for the new ordering contract, though
   CLAUDE.md routes anyone changing `clodex patch` there. And the new duplicate guard in
   `filterServerModelsByFavorites` is necessary — the flatMap rewrite newly makes duplicate output
   possible — but unpinned; a two-favorite duplicate row would hold it.

10. **The body's stats section predates the transplant** (`710 additions, 288 deletions`, "six
    logical commits"; it is now 803/295 over seven). Your follow-up comment already carries the
    correct test numbers, so this is just the body catching up.

## Checked and cleared — please do not re-derive these

- **Users with 20 or fewer favorites are unaffected.** `clodex models --list` and
  `clodex claude --dry-run` are byte-identical between `main` and head, and a differential against a
  verbatim copy of `main`'s `patcher.ts` at 1/5/19/20 favorites deep-equals `config`,
  `unknownWindows`, `rejectedAliases` and `rejectedAliasRejections`. The only intended differences are
  note 4 and note 6.
- **No spurious re-patch on upgrade, and no stale-patch hole.** Because `config` is unchanged at ≤20,
  `computePatchConfigHash` is unchanged and an existing install stays `current`; meanwhile reordering
  a beyond-window favorite into the window, an in-window favorite going unavailable, and an alias
  crossing the boundary each correctly produce `stale-config`.
- **`PATCH_TRANSFORMS_VERSION` correctly stays at 6** — `src/patch-transforms.ts` is byte-identical
  to `main`. A bump would have forced a needless re-patch on every install.
- **Capping exposure does not desynchronize the patch sites.** The real `applyClodexPatches` over the
  extracted 2.1.226 and 2.1.228 bundles with 25 favorites and aliases at positions 0/19/20/24: every
  exposed identity present in the Agent-tool enum and the known-alias validator, every aliased
  identity with its self-mapping and picker row, leak-checked by baseline hit-count against the
  unpatched source rather than a bare `not.toContain`.
- **The starting-model reservation is correct at the wire**, not just in the returned array: a real
  listener answering `GET /v1/models` returns 20 unique ids for all four saved positions of the
  starting model (absent, first, middle, past position 20), with the starting model present exactly
  once and never also in `capacitySkippedFavorites`.
- **Nothing is both exposed and reported capacity-omitted, and no surface backfills**, proved by
  differential across the catalog, HTTP-proxy, server and patcher builders.
- **`buildFavoritesList` really was dead code.** Zero hits for the value tree-wide at head; at the
  merge base only its definition, its own tests, and two comments this PR rewrites. The coverage its
  tests owned — the empty-API-key guard and the `authType: 'none'` exemption — is live at
  `src/http-proxy/routes.ts:79` and separately pinned, each proved by its own mutation.
- **Saved configuration is never truncated** — one writer, and the projection is pure
  `filter`/`slice`.
- **Aliases to omitted targets are genuinely inactive everywhere, persist, and reactivate on
  reorder.** On the default proxy path the name fails closed with a local 400 rather than reaching
  Anthropic. Your new `target is outside the active Claude Code catalog` wording is also strictly
  better than `main`'s, which said `target is not a saved favorite` — flatly false for a saved
  favorite the availability filter had dropped.
- **The new warnings are not swallowed by the launch-time stdout mute.** Every one is emitted before
  `launchClaude` installs it; confirmed by a real `clodex claude -- --model haiku -p ...` smoke with
  25 saved favorites, which printed the full capacity block and then the model's reply.
- **The manager stays usable at 100 favorites** — the prompt library clamps and pages a 102-option
  select; `config.json` is 6,003 bytes at 100 vs 1,363 at 20.

Some concerns we raised internally did not survive checking, and it is only fair to say so: the
patcher's narrower availability predicate and the `claude-*` id collision in the endpoint catalog are
both identical on `main`; duplicate saved favorites are reachable only by hand-editing `config.json`;
and the endpoint switch-menu coverage gap is pre-existing and manual-only by repo policy.

One caveat we could not close: nobody exercised a live translated-provider request or a real patched
binary at runtime, so the patch evidence above is static analysis over real bundles plus unit
execution, not a launched patched `claude`.

Everything else here is a follow-up. B1 is the only thing blocking.

PR #111, opened 2026-08-11
fix(transport): clarify usage-limit semantics

zachspartofaday:
Follow-up to #96.

## Summary

- correct the durable mechanism narrative: mapping `usage_limit` does not save retry attempts because both 429 and 5xx remain retryable
- retain the ordinary 60-second retry-after cap while omitting oversized plan-level reset claims rather than falsely reporting 60 seconds
- let structured `usage_limit` override generic 5xx classification while preserving explicit deterministic 4xx authority
- normalize finite numeric WebSocket diagnostic identifiers to bounded strings
- pin HTTP and WebSocket status/retry/header/body/log/summary behavior, including output-less terminal summaries and both forms of `previous_response_not_found`
- remove one proven redundant fallback without broadening unverified classifier rules

`resets_in_seconds` remains unchanged because no captured provider frame verifies that shape. No speculative chunk-discriminator rule was introduced.

For the current bundled client, omitting the misleading `retry-after: 60` leaves the configured retry count unchanged but restores its normal exponential spacing—about 1.125–1.5 seconds total for the default two retries instead of 120 seconds.

## Review provenance

This branch is based directly on current upstream v2.4.0 (`ad907a54`). Its complete diff remains exactly four files: `src/upstream-error.ts`, `src/oauth/responses-websocket.ts`, and their two direct test files.

The original fork review found the same plan-limit discriminator-loss family at two representation boundaries. The required whole-family exhaustion audit mechanically enumerated every `usage_limit` / `rate_limit_error` occurrence, retry field/prose parser, all six WebSocket synthetic-error construction routes, status-bearing/status-less terminal paths, and `data` / serialized-`responseBody` parser. The mechanically bounded class was exhausted with no unresolved members; the complete upstream-error file and targeted WebSocket transition logic were judgment-checked.

The maintainer review's four independently removable classifier branches and applicable optional edges were resolved fork-first in [zachspartofaday/clodex#44](https://github.com/zachspartofaday/clodex/pull/44). Exact source head `746c0c5115fc5f464f5086f3b052df91dd583512` received a clean serialized Codex result; the validated fork merge advanced this upstream PR to `76462b7f7ee3fe103a39c41c91846ca795354512`.

## Validation

Node 24.19.0 / pnpm 10.34.5 with ambient `CLODEX_*` and proxy variables removed:

- focused upstream-error/WebSocket suite: 2 files / 169 tests passed
- `pnpm typecheck`: passed
- `pnpm test`: 85 files / 1,391 tests passed
- `pnpm build`: passed
- `git diff --check ad907a5436a13cc807e988ad5b4dff0437483d36..HEAD`: passed

Targeted falsifiers confirmed the new tests fail if either non-negative retry-hint gate is removed. No existing assertion was weakened.

## Safety

- synthetic fixtures only; no live provider traffic, credential mutation, or raw-log publication
- no WebSocket continuation restructuring, post-output replay, or parallel classifier/redaction path
- no dependency, package, lock, release, registry/auth, generated-catalog, or committed `dist/` changes

### DIFF
--- a/src/oauth/responses-websocket.ts
+++ b/src/oauth/responses-websocket.ts
@@ -12,7 +12,12 @@ import type { FetchFunction } from '@ai-sdk/provider-utils';
 import type { RawData, WebSocket as WsWebSocket } from 'ws';
 import { CODEX_RESPONSES_WEBSOCKETS_BETA } from '../constants.js';
 import { outboundWsProxyAgent } from '../outbound-proxy.js';
-import { anthropicErrorType, clampRetryAfterSeconds, frameStatusCode } from '../upstream-error.js';
+import {
+  anthropicErrorType,
+  clampRetryAfterSeconds,
+  frameStatusCode,
+  MAX_RETRY_AFTER_SECONDS,
+} from '../upstream-error.js';
 import { sanitizeToolInput } from '../tool-input-sanitize.js';
 
 const RESPONSES_LITE_HEADER = 'x-openai-internal-codex-responses-lite';
@@ -874,8 +879,9 @@ function responseErrorMessage(event: unknown): string | undefined {
 }
 
 function boundedDiagnosticIdentifier(value: unknown): string | undefined {
-  if (typeof value !== 'string') return undefined;
-  const normalized = value.trim();
+  const identifier = typeof value === 'number' && Number.isFinite(value) ? String(value) : value;
+  if (typeof identifier !== 'string') return undefined;
+  const normalized = identifier.trim();
   return normalized && /^[a-zA-Z0-9_.:/-]+$/.test(normalized)
     ? normalized.slice(0, 128)
     : undefined;
@@ -1640,21 +1646,24 @@ function handleSocketMessage(entry: ConnectionEntry, data: RawData): void {
   }
 
   if (errorStatus !== undefined) {
+    const usageLimited = [responseErrorType(event), errorCode].some(value => (
+      typeof value === 'string' && value.toLowerCase().includes('usage_limit')
+    ));
+    // A generic 5xx field is only a fallback when a structured plan-limit class
+    // is present. Otherwise it masks the more specific 429 signal carried by the
+    // same frame. Explicit deterministic 4xx statuses remain authoritative.
+    const statusCode = usageLimited && errorStatus >= 500 ? 429 : errorStatus;
     // The AI SDK strips unknown frame fields, so a backoff hint survives only
     // baked into the message text — same reason the connection-limit branch
-    // above spells it out. Clamped, so a hostile hint cannot park a client past
-    // the 120s no-event stream abort.
-    // Only when upstream actually gave one. `clampRetryAfterSeconds` supplies a
-    // 5s DEFAULT for a missing hint, so clamping unconditionally would have
-    // every 429 assert a backoff upstream never stated — and that value becomes
-    // a real `retry-after` header downstream. Worse on a plan-level limit,
-    // where the reason says hours: a prose-only "retry after 1800s" would get
-    // "; retry after 5s" appended, and the client reads the first match.
-    const statedRetryAfter = errorStatus === 429 ? responseRetryAfterSeconds(event) : undefined;
-    const retryAfterSeconds = statedRetryAfter === undefined
-      ? undefined
-      : clampRetryAfterSeconds(statedRetryAfter);
-    const reason = responseErrorMessage(event) ?? `OpenAI rejected the request (HTTP ${errorStatus})`;
+    // above spells it out. Only forward a value upstream actually gave. An
+    // oversized ordinary rate-limit hint is capped, but omitting an oversized
+    // plan-limit hint is more honest than claiming the 60s cap is its reset time.
+    const statedRetryAfter = statusCode === 429 ? responseRetryAfterSeconds(event) : undefined;
+    const retryAfterSeconds = statedRetryAfter !== undefined
+      && (!usageLimited || statedRetryAfter <= MAX_RETRY_AFTER_SECONDS)
+      ? clampRetryAfterSeconds(statedRetryAfter)
+      : undefined;
+    const reason = responseErrorMessage(event) ?? `OpenAI rejected the request (HTTP ${statusCode})`;
     failContext(
       entry,
       ctx,
@@ -1669,10 +1678,10 @@ function handleSocketMessage(entry: ConnectionEntry, data: RawData): void {
         // file's diagnostics. The connection-limit branch can pass its code raw
         // only because it has just been compared `===` to a known constant.
         errorCode: boundedDiagnosticIdentifier(errorCode),
-        mappedStatusCode: errorStatus,
+        mappedStatusCode: statusCode,
         ...(retryAfterSeconds !== undefined ? { retryAfterSeconds } : {}),
       },
-      errorStatus,
+      statusCode,
       retryAfterSeconds,
     );
     return;
@@ -1720,19 +1729,20 @@ function handleSocketMessage(entry: ConnectionEntry, data: RawData): void {
     // `code: '401'` matches no rule, lands on the 500 default, and gets
     // rewritten as a retryable 502.
     const numericOrNamed = typeof details.errorCode === 'string' ? details.errorCode : undefined;
-    const classified = numericOrNamed !== undefined || discriminator
-      ? frameStatusCode(numericOrNamed, discriminator)
-      : 500;
+    const classified = frameStatusCode(numericOrNamed, discriminator);
     const statusCode = classified !== 500
       ? classified
       : settledReason ? 400 : 502;
     const usageLimited = statusCode === 429;
+    const planLimited = discriminator.includes('usage_limit');
     // Only when upstream stated one, and only baked into the TEXT: the AI SDK's
     // chunk schema is a closed zod object that strips `retry_after_seconds`, so
     // a hint carried only in the frame field never reaches the client. This is
     // the same reason the status-carrying branch above spells it out.
-    const retryAfterSeconds = usageLimited && responseRetryAfterSeconds(event) !== undefined
-      ? clampRetryAfterSeconds(responseRetryAfterSeconds(event))
+    const statedRetryAfter = usageLimited ? responseRetryAfterSeconds(event) : undefined;
+    const retryAfterSeconds = statedRetryAfter !== undefined
+      && (!planLimited || statedRetryAfter <= MAX_RETRY_AFTER_SECONDS)
+      ? clampRetryAfterSeconds(statedRetryAfter)
       : undefined;
     // The BOUNDED summary is the message, upstream's prose is not used at all.
     // A `response.failed` body can echo request content or backend detail, and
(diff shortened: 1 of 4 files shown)


bman654 (changes requested):
## Verdict: changes requested — one test-coverage ask

The implementation is correct as written, and it fixes the inaccurate retry-budget explanation inherited from #96 rather than merely rewording it. The replacement at `src/upstream-error.ts:183-186` matches the code: `frameIsRetryable` treats both 429 and every status at or above 500 as retryable (`src/upstream-error.ts:213-217`), so mapping `usage_limit` to 429 changes the downstream error class without saving an upstream attempt. The updated WebSocket test now pins both 429 and `isRetryable: true` (`tests/responses-websocket.test.ts:1417-1436`).

Validation on `1a84a12` also matches the PR body: the focused two-file suite is 157/157, `pnpm typecheck` is clean, the clean committed tree is 85 files / 1,379 tests, and `pnpm build` succeeds. The production diff is confined to the two classifier/error-transport files; continuation matching, head selection, transport replay safety, and eviction logic are unchanged from `ad907a5`.

---

## Blocking: four advertised classifier branches are independently unpinned

I snapshotted before each edit, deleted or weakened one branch, and ran the complete direct suites under an isolated `CLODEX_HOME` and dead ambient proxy:

`pnpm vitest run tests/upstream-error.test.ts tests/responses-websocket.test.ts`

Every mutation below still reported 2 files / 157 tests green:

| # | mutation that survives | production site |
|---|---|---|
| 1 | remove `error.code` from `apiErrorIsPlanUsageLimit` | `src/upstream-error.ts:83-87` |
| 2 | restore base's unconditional clamp in `providerErrorFrame` | `src/upstream-error.ts:272-281` |
| 3 | remove `responseErrorCode(event)` from `responsePlanLimitType` | `src/oauth/responses-websocket.ts:834-839` |
| 4 | widen the 5xx-only override to `usageLimited ? 429 : errorStatus` | `src/oauth/responses-websocket.ts:1657-1663` |

**Minimum change that unblocks this: add one behavioral row for each independently removable branch.** I executed all four proposed rows against HEAD, then repeated each feature deletion; each row failed only for its corresponding mutation.

1. **Code-only APICallError plan identity.** The installed OpenAI error schema permits `error.code` without `error.type`, and `apiErrorIsPlanUsageLimit` deliberately reads both. Add an `APICallError` with `data.error = { code: 'usage_limit_reached', message: '…', retry_after_seconds: 21600 }` and no `type`; `sdkUpstreamErrorDetails` must return 429 with no `retryAfterSeconds`. The existing fixtures provide the same plan identifier in both fields (`tests/upstream-error.test.ts:95-129`), so deleting the code arm is currently invisible.

2. **Raw post-output provider-frame omission.** Add a plain provider payload such as `{ type: 'usage_limit_reached', message: 'weekly limit reached; retry after 21600s' }`; `sdkUpstreamErrorDetails` must return 429 with no retry hint, while an otherwise equivalent ordinary `rate_limit_exceeded` control still clamps to 60. This is production-reachable, not merely a direct-call shape: after model output the AI SDK can enqueue a valid provider error payload, `generateAnthropicResponse(..., { forceStream: true })` rethrows that plain object (`src/sdk-adapter.ts:917-960`), and non-streaming OAuth requests deliberately collect that internal stream before writing the HTTP response (`src/proxy.ts:678-697`, `src/server/router.ts:458-466`). The value is then eligible for the downstream `retry-after` header (`src/proxy.ts:753-762`, `src/server/router.ts:497-507`). Generic post-output provider errors are present in the local production diagnostics; the exact plan-limit instance remains synthetic.

3. **Code-only WebSocket plan identity.** Use a schema-valid sequenced `response.failed` event whose `response.error` contains `code: 'usage_limit_reached'`, `message`, and an oversized retry field, but no `type`. That is the representation documented by this module (`src/upstream-error.ts:106-111`) and accepted by the Responses schema. Assert the synthetic wire frame keeps `type: 'usage_limit_reached'`, emits `code: '429'`, omits the hint, and classifies through the real SDK as retryable 429. Every current WebSocket plan-limit fixture puts the signal in `error.type`, so half of the new detector can disappear green.

4. **Explicit 4xx authority.** Add the missing conflict row: a valid `type: 'error'` event with top-level `status: 403` and a structured `usage_limit_reached` identifier must remain 403 and non-retryable. The production handler explicitly supports the Codex backend's top-level numeric status shape (`src/oauth/responses-websocket.ts:856-871`), and the nested identifier fields are normal Responses error fields; no test-only option or alternate catalog is required. I have not observed this exact contradictory combination in captured traffic, but it is the specific precedence guarantee the PR body advertises and the `>= 500` guard implements. The suite currently pins only the 500 conflict (`tests/responses-websocket.test.ts:1439-1453`).

These are coverage defects, not current runtime defects: HEAD passes all four behavioral probes. The issue is that four parts of the PR's stated contract can independently revert to the old or weaker behavior without making either direct suite fail.

---

## Checked and cleared

- **Finite numeric WebSocket identifiers are not a production blocker.** The base-to-head classifier change is real for a fake numeric `response.error.code`, but every relevant installed Responses event schema requires a string code, and the only production WebSocket caller is the fixed ChatGPT/Codex OAuth route (`src/provider-factory.ts:167-202`). In the current local census (220,394 diagnostic records, 161 `ws_response_error` records), every populated `errorCode` is a string and there are no `usage_limit` records. A numeric-code regression therefore requires the fixed backend to violate the current event schema. Optional defensive hardening at most.
- **The raw `planLimitType` propagation is identifier hygiene, not a disclosure or log-injection defect.** A malformed trusted-upstream identifier can reach the synthetic frame, but `JSON.stringify` escapes newlines. Ordinary inference logging omits `errorContent`; the explicit `CLODEX_LOG_REQUEST_PREVIEW=1` path compacts, redacts, and caps it at 2,000 characters (`src/trace-log.ts:529-546`). The separate proxy debug line is active only under trace logging (`src/proxy.ts:195-202`, `src/proxy.ts:740`).
- **#96's runtime guarantees remain intact.** `emptyFailureTerminal`, `transportReplaySafe`, and `retryTransportFailure` are unchanged. Context-length failures still map to 400 for auto-compaction; deterministic content-filter/max-output outcomes remain non-retryable 400s; bare 401 behavior is unchanged. Restoring the removed `numericOrNamed !== undefined || discriminator` fallback changes no output because `frameStatusCode(undefined, '')` already returns 500.
- **The new status-less `previous_response_not_found` row is load-bearing.** Removing `&& !willRetry` from the terminal gate fails that row; it is not padding.
- **An omitted hint is not recreated downstream.** Every consumer of `retryAfterSeconds` checks for `undefined` before setting a header (`src/proxy.ts:753-755`, `src/server/router.ts:505`, `src/server/router.ts:681`); nothing reclamps a missing value.

---

## OPTIONAL (none of these blocks the merge)

- **OPTIONAL — document the client-spacing consequence of omitting the header.** Claude Code's bundled Anthropic SDK defaults to two retries. I executed its extracted retry calculation: without `retry-after`, the two sleeps are 375–500 ms and 750–1,000 ms; with the former `retry-after: 60`, they are 60 seconds each. The configured retry count is unchanged, but the default total wait drops from 120 seconds to 1.125–1.5 seconds, so attempts arrive roughly 80–107 times sooner. That may be the right trade—60 seconds was not the six-hour reset—but it is user-visible and worth one sentence in the PR description.
- **OPTIONAL — bound only the emitted plan identifier, not the detector.** `responsePlanLimitType` currently forwards the matched trusted-upstream string verbatim. If you apply `boundedDiagnosticIdentifier`, keep raw substring detection separate and use a canonical bounded fallback for the wire type; using the rejected/undefined bounded value as the detector would silently restore the false 60-second hint for an exotic but recognizable plan-limit identifier.
- **OPTIONAL — make the serialized-body precedence explicit.** `numericRetryAfterSeconds` now reads the parsed `responseBody` fallback before the explicit `retry-after` header (`src/upstream-error.ts:52-80`). For a schema-invalid body with `retry_after_seconds: 21600` and header `30`, the body wins and a plan limit emits no hint; base forwarded 30. Pin that precedence if intentional. Also correct the new test comment at `tests/upstream-error.test.ts:111-116`: its fixture has the string `error.message` required by the installed OpenAI error schema, so the normal failed-response handler would populate `data`; the fallback is useful for malformed/non-conforming bodies, not that fixture as written.
- **OPTIONAL — reject negative numeric hints before clamping.** `responseRetryAfterSeconds` accepts `-1`, the plan guard treats it as within the cap, and `clampRetryAfterSeconds(-1)` fabricates the 5-second default. This is pre-existing, but `statedRetryAfter >= 0` in both WebSocket gates would align this edge with the PR's hint-honesty goal.
- **OPTIONAL (nit) — the plan bound is applied before rounding.** A numeric plan hint of `60.4` is omitted, while an ordinary hint of `60.4` is rounded to 60. This may be the intended raw-value policy; a small test would make that choice explicit.
- **OPTIONAL (nit) — document the dual meaning of synthetic `type` and `code`.** An explicit 403 plan-limit frame intentionally emits `{ type: 'usage_limit_reached', code: '403' }`: the type preserves plan identity while the code preserves authoritative HTTP status. Downstream classification is correct, but the two fields can look contradictory without that convention stated.

Validation limit: no live provider `usage_limit` event or manual `clodex claude -p` smoke was run; all plan-limit frames remain synthetic. The real WebSocket handler, AI SDK parser/classifier, full direct suites, typecheck, full test suite, build, and feature-deletion mutations were executed.

Once the four rows land, this is good to merge.

zachspartofaday:
Resolved fork-first in [zachspartofaday/clodex#44](https://github.com/zachspartofaday/clodex/pull/44), which received a clean serialized review at exact source head `746c0c5115fc5f464f5086f3b052df91dd583512` before the fork merge advanced this PR to `76462b7f7ee3fe103a39c41c91846ca795354512`.

Blocking coverage:

1. **Code-only APICallError identity:** added a no-`type` `error.code=usage_limit_reached` row; it returns retryable 429 with no oversized retry hint.
2. **Raw provider-frame omission:** added a raw plan-limit payload plus an ordinary `rate_limit_exceeded` control; the plan hint is omitted while the ordinary hint still caps at 60.
3. **Code-only WebSocket identity:** added a sequenced schema-valid `response.failed` row; synthetic type remains `usage_limit_reached`, code is 429, the hint is omitted, and the real SDK classifies it retryable.
4. **Explicit 4xx authority:** added the 403 conflict row; it remains 403/non-retryable while retaining plan identity.

Optional items:

- Updated the PR body with the bundled client's retry-spacing consequence; retry count is unchanged.
- Kept raw plan-identifier detection: JSON serialization and bounded trace diagnostics already contain the trusted-upstream value, while bounding the detector would create false negatives.
- Pinned serialized-body precedence over a conflicting header and corrected the fallback comment.
- Rejected negative hints before clamping in both WebSocket terminal paths.
- Pinned the raw `60.4` plan bound and ordinary rounding control in both paths.
- Documented synthetic `type` (plan identity) versus `code` (authoritative HTTP status) at the construction site.

Validation with ambient CLodex/proxy variables removed: focused 169/169; full 85 files / 1,391 tests; typecheck, build, and diff-check passed. The new negative-hint tests fail if either non-negative gate is removed. No live provider request or automated `claude -p` was run.

PR #126, opened 2026-08-18
fix(patch): restore patching on Claude Code 2.1.233 so model aliases and effort settings work

JohanCodinha:
`clodex patch` fails on every run under Claude Code 2.1.233, so anyone on that version loses their model aliases, per-model context windows and effort settings. This makes patching work again. Nothing is damaged meanwhile — the failed patch rolls back and the binary stays byte-identical to the pristine backup.

**Reachability.** Every patch attempt on every 2.1.233 native install, default path, no special configuration.

**Root cause.** To patch, clodex temporarily renames a module inside the binary and renames it back afterwards, then refuses to publish if the temporary name survives anywhere in the file. On 2.1.233 a repack leaves an inert copy of that name elsewhere in the file, so the refusal fires on a binary that is actually correct.

**Change.** The real name is now written over every copy found, then a re-scan proves none is left; both names are the same length, so each is an in-place overwrite. `shimEntryModuleName` declines a binary that already carries the marker, which is what makes rewriting sound — no clodex-produced state reaches that, so it is hardening rather than a fix for a reachable failure. `.claude/docs/patcher.md` covers why this is not weaker than refusing.

**Left out.** Bumping tweakcc to 4.3.3, which removes the need for the rename entirely: `minimumReleaseAge` refuses it until 2026-08-23, and alone it does not fix this — with 4.3.3 swapped into 2.6.0's `node_modules` the failure reproduces unchanged. Deleting the rename is the follow-up.

**Tests.** Five wrong implementations were written and each confirmed to turn the tests red: no sweep, rewriting only the first copy, scanning only the tail, dropping the read-boundary offset, no guard. The fixture is sized so the last two are reachable at all — a 20 MB binary with an orphan copy beyond the tail-scan window and another straddling the 8 MiB read boundary.

**Evidence.** Real 2.1.233 installs that reproduce the failure, on Linux (Ubuntu 26.04 x86_64, Node 22.22.1, tweakcc 4.3.0 unbumped) and macOS (arm64, signed Mach-O): 11 patches applied on both, no leftover copy in the result, `--model luna` and `--model haiku` answer on Linux, and the re-signed Mach-O passes `codesign --verify` and runs.

**Not verified.** Windows. Whether the survivor is genuinely dead space — only that no module entry references it. Other Claude Code versions; only 2.1.233 was reproducible here.

**Failure and rollback.** Unchanged: a failed restore throws, the candidate is discarded before the install is touched, and `--restore` still returns the pristine binary. Cost: the guard reads the whole binary once per rename.

### DIFF
--- a/.claude/docs/patcher.md
+++ b/.claude/docs/patcher.md
@@ -35,12 +35,15 @@ capabilities/defaults, and child-command network isolation.
 
 ## The entry-module shim (`bun-entry-module.ts`)
 
-tweakcc finds the module holding the bundle **by name**, and Claude Code 2.1.231 renamed it from
+tweakcc finds the module holding the bundle **by name**, and Claude Code 2.1.229 renamed it from
 `/$bunfs/root/src/entrypoints/cli.js` to `/$bunfs/root/cli`, which none of tweakcc's six accepted
-names match — so `readContent` threw and 2.1.231 could not be patched at all.
-tweakcc **4.3.3** added `/cli` to that list, so the shim is unnecessary from that version on;
-`.claude/docs/claude-code-internals.md` has the bundle-side detail. **Delete this shim once the
-tweakcc pin reaches 4.3.3, or once tweakcc identifies the module by `entryPointId`.**
+names match — so `readContent` threw and 2.1.229 and later could not be patched at all. (2.1.228 is
+the last `discoverable` release; 2.1.230 was never published.)
+tweakcc **4.3.3** added `/cli` to that list, but bumping the pin alone changes nothing: clodex
+mirrors the accepted-name list in `tweakccRecognizesModuleName` and it has no `/cli`, so the shim
+keeps firing until it is deleted. `.claude/docs/claude-code-internals.md` has the bundle-side
+detail. **Delete this shim once the tweakcc pin reaches 4.3.3, or once tweakcc identifies the module
+by `entryPointId`.**
 
 The name is used for identification only, so `shimEntryModuleName` swaps it for a stand-in of
 **identical byte length** (`/clodex--/claude` for a 16-byte original) that tweakcc does recognize,
@@ -75,7 +78,8 @@ tweakcc's own repack reads back as an ordinary module name.
   misparse returns null — leaving tweakcc's own error — instead of overwriting sixteen bytes at a
   guessed offset.
 - **More than one trailer can be in the file, so the scan validates rather than trusting position.**
-  Repacking is not size-neutral (an identity repack of 2.1.231 *shrinks* the blob by 61 bytes) and
+  Repacking is not size-neutral (an identity repack of 2.1.231 on Mach-O *shrinks* the blob by 61
+  bytes; ELF relocates instead, see below) and
   the replacement section content is written over the old, so the previous blob's trailer survives at
   a **higher** offset. Real binaries also carry a decoy `---- Bun! ----` around 55 MB — Bun's runtime
   ships the literal in `__TEXT`. Candidates are therefore tried from EOF backwards and the first one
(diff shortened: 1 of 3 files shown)


bman654 (approved):
Thank you — this is a real fix for a real, unreported breakage, and the diagnosis of the mechanism
is correct. I reproduced the failure and the fix against real binaries, and verified the published
artifact actually runs. My notes are almost entirely about **scope**: the change is right, and the
claims around it are wider than what happened.

## Verified

Probe = shim → tweakcc `readContent` → restore → shim → `writeContent` → restore, run under each
worktree's own `src/`, against real 2.1.233 platform tarballs.

| package | format | pristine | after repack | `main` | this PR |
| --- | --- | --- | --- | --- | --- |
| darwin-arm64 | Mach-O | 306,981,408 | 306,980,112 | OK | OK |
| darwin-x64 | Mach-O | 315,654,448 | 315,653,152 | OK | OK |
| win32-x64 | PE | 320,400,544 | 320,400,023 | OK | OK |
| win32-arm64 | PE | 308,298,912 | 308,298,501 | OK | OK |
| linux-x64 | ELF glibc | 324,598,064 | 770,234,672 | **THREW** | OK |
| linux-arm64 | ELF glibc | 321,771,752 | 769,779,944 | **THREW** | OK |
| linux-x64-musl | ELF musl | 318,721,832 | 770,240,296 | **THREW** | OK |
| linux-arm64-musl | ELF musl | 314,916,456 | 769,805,928 | **THREW** | OK |

- **The fix works and the artifact runs.** The head-produced linux-arm64 binary has zero surviving
  stand-ins and prints `2.1.233 (Claude Code)` under `docker run --platform linux/arm64`. A full
  `clodex patch` inside an arm64 Linux container reproduces the `main` failure exactly
  (`Patch failed: the entry-module stand-in /clodex--/claude survived restoration`) and leaves the
  install untouched, confirming the rollback claim.
- **Your "not verified — whether the survivor is genuinely dead space" is answered, in your favour.**
  I ran the *unfixed* candidate that still carries the stand-in at its original offset: it starts and
  reports its version normally. The orphan is inert and the live table was correct, so `main`'s
  refusal really was a false alarm.
- **No macOS regression.** End-to-end `clodex patch` on darwin-arm64 2.1.233 with this PR:
  `11 applied, 0 skipped, 0 failed`, binary runs, `codesign --verify` clean, zero markers. Output is
  byte-identical to `main`'s. Re-checked with content deltas of +5 MB, +50 MB and half-size — one
  copy every time, restore OK.
- **Your five wrong implementations all turn the tests red**, as claimed; widening and removing the
  new guard are both caught too.
- **Your tweakcc 4.3.3 claims are true**, checked against the published tarballs (4.3.0 and 4.3.2 have
  no `/cli`; 4.3.3 does; published 2026-08-13, so `minimumReleaseAge` does block it until 2026-08-23).
  The "Left out" reasoning holds for a reason worth writing down — see M5.
- Sweep arithmetic is correct: exact offsets at position 0, EOF−16 and every alignment around the
  8 MiB boundary; adjacent and self-overlapping needles found exactly once; truncation and short
  writes fail closed. Gates green: typecheck, build, and 1800/1800 tests (base 1797/1797).

## M1 — the scope is wrong on both axes, and it is wrong in the docs (I am fixing this for you)

The failure is **not** macOS and **not** 2.1.233-specific. It is ELF-only, and it starts at **2.1.229**:
`inspectEntryModule` reports 2.1.228 `discoverable` and 2.1.229 `needs-shim` on both platforms;
`main` throws on linux-x64 2.1.229, 2.1.231, 2.1.232, 2.1.233 and 2.1.234, and on linux-arm64 and
both musl builds. 2.1.230 was never published. Mach-O and PE are unaffected — a maintainer's hourly
canary has been patching darwin-arm64 2.1.231–2.1.234 cleanly on `main` this whole time.

Rather than send you round the loop for prose while four platforms stay broken, I am pushing the
corrections to your branch myself before merging — flagging them here so the record is yours and you
can object. What changes:

- **Reachability:** "every patch attempt on a linux-x64/arm64 (glibc or musl) native install of
  2.1.229 or later; macOS and Windows are unaffected."
- **Evidence:** the macOS run was a no-regression check, not a reproduction.
- **"Not verified — other Claude Code versions":** replace with the measured range above.
- **`.claude/docs/patcher.md`** — "Claude Code 2.1.233 produces the benign one on every patch" and
  "That refusal made 2.1.233 unpatchable" are the ones that matter most, because that file is what
  the next maintainer reads before deciding whether the sweep can be deleted. As written it invites
  "2.1.233 is old, this can go."

A more precise mechanism for the doc, if useful: tweakcc's `repackNativeInstallation` branches on
container format, and only the ELF-with-a-`.bun`-section path **relocates** the section to
`align(nextVirtualAddress())` and repoints `BUN_COMPILED`, stranding the original bytes — PE and the
legacy overlay path assign in place. It is relocation, not appending.

## M2 — the sweep can silently rewrite content that is not a module name (MINOR, we will follow up)

Rewriting *every* byte range that matches the marker has no type invariant. With local patches
enabled (`--enable-local-patches`), a patch whose emitted source contains the literal
`/clodex--/claude` has that string rewritten to `/$bunfs/root/cli` in the published binary while
`clodex patch` reports success — reproduced independently twice, on Linux and macOS. `main` fails
closed here, so this is a real fail-closed → silent-corruption regression.

Reachability is adversarial-only: double-gated opt-in, and the trigger is clodex's own private
sentinel, which appears nowhere in `src/`. That is calibration, not a hole, and it must **not** block
this merge — refusing would leave four platforms unpatchable, a certain harm against a self-inflicted
one.

Nothing for you to do here — we will land the narrowing ourselves. Recording the method because it
is cheap: Bun NUL-terminates every string in the blob, and `readBunModuleNames` already relies on it. Measured on a real repacked candidate, the live copy and the orphan are both
followed by `0x00`, while JS string literals are followed by `0x22`/`0x27` — so requiring
`byte[offset + 16] === 0` discriminates perfectly. One caveat found the hard way: narrowing the sweep
alone leaves the literal in the file, which then trips the new `shimEntryModuleName` guard and makes
the binary unreadable on the next patch — the sweep **and** the guard need the same predicate.

## M3 — the summary line (maintainer will set the squash subject, so this is on us)

`restore patching on Claude Code 2.1.233 …` would be the third near-identical changelog bullet in a
row (2.5.1: "…on Claude Code 2.1.228…", 2.5.2: "…on Claude Code 2.1.231…"), and it misleads both
readers it reaches: a macOS user thinks they were broken, a Linux user on 2.1.231 thinks they were
not. Naming the platform is what makes it decidable, e.g.
`fix(patch): make clodex patch work again on Linux so model aliases and effort settings apply` (92).

## M4 — disclose the size result (MINOR, and explicitly *not* attributed to you)

A patched Linux binary is ~2.4x the pristine one (324 MB → 770 MB), with a measured peak around
2 GB during the patch as candidate, backup and tweakcc's temp coexist. This is pre-existing
tweakcc-on-ELF behaviour — 2.1.228, which needs no shim at all, balloons 2.45x identically — and it
does not compound, because the candidate is reseeded from pristine bytes each run. But this PR is
what makes a Linux user actually receive it, so one line in the body and one in `patcher.md` would
save someone a confused bug report. Please do **not** add a size sanity bound; at 2.37x today any
plausible cap would just recreate the failure this PR fixes.

## M5 — two small doc precisions

- "~37% into the file" is denominator-dependent: the orphan sits at a fixed absolute offset, which
  reads 37% under tweakcc 4.3.0 and 51% under 4.3.3 purely because the output size differs. Better to
  say *where* it is — the previous module table, left at its original offset.
- Worth stating why bumping the pin is not self-executing: clodex mirrors tweakcc's accepted-name
  list in `tweakccRecognizesModuleName` (`src/bun-entry-module.ts:72`) and it has no `/cli`, so the
  shim keeps firing until it is deleted. That is exactly why your 4.3.3 experiment reproduced, and
  it is the missing half of the "Left out" argument rather than a contradiction in it.

## Ours, not yours — landing separately

The rename landed in **2.1.229**, but our code comment, `patcher.md`, the `shimEntryModuleName`
docstring and the `claude-code-internals.md` heading all say 2.1.231 — our internals doc did flag
that 2.1.229/2.1.230 "were never available here", so it was an under-informed claim rather than a
wrong one, and it is now resolvable. We will fix ours in the same follow-up as M2; you inherited the habit
from us and should not have to carry it.

## Not verified

Whether a *third* stand-in copy can ever appear on some future repack, and behaviour on any Claude
Code release after 2.1.234.

bman654:
I pushed one docs-only commit to your branch (e143024) rather than send you a round-trip for prose —
shout if you'd rather revert it and do it yourself.

It rewrites the `patcher.md` scope claims to match what I measured across all eight published
platform packages, and it corrects a version that was **ours, not yours**: the entry-module rename
landed in **2.1.229**, not 2.1.231. 2.1.228 is the last release tweakcc can discover, 2.1.229 already
carries `/$bunfs/root/cli` on both platforms, and 2.1.230 was never published. You inherited 2.1.231
from our own docs and changelog; the remaining instances in `src/` comments and
`claude-code-internals.md` are on us and follow separately.

Merging with the summary line `fix(patch): make clodex patch work again on Linux so model aliases and
effort settings apply` — the changelog already carries two near-identical "restore patching on Claude
Code 2.1.2xx" bullets, and naming the platform is what lets a reader decide whether they care.

Thanks for this — it had been broken on every Linux install since 2.1.229 and nobody had noticed,
including our own release canary, which only ever tested the one platform that was immune. Fixing
that gap is the other thing your PR bought us.

PR #131, opened 2026-08-20
feat(models): choose a context stop against a provider pricing boundary

camjac251:
Clodex reported GPT-5.6's context window as a plain capacity number, so there was no way to ask for a bigger one and no sign that doing so costs more. It is really a cost dial: OpenAI bills prompts above 272,000 input tokens at a higher rate for the *whole* request, which is why Codex reports 272,000 rather than the model's ceiling. You can now choose how much of a model's window to use, with `clodex models --context sol=max --save` for a default or `clodex claude --context sol=max` for one launch, and clodex tells you when a stop crosses that line and warns once if a request actually does. `clodex models --json` prints the resolved numbers for scripting.

## The gap

`gpt-5.6-*` was pinned at a flat `contextWindow: 272_000`. Three consequences:

- No way to use more, even where the account allows it.
- No sign that a larger window changes billing, so nothing distinguished "more room" from "more expensive".
- The Codex catalog's own `max_context_window` and 95% effective-window convention were discarded, so the reported window was the raw number, not the one a client should fill.

## What changed

**Context stops.** A model carries a raw window, a ceiling, the share of the window to fill, and the boundary above which billing changes. `standard` is the provider's tuned window; `max` is the ceiling; a token count is clamped to it. `models --context` selects, `--save` stores, `claude --context` applies to one launch and is never written.

**Discovery.** `/backend-api/codex/models` already supplied `context_window`; `max_context_window` is now read as well. A catalog cached before that field existed is backfilled from the seed, because reading it back unchanged reports a raw window with no reachable ceiling and silently collapses `max` onto `standard`.

**Boundary warning.** Clodex warns once per model per process when a request's *own reported* input-token count crosses the boundary. Window math cannot see this: the client counts tokens in Anthropic shape and the provider counts them after translation, so only the returned usage settles it. It reads the prompt total rather than the Anthropic-split `input_tokens`, which has the cached portion removed and would under-report a crossing. Both bridge modes and `clodex server --endpoint` raise it, through the parent notice channel so it survives the stdout/stderr mute `clodex claude` holds for the child.

**`models --json`.** Resolved metadata for scripting. It reports clodex's own state rather than any consumer's config schema, and reports the client-native effort ladder separately from the provider's, because providers advertise levels Claude Code rejects outright (`none`) and every consumer would otherwise have to know which list is safe.

## Behaviour change for existing OAuth users

Reported windows for ChatGPT/Codex OAuth models drop 5% below the raw catalog value: `gpt-5.6-sol` reports **258,400 rather than 272,000**. That is the Codex client's own headroom convention, and it keeps a session under the pricing boundary rather than sitting exactly on it. It is scoped to that provider; API-key and OpenCode Go models keep their full window, which `tests/oauth-seed-overlay.test.ts` pins.

The practical consequence: patch output for the default configuration is **not** byte-identical to before, because the baked context window moves with it. Existing installs re-patch and take the new number.

## Deliberately left out

- **A per-model auto-compact patch.** An earlier revision added one. Executing the real 2.1.238 resolver shows a clodex model already reaches a rung returning the patched per-model window, so a second baked table could only lower it: at the `max` stop it withheld 82,840 tokens at exactly the window a user opted into and pays a higher input rate to reach. The patched context window delivers the larger window on its own.
- The boundary is claimed only for model families whose pricing is documented. `codex-auto-review` reports the same windows as GPT-5.6 in the catalog, but the rate card puts code review on a different model, so it gets no boundary rather than an assumed one.
- No ceiling is invented for a model the seed does not know; `terra`/`luna` carry the boundary but take their ceiling from discovery.
- No change to `[1m]` handling. The `max` stop resolves below 1,000,000 on this account, so the suffix path is untouched; a test pins that.

## Discriminating tests

Feature-deletion mutations, re-run against the current head, full-file runs. Every one turns its file red. The stale-cache backfill originally left the suite **green**, which is how I found it was unpinned; `tests/oauth-seed-overlay.test.ts` now covers it.

| Mutation | Result |
| --- | --- |
| boundary guard never reports | 5 red |
| boundary warning back on `console.error` | 1 red |
| latch consumed before the write | 1 red |
| latch keyed on display name | 1 red |
| effective-window headroom removed | 5 red |
| ceiling clamp removed | 1 red |
| discovery ceiling capture removed | 1 red |
| prompt-total reporting removed | 2 red |
| stale-cache backfill removed | 2 red (was green) |

Failure paths triggered rather than reasoned about: malformed saved stop, wrong-typed preference block, and control characters in a stop value all degrade to the standard window or a clean error, none throw.

## Runtime evidence

Environment: Linux, node v24.19.0, shell not bridged (no proxy vars set), isolated `CLODEX_HOME` where noted.

- **Real-resolver execution against a pristine 2.1.238 linux-x64 bundle.** Patches applied, then the auto-compact resolver extracted from the patched output and executed with every free identifier intercepted, so an unanticipated name throws rather than passing vacuously. With the context window baked at 828,400 and no auto-compact table present, a clodex model resolves to `{"window":828400,"configured":828400,"source":"unknown-model"}`; changing the baked value to 258,400 moves the result to 258,400. An explicit setting returns `source:"settings"` and the environment override returns `source:"env"`. This is what retired the auto-compact patch.

  The rung is `unknown-model`, not the `auto` fallback below it. That distinction matters rather than being cosmetic: enforcement is gated on the source *not* being `auto`, so landing on `auto` would switch auto-compaction enforcement off. Both rungs return identical `{window, configured}`, so the numbers above are unaffected. An earlier revision of this section said `auto`, which came from stubbing the zero-argument enforcement gate false and falling through a rung.
- `clodex patch` against a pristine ELF binary: all sites applied, 0 failed, baked table verified in the bytes, and the patched binary still runs.
- `clodex patch --restore` returns the byte-exact release and leaves no marker.
- Smoke, proxy mode: Anthropic passthrough `--model haiku` returns `HAIKU-OK`; translated `--model sol` returns `MODEL-OK` and reports `258.4K context`.
- Suite: 1870/1873 locally, with the three failures described below. CI on this branch is green.

## What I could not verify

- **Three `http-proxy-server.test.ts` service-tier tests time out on my machine.** They fail identically on unmodified `origin/main` at `d904051` in a separate clean worktree, so they are not caused by this change. They are not reproducible on the maintainer's machine or in CI. Not environment variables (a scrubbed `env -i` run fails the same way) and not blocked egress (`api.openai.com` answers). Cause still unknown; happy to chase it as its own issue.
- **Endpoint mode in `--print` never completes** on my machine. It starts the SDK adapter proxy, logs the model, and hangs until timeout. It reproduces on published 2.6.1 with none of this change applied. The maintainer reports it completing in about 7s for them, so this is likely the same local fault as above.
- **Whether the higher rate reaches ChatGPT-plan credits.** The multiplier is documented for API pricing; the Codex credit rate card lists flat rates and does not mention a boundary. Credits map 1:1 to API dollars, so it plausibly applies, but that is inference and I did not measure it. The larger stop carries roughly 3x the input tokens per turn regardless.
- **Ceilings are account-scoped and they move.** The 872,000 in the seed is what the catalog returned for a Pro account. The maintainer measured the real limit above 899,000, and their own transcripts show it as low as 398,000 in mid-July, drifting within a single day. Discovery overrides the seed and `max` clamps to whatever is reported, so a stale ceiling shrinks the stop rather than overshooting it.

## Failure and rollback

An unreadable or empty discovery response keeps the seed rather than falling back to a bare default, so "unavailable" and "272,000" stay distinct. A malformed saved stop resolves to `standard`. `models --json` writes diagnostics to stderr so stdout stays parseable. A launch-scoped `--context` is excluded from the patch config, which describes a binary outliving the process, so a one-off choice cannot bake itself in or mark a healthy install stale.

## Concurrency

`src/patch-transforms.ts` is byte-identical to the merge base, so the 2.1.238 picker fix in #133 and its `PATCH_TRANSFORMS_VERSION = 7` merge in untouched and there is no version collision to resolve. PRs #101-#104 and #110 touch some of the same files; #103 in particular introduces an effort policy with a canonical ladder that includes `none` and an explicit unsupported-level mapping. That subsumes the ad-hoc projection used here, and if #103 lands first this should be rebased onto it and the effort projection moved onto `nativeEffortLevel`. Those PRs do not touch `effectiveContextPercent` or `maxContextWindow`.


bman654 (changes requested):
Reviewed at `ba6cc99`. Four lenses plus refuters, in isolated worktrees; every claim below was
executed, not reasoned about. This is strong work — the context-stop model is coherent, the failure
paths really are hardened, and your central technical argument is correct (see *Confirmed* below).
Two things need to change before it lands, and one claim in the body needs retracting.

## Verdict: request changes — 1 blocker, 3 major

---

### BLOCKER — the pricing-boundary warning never reaches the user

`reportPricingBoundaryCrossing` (`src/pricing-boundary.ts:36-38`) defaults `emit` to
`console.error`, and neither call site (`src/proxy.ts:709`, `:739`) passes one. But `launchClaude`
replaces `process.stdout.write`/`process.stderr.write` with `muteWrite` for the entire lifetime of
the child (`src/launch.ts:63-64`, unconditional, all three call sites), and the gateway runs in that
same process. `console.error` routes through the replaced write and is swallowed.

`src/parent-notice.ts` exists *specifically* because of this. Its docstring names the exact previous
victims — "the reasoning- and tool-argument normalization canaries, the retry clamp notice … were
written straight into the mute and were never seen." `src/upstream-retry.ts:27` already uses
`emitParentNotice` **as a default parameter value**, which is precisely the shape this needs.

Worse, `warnedModels.add(modelLabel)` runs *before* `emit`, so the once-per-model latch is consumed
on the swallowed write — a later crossing where stderr is live is suppressed too.

Proved outside vitest (vitest intercepts `console`, which would mask it): an esbuild bundle run
under plain node, installing the real mute and the real notice sink, gave
`wentThroughMutedStderrWrite: 1, reachedParentNoticeQueue: 0` with the control reaching the queue.
An independent refuter attacked this and could not break it, confirming both bridge modes share the
muted process.

This is the PR's only runtime safety net for a documented 2x-input billing event, README documents
it as working, and it is reachable in 100% of `clodex claude` sessions — the boundary is attached at
any stop, not just `max`. **Fix: default `emit` to `emitParentNotice`.** One line.

---

### MAJOR 1 — every ChatGPT-OAuth model silently loses 5% of its window, and it gets baked in

`applyOAuthSeedContextMetadata` (`src/data/openai-oauth-models.ts:110-112`) sets
`effectiveContextPercent ?? DEFAULT_EFFECTIVE_CONTEXT_PERCENT` (95) for **every** model at read
time, via `projectProviderCachedModels` (`src/registry/materialize.ts:67`) — no refresh, no user
action, no opt-in:

```
BASE  sol/luna/terra 272K     gpt-5.3-codex-spark 128K     codex-auto-review 272K
HEAD  sol/luna/terra 258.4K   gpt-5.3-codex-spark 121.6K   codex-auto-review 258.4K
```

It applies where your own rationale says it should not: `gpt-5.3-codex-spark` has no pricing
boundary, and you explicitly exclude `codex-auto-review` because its rate card differs — both shrink
anyway.

**And the 95 is not the catalog's convention.** A live authenticated GET to
`/backend-api/codex/models` on a Pro account returns `max_context_window` for every model — matching
your seed *exactly*, including the absences, which is genuinely impressive — but returns
**no `effective_context_window_percent` at all**, and `auto_compact_token_limit` is **null for every
model**. So the body's "it also supplies `max_context_window` and `auto_compact_token_limit`, which
are now read" is half right; in practice the 95% is a clodex default that always wins.

Direction is safe (earlier compaction, never later), so this is not a correctness bug. It is an
undisclosed capability reduction for every existing OAuth user. Either gate the default on the field
actually being present, or keep it and say so in the summary line — per `CLAUDE.md` that line is the
changelog entry, and `feat(models): choose a context stop…` reads as purely additive.

### MAJOR 2 — "byte-identical patch output on default config" is false

From *Failure and rollback*: "PATCH 11 is skipped entirely when no model has a compaction target, so
the default configuration produces byte-identical patch output to before."

First half verified true. Second half is not. Running the real `buildDesiredPatchConfig()` on base
and head against one identical realistic `CLODEX_HOME` (three OAuth models, no `modelContextModes`,
no `--context`):

```
base  "context":272000 (all three)   digest 0e0d7870a4e675df…
head  "context":258400 (all three)   digest 034702041682f3e9…
PATCH 11 absent from results on BOTH sides — 11 sites, identical statuses.
```

PATCH 11 is inert exactly as you say; it is **PATCH 7** that changes, because MAJOR 1's 95% flows
into the baked window. And since `PATCH_TRANSFORMS_VERSION` goes 6→7, every existing install is
marked stale, so every patched user re-patches and silently takes the 5% cut. Rollback risk is the
one thing that section is read for, so this claim should not survive to merge.

### MAJOR 3 — a new test is red whenever `CLAUDE_CODE_AUTO_COMPACT_WINDOW` is set

`tests/patcher.test.ts`, "yields to an explicit setting and to the environment override". The
patched resolver checks `!process.env.CLAUDE_CODE_AUTO_COMPACT_WINDOW` (`src/patch-transforms.ts:592`),
but the test only saves/restores that variable around its **last** leg; the first three read ambient.

```
$ CLAUDE_CODE_AUTO_COMPACT_WINDOW=250000 npx vitest run tests/patcher.test.ts -t "yields to an explicit setting"
AssertionError: expected { window: 150000, …} to deeply equal { window: 900000, …}
```

`main` is green in the same environment. Anyone who has set that documented override — and every
contributor running the suite from inside a Claude Code session — gets a red suite. Four independent
lenses hit this. Fix: clear it in `beforeEach`, restore in `finally`, covering all four legs.

---

### MINOR

- **The harness passes vacuously.** `patch11-autocompact-execute-real-resolver.harness.ts` exits 0
  with 12 tests **skipped** when `REVIEW_BUNDLE_DIR` is unset. It is a good harness when pointed at
  bundles; as written, a green run proves nothing. Make the absence fail loudly.
- **`clodex server --endpoint` never warns at all** — `src/server/router.ts:462`/`:474` call the
  generators without `onPromptTokens`. Only `src/proxy.ts` was wired.
- **"Once per model per session" is wrong for a long-lived server.** `warnedModels` is a module-level
  `Set`, never cleared. One tenant crossing silences the warning for every other tenant for the life
  of the daemon; two providers sharing a display name share one slot. README says "session".
- **`--context` on `clodex claude` is inert in proxy mode** (the default) and prints no cost warning,
  while the `models` form does. It also invalidates the patch-config hash, so it nags "not patched"
  on every launch. Either make it behave, or scope the README/help to endpoint mode.
- **No CLI-level coverage for the new flags.** `--context`, `--json`, `--save` have no test hits;
  `tests/cli.test.ts:290-306` has an established block pinning every other flag into the help text.
- **Orphaned JSDoc** — `reportPromptTokens` was inserted between the `toAnthropicUsage` doc comment
  and `toAnthropicUsage`, so the comment explaining the cache-subset arithmetic now binds to the
  wrong function. That is the very doc a reviewer needs for your central claim.

---

### Confirmed — checked and cleared, so you do not re-derive them

- **Your central reasoning is right.** Reading the pre-split prompt total rather than Anthropic
  `input_tokens` is correct: `toAnthropicUsage` computes `input_tokens = max(0, total - cacheRead -
  cacheWrite)`, so the split really would under-report a crossing. Verified against the code and
  your own test's 284,102 → 3,102 split. Only the delivery channel is broken.
- **PATCH 11's anchor is sound** — unique, self-identifying, binds on all 8 platform builds of
  2.1.238 *including the three where PATCH 5 currently breaks*, fails safe on ambiguity, 305-byte
  surgical insertion, prototype-safe.
- **The larger stop genuinely works on `openai-oauth`.** Live probes: 399,887 and 899,763 input
  tokens both `completed` on `gpt-5.6-sol`; 1.69M returned `context_length_exceeded`. The backend
  hard-errors rather than truncating, and the real limit is above the advertised 872,000 — so using
  `max_context_window` as the ceiling is conservative. Your seed matches live discovery exactly.
- **`models --json` does not leak credentials** (canary-injected scan across `apiKey`, `authRef`,
  `oauthAccounts.refreshToken`, every `apiUrl` — zero hits) and stdout stays pure JSON with all
  diagnostics on stderr, as you claim.
- **Endpoint `/v1/models` updates automatically**; **malformed input degrades cleanly** (`sol=banana`,
  `nope=max`, `sol=2000k`, `--save` with no `--context` — none throw), matching your description.

### On the reported baseline — not blocking, but worth correcting

We could not reproduce either "pre-existing failure". `tests/http-proxy-server.test.ts` passed 25/25
in 10 runs on the merge-base across ambient, unreachable-proxy and cleared environments, and **CI on
this exact head is green at 1880/1880** (not 1877/1880) on Ubuntu/Node 24.19.0. Endpoint `--print`
also completed for us in ~7s with a full trace through response and shutdown.

One likely explanation, offered rather than asserted: an agent session exports
`CLAUDE_CODE_ENTRYPOINT=sdk-cli`, which makes `tests/proxy.test.ts` fail spuriously — a different
file from the three you named, but the same class of thing. If you can still reproduce the hang,
send the exact command, model, Claude Code version and trace log; we will chase it as its own issue.

### Merge order

An in-flight fix for the Claude Code 2.1.238 PATCH 5 break also sets `PATCH_TRANSFORMS_VERSION = 7`
and touches `src/patch-transforms.ts`, `tests/patcher.test.ts` and `.claude/harnesses/README.md`.
The code regions are disjoint and there is no semantic interaction with PATCH 11, but whoever lands
second must bump to **8** and re-pin the digest. If yours lands second we will handle the rebase.

bman654:
Follow-up after tracing this against the real Claude Code bundle. **One correction I owe you, one
finding that removes work rather than adding it, and the blocker still stands.**

## Correction — withdraw my framing of the 95%

I said the 95% effective-context default "is not the catalog's convention" and implied it was
ungated. **That was unfair and I withdraw it.** The scoping is real, it is just *provider*-level
rather than model-level, and it is deliberate and pinned by `tests/oauth-seed-overlay.test.ts:100-110`
with the rationale in as many words: "Applying it to an API-key provider would shrink every other
model by 5%." Non-OAuth providers resolve to `undefined` and keep their full window.

So the 5% reduction is an intentional, tested, provider-scoped decision, not an accident. The most
that survives is a wording nit: the constant's comment says "applied only to models that declare it",
which reads as model-level gating when the gating is provider-level.

The underlying observation still stands and is worth a line in the PR body or release note — every
existing ChatGPT-OAuth user's reported window moves 272,000 → 258,400 on upgrade — but as a
disclosed consequence of a defensible default, which is a much smaller ask than what I posted.

## Your per-process premise is TRUE — I verified it

"The client resolves one auto-compact window per process, so raising one model's window drags every
other model off its tuned default." Confirmed by execution against the real resolver. Both channels
that can set it — the `autoCompactWindow` setting and `CLAUDE_CODE_AUTO_COMPACT_WINDOW` — take a
single scalar for the whole process, and neither accepts a per-model map:

```
claude-sonnet-5, nothing set   → {"window":967000, "source":"model-default"}
claude-sonnet-5, set to 745560 → {"window":745560, "source":"settings"}   ← leaked
```

And the direction you needed made it worse, not better. You were trying to **raise** `sol`, and the
only other channel — `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` — is a fraction, so raising one model through
it means raising the fraction for *every* model. That is why reaching for the absolute setting was
the reasonable move: it is the only channel that can express a bigger number. It just happens to be
process-global too, so it flattens every other model onto `sol`'s value. **Neither channel can target
one model, which is exactly why you built a patch.** The reasoning is sound; it is only the
conclusion that a *new* patch was needed that does not hold.

Claude Code carries its own per-model table (`claude-sonnet-5` → `default: 967000`, with
`remote_cowork`/`local-agent` surfaces at 500,000), and setting either channel overrides it for every
model. You identified a real problem, and your `MIN`/`MAX_AUTO_COMPACT_WINDOW` bounds of 100,000 and
1,000,000 correctly mirror the client's own schema (`int().min(1e5).max(1e6)`).

## But PATCH 11 is not needed — please drop it

**PATCH 7 already solves this.** Executing the real resolver shows a clodex model lands on the
`unknown-model` rung, which returns `{window: <the PATCH 7 value>}` — already per-model. The
per-model number is being delivered today, without touching the process-global knob at all.

So PATCH 11 never raises anything. It only lowers:

| | Compaction fires at |
| --- | --- |
| without PATCH 11 | **91.2%** of the raw window |
| with PATCH 11 | **81.7%** of the raw window |

For `sol=max` that is 828,400 → 745,560 — withholding **82,840 tokens at precisely the stop a user
opted into and is paying a 2x input premium to reach.** Dropping it is not a regression: compaction
still fires with 76,600 raw tokens free, and the compaction call itself fits with 44,600 to spare.

**What the primary goal actually needs:** bake the higher number into the metadata `clodex patch`
reads, so that when **PATCH 7** is applied the higher number gets baked into Claude Code's existing
per-model table. No second table, and no new patch site.

Dropping PATCH 11 also dissolves most of my earlier findings rather than requiring fixes:

- the "byte-identical patch output on default config" claim (it was PATCH 7's baked window changing);
- the new `tests/patcher.test.ts` case that goes red whenever `CLAUDE_CODE_AUTO_COMPACT_WINDOW` is set;
- an arithmetic defect I had not reported yet: `MAX_STOP_AUTO_COMPACT_PERCENT` is documented as
  "900,000 against a 1,000,000 window" — a fraction of **raw** — but applied to **effective**,
  shipping 0.855 of raw, 39,240 tokens off its own stated rule;
- the PATCH 11 anchor matching **0 times on 2.1.208-2.1.221** while the CLI still prints
  `auto-compact at 745,560`;
- and a `PATCH_TRANSFORMS_VERSION` collision with an in-flight fix for the Claude Code 2.1.238
  PATCH 5 break, which also sets it to 7.

## Still blocking: the boundary warning never reaches the user

Unchanged from my review, and independently confirmed by a refuter that tried to break it: the
warning goes to `console.error`, `launchClaude` mutes stderr for the child's lifetime, and the
gateway runs in that process. `src/parent-notice.ts` exists for exactly this, and
`src/upstream-retry.ts:27` already uses `emitParentNotice` **as a default parameter value** — the
shape to copy. The latch is also consumed before the emit, so a later crossing where stderr is live
is suppressed too.

That one is a one-line fix and it is the difference between the cost feature having a safety net and
not having one.

## One vendor fact, offered rather than asked

We measured the OAuth path directly. Requests of 399,887 / 873,776 / 899,763 input tokens all
completed on `gpt-5.6-sol`; 1.69M was rejected with `context_length_exceeded`. So the backend
hard-errors rather than truncating, and the real ceiling is **above** the advertised 872,000 — your
use of `max_context_window` is conservative, and your seed matches live discovery exactly, including
the absences.

Worth knowing though: that ceiling has moved. Our own transcripts show a hard cutoff near 398,000 in
mid-July (397,468 succeeded; 398,284 rejected 17 seconds later on the same route), drifting between
381,000 and 398,000 within a single day. It is now past 899,000. Nothing to change in the code — the
clamp handles it — but "account-scoped" is worth extending to "and it moves".

camjac251:
Pushed as `2f84928`. I took the follow-up over the original review where they disagree, and I reproduced your central claim before acting on it rather than deleting on assertion.

## PATCH 11 is gone

I extracted the pristine 2.1.238 linux-x64 bundle, applied the patches with **no** auto-compact table, pulled the auto-compact resolver out of the patched output, and executed it with every free identifier intercepted through a `with` + Proxy scope, so a name I failed to anticipate throws instead of passing vacuously:

```
PATCH 11 marker present: false

sol, nothing set   : {"window":828400,"configured":828400,"source":"auto"}
sol, baked 258400  : {"window":258400,"configured":258400,"source":"auto"}
sol, setting 400k  : {"window":400000,"configured":400000,"source":"settings"}
sol, env 150000    : {"window":150000,"configured":150000,"source":"env"}
```

Changing only the baked context window moves the resolved auto-compact window with it. The per-model number is already being delivered, the terminal rung returns it unclamped, and a second table could only lower it. The patch, its config plumbing, its tests, and the harness are all removed, and `PATCH_TRANSFORMS_VERSION` is back to 6.

That also removes the arithmetic defect you had not filed yet, which I confirmed by inspection: `MAX_STOP_AUTO_COMPACT_PERCENT` documented itself as a fraction of raw and was applied to effective, shipping `raw * 0.855`. It went with the derivation.

## Blocker fixed

`emit` now defaults to `emitParentNotice`, copying the shape in `src/upstream-retry.ts:27`, and the latch is set **after** the write returns rather than before it. Three mutations cover it, each red on its own: reverting to `console.error`, moving the `add` back above the `emit`, and keying the latch on the display name.

## Version collision is resolved rather than deferred

`src/patch-transforms.ts` is now **byte-identical to the merge base**, so #133's picker fix and its `PATCH_TRANSFORMS_VERSION = 7` merge in untouched. No bump to 8, no digest re-pin, no rebase needed on your side. `tests/patcher.test.ts`'s pin is back to the base values for the same reason.

## MAJOR 1: taking the second option, with the disclosure

I had a gating change drafted before your follow-up landed. I dropped it: the provider-scoped 95% stays as it is, and the consequence is disclosed instead, which keeps the headroom that holds a session under the boundary. The PR body now carries a **Behaviour change for existing OAuth users** section stating the 272,000 to 258,400 move plainly, and the commit body carries it too so it reaches the changelog. I also fixed the constant's docstring, which said "applied only to models that declare it" when the gating is provider-level.

The "byte-identical patch output" claim is **retracted**, not engineered around. It was PATCH 7's baked window changing, that stays true, so the body now says patch output is not byte-identical and existing installs take the new number.

## Minors

- **Endpoint mode warns now.** `ServerModelInfo` carries `pricingBoundary` and both generator calls in `src/server/router.ts` pass `onPromptTokens`.
- **Cross-tenant latch fixed**, and the README now says "once per model for the life of the process" rather than "session".
- **`--context` no longer nags.** `buildDesiredPatchConfig` reads saved stops only. A launch-scoped stop was changing the config hash for a binary that outlives the process. The launch path also prints the same cost accounting the `models` form does, and the help and README now say plainly that a patched binary keeps its baked window until the stop is saved and the patch re-run.
- **CLI coverage added** for `--context`, `--json`, and `--save` in the help-text pin block.
- **Orphaned JSDoc fixed.** `reportPromptTokens` moved above the `toAnthropicUsage` doc comment.
- **Harness** is deleted along with the patch, so the vacuous-pass point is moot. Had it survived I would have made the absence fail loudly, as `pr78-patch10-anchor-over-real-bundles` does.

## On the baseline, where I have to push back slightly

The three `http-proxy-server.test.ts` service-tier timeouts still reproduce here, and they reproduce on **unmodified `origin/main` at `d904051`** in a separate clean worktree, so they are not caused by this branch. I checked your hypothesis and it does not hold on my side: `CLAUDE_CODE_ENTRYPOINT` is `cli`, not `sdk-cli`, and a fully scrubbed `env -i HOME=... PATH=... TERM=dumb` run fails identically. Blocked egress is also out, `api.openai.com` answers with a 401 in about a second.

So it is real here and absent for you and in CI, which points at something about this host rather than the code. My **1877/1880** figure was wrong regardless and I withdraw it. CI on this branch is green, and locally I now see 1870/1873 with exactly those three. Happy to open a separate issue with a trace log if you want to chase it, but nothing here is blocked on it.

Full mutation table, the resolver transcript, and the corrected claims are in the updated PR body.

bman654 (changes requested):
Re-reviewed at `2f84928` as a fresh panel, not a re-read — dropping PATCH 11 made this a redesign.
Four lenses plus refuters in isolated worktrees, then a cross-family completeness pass. **The
blocker is fixed, our own advice to drop PATCH 11 is vindicated by execution, and nothing blocking
survived.** What is left is five small things, two of which are ours rather than yours.

## Verdict: request changes — 0 blockers, 0 majors, 5 minors

Nothing here is structural. **Three must-fix items, each one to three lines:**

1. Carry `pricingBoundary` through `startProxy`, so the runtime warning can fire on
   `clodex claude --endpoint` (MINOR 1).
2. Give `models --json` a session-aware path, so it stops contradicting its own stderr (MINOR 2).
3. Drop the `Add --save` footer from the launch path and move the report after the `--help`
   short-circuit (MINOR 3).

The other two are a test-coverage ask where **our round-1 guidance was the insufficient part**
(MINOR 4) and a prose fix in the PR body (MINOR 5). I am asking rather than approving because two
of the three put wrong data or a missing safety net in front of a user, and one is a regression
that our own round-1 ask introduced — I would rather catch that before it ships than after. If you
would rather land now and follow up, say so and I will approve instead.

---

## First, what we got right — and what we got wrong

**Dropping PATCH 11 was correct, and it is now proven rather than argued.** A lens extracted all
16 supported bundles and executed the real auto-compact resolver out of the *patched* output, with
free identifiers intercepted so an unanticipated name throws instead of passing vacuously. PATCH 7's
baked window feeds the resolver on every release from 2.1.208 to 2.1.238, and on all five 2.1.238
platform builds. A real `clodex patch` run against a pristine binary baked `sol=828400`, and the
resolver read back 828,400 from the actual patched bytes.

The headroom numbers from my follow-up reproduce exactly: trigger at 795,400, 76,600 raw tokens
free, and 44,600 spare after Claude Code's own 32,000-token compaction request. So the case I was
most worried about — that we had talked you into re-opening "Prompt is too long" at the `max` stop —
did not happen.

**One retraction I owe you: your mutation table is right and I nearly filed against it.** A lens
reported that "effective-window headroom removed" yields 4 red, not the 5 you claim. A refuter ran
both spellings and found that removing the headroom *where the headroom is computed* —
`effectiveContextWindow` itself — gives exactly **5 red**, full-file and whole-suite agreeing. That
is the more canonical reading of your own row. **Nine of nine mutation-table counts reproduce.**
Nothing for you to change; I am recording it so the claim is not re-litigated.

---

## MINOR 1 — `clodex claude --endpoint` drops the pricing boundary, so the warning cannot fire there

`startProxy` (`src/proxy.ts:874-921`) builds its `ProxyRoute` as an object literal with no spread,
and its `sdk` parameter type has no `pricingBoundary` member — so `src/cli.ts:1574` could not pass
one even if it tried; `tsc` would reject it. Both consumers read `route.pricingBoundary`, which is
`undefined`, and `reportPricingBoundaryCrossing` returns `false` at its first guard.

Proven twice independently: an esbuild probe under plain node with the real mute installed, and a
vitest probe driving real HTTP through the real proxy. Control route with the field set → warning;
`startProxy` route → nothing. A two-line addition to `src/proxy.ts` plus one at the call site makes
the warning appear.

Reachability is wider than "explicit `--provider`/`--model`". `planLaunchWizard` returns
`skip: true` for three cases, including **print mode** — every `clodex claude --endpoint -p …`, even
with favorites saved — plus the non-TTY fallback and the no-favorites case. `emitParentNotice` goes
to stderr, so `-p` mode would still receive it.

**Why this is MINOR and not the old blocker.** The default bridge mode is proxy, which carries the
field; the endpoint switch-menu path and `clodex server --endpoint` both carry it. The same wrapper
already drops `refreshToken` on the merge base and on `main` — a strictly worse omission, costing an
expired-token 401 — so this is a pre-existing wrapper-fidelity pattern, not a departure. And nobody
reaches the higher-rate band unwarned: `applyContextStopAssignments` prints the selection-time
warning in `main()` before bridge dispatch, in every mode. The runtime line is the confirmation, not
the disclosure.

Two asks: carry the field through, and add the composition test — nothing in the suite observes it
today, in either direction (87/87 stayed green with the fix applied).

While you are there: `ProxyRoute.maxOutputTokens` (`src/proxy.ts:243`) is written by
`src/catalog.ts:36` and **read by nothing**. I traced every hit of the value; the only SDK read takes
it from `body.max_tokens`, never from the route.

## MINOR 2 — `models --context X --json` reports the saved stop while its own stderr reports the session stop

This one is ours. `buildDesiredPatchConfig` correctly reads saved stops only now — that is exactly
what round 1 asked for — but `buildModelMetadata` (`src/model-metadata.ts:93`) also calls it, so the
JSON surface went saved-only with it.

```
$ node dist/cli.js models --list --context sol=max
   sol  … (828.4K context)                         ← session-aware materializer

$ node dist/cli.js models --context sol=max --json
stderr: sol …: max: 828,400 tokens.
stdout: "context":{"stop":"standard","raw":272000,"effective":258400,…}
```

Two output surfaces of one command, in one process, disagreeing. At `ba6cc99` stdout said `max`;
at `2f84928` it says `standard`. Not present on the merge base or `main` (neither flag exists there).

It stays MINOR because the blast radius is exactly one undocumented flag combination — `--context`
with `--json` and *without* `--save`, which the help's Usage and Examples do not show. With `--save`
the two surfaces agree. There is one production caller, under `opts.json`; no launch, server, patch
or routing path reads it, so nothing misroutes and nothing bakes a wrong number. But it is wrong
data on a surface whose whole purpose is being machine-readable. Keep the saved-only rule on the
patch-config path and give the metadata path a session-aware one; pin it with a test that asserts
stderr and parsed stdout agree.

## MINOR 3 — `clodex claude --context` tells users to pass a flag that command does not accept

`applyContextStopAssignments`'s footer is reused on the launch path (`src/cli.ts:808-809`), so every
launch-scoped assignment prints `Add --save to make it the default`. But `--save` is consumed only in
the `models` branch; on `claude` it falls through to `claudeArgs` and is forwarded to the child, and
Claude Code silently ignores it (`claude --save --version` → exit 0, no such flag). A user who
follows our own instruction gets no error and no saved stop.

Same call also runs before the `--help` / `--version` short-circuits:

```
$ node dist/cli.js claude --context sol=max --help
sol …: max: 828,400 tokens.
… Applies to this run only. Add --save to make it the default.
clodex claude v2.6.1                                    ← help, three lines late
```

and with an unknown target it suppresses help entirely and exits 1. Both arrived with this commit;
at `ba6cc99` that command printed help with no preamble. Use a claude-specific footer naming
`clodex models --context … --save`, and move the call after the help/version short-circuits.

## MINOR 4 — the two new CLI surfaces are unpinned, and the pins we asked for cannot catch it

**Our earlier ask was the insufficient one.** Round 1 pointed you at the help-text pin block by name
as the model to follow, and you delivered exactly that. Those pins prove the flags appear in help;
they cannot catch the feature being gone. Two feature deletions clear the entire gate —
`typecheck`, 1873/1873, and `build`:

- `buildModelMetadata` → `return []`, and `clodex models --json` silently returns `[]` for every user.
- delete the `if (parsed.contextStops?.length)` block from `main()`, and every `--context`
  assignment stops reaching saved/session state and cost reporting. Verified dead against the real
  built CLI: `modelContextModes: null` and exit 1, versus the saved stop and the boundary warning at head.

I checked whether this is your problem or the repo's, by mutating pre-existing shipped surfaces on
the merge base: deleting `models --list` and `clodex providers` wholesale both leave 1802/1802
green. So two of three long-shipped CLI surfaces are equally unpinned and this does not depart from
posture — which is why it is MINOR and not the MAJOR it was filed as. What keeps it alive is the
third: `models --unalias` *is* pinned, via `runModelsCommand` driven against a real isolated
`CLODEX_HOME` in `tests/models-alias-command.test.ts`. Both of your new surfaces live in that same
function and `tests/cli.test.ts` already drives `main()`. That is the pattern we should have pointed
you at; extending it is cheap.

Related, and not something round 1 caught either: `tests/model-metadata.test.ts` never invokes
`buildModelMetadata`. Four of its seven tests assert properties of the hand-written `SOL` literal
declared 30 lines above them, so they cannot fail for any change to `src/` — and the literal is not
the shape production emits (real output has `"GPT-5.6 Sol (OpenAI (ChatGPT))"`, the fixture says
`"GPT-5.6 Sol"`).

## MINOR 5 — the resolver transcript in the body reports the wrong `source`, and the right one helps you

The body's runtime-evidence section says a clodex model resolves to `source:"auto"`. Enumerating
every branch of the terminal gate in the real 2.1.238 bundle, a clodex identity on a default install
lands on `unknown-model`; reaching `auto` needs auto-compact disabled, the enforcement env override,
a `[1m]` id, or an id Claude Code recognizes — none of which is this path.

This matters more than a label, and in your favour. `nGe()` gates enforcement on
`source !== "auto"`, so `"auto"` is the one value that turns auto-compaction enforcement **off**. If
the transcript were literally true it would undercut dropping PATCH 11; the real answer is the one
that makes your conclusion right. Both rungs return identical `{window, configured}`, so every
number you published is correct and PATCH 11 stays dead. Prose fix only, in the section that exists
to justify a removal. (Our own `.claude/docs/claude-code-internals.md:81` says the same thing and
predates the PR, so this did not come from us.)

---

## Confirmed — checked and cleared, so you do not re-derive them

- **The blocker is genuinely fixed.** Proven outside vitest, since vitest intercepts `console`: an
  esbuild bundle under plain node with the real `launchClaude` mute installed and the real notice
  sink. The warning appears after the child exits, i.e. it reached the parent-notice queue, and the
  muted control write had zero hits. Both bridge modes and `clodex server --endpoint`.
- **The latch fixes hold.** Keyed on identity, set after the write returns. The two mutations each
  turn exactly the right test red. Proxy and endpoint key spellings differ but are mutually
  exclusive branches in one process, so no double warning.
- **`clodex server --endpoint` wiring is real, not decorative.** Traced from the OAuth seed through
  `projectContextStop` → `LocalProviderModel` → `ServerModelInfo` → router: `272000` at every stage,
  and the request emitted the warning.
- **The non-streaming path really invokes `onPromptTokens`**, with the pre-split total: stub reported
  300,000 prompt / 281,000 cached, callback received `300000`, Anthropic `input_tokens` was 19,000.
- **A foreground server with no parent does not drop the notice** — it falls back to real stderr.
- **`--context` no longer nags, and does not write under `--dry-run`.** Session `max` reaches the
  runtime catalog at 828,400 while the patch hash is unchanged; a *saved* `max` bakes 828,400,
  changes the hash and marks an existing install stale. Zero files changed under `--dry-run`.
- **PATCH 11 removal is complete.** No residue for `autoCompactWindow`, the three constants,
  `ccpatch:autocompact`, `745,560`, `82,840` or `0.855`; the harness and its README row are gone.
- **Landing is clean.** I trial-merged `origin/main` (now `d904051`, 2.6.2, carrying #133's
  `PATCH_TRANSFORMS_VERSION = 7`) into your head on a verified-clean tree: no conflict, version 7
  inherited untouched, **typecheck pass, 1889/1889, build success**. Your byte-identity claim for
  `src/patch-transforms.ts` and `tests/patcher.test.ts` checks out (`cmp` exit 0 both). Nothing for
  you to rebase.
- **The three `http-proxy-server.test.ts` timeouts are not yours.** 12 runs across merge base and
  head, ambient and dead proxy: 300/300 passed. It does not reproduce here and the differential
  gives no support for a PR-caused regression, which matches your account that it is host-specific.
  Your withdrawal of the 1877/1880 figure is noted; no need to chase it for this PR.
- **The 5% OAuth disclosure and the byte-identical retraction are accurate.** Base/head production
  execution: 272,000 → 258,400, hashes differ. The provider-scoped default and the constant's
  corrected docstring are both right.
- **Commit hygiene passes.** Both summaries ≤ 73 chars, no body line over 100, commitlint green.

## What nobody verified, stated plainly

**No lens sent a real request to a real provider at this head.** Every warning proof used a stubbed
usage report of 300,000 tokens at the `onPromptTokens` seam. That seam is exercised correctly, but
"warns once if a request actually does cross" has not been observed against a real provider usage
payload by anyone but you. Likewise nobody drove a real patched binary at `sol=max` past 795,400
tokens to watch compaction actually fire; the evidence there is executed-resolver arithmetic plus
the prior round's live 899,763-token completion. I rate the residual risk low, but it is not zero,
and you should know which of our green results are stubs.

One thing worth flagging for later rather than now: the account ceiling has measurably drifted
(≈398,000 in July, past 899,000 now). A binary baked at 828,400 keeps that number until re-patch, so
a downward drift would turn the 795,400 trigger into a hard `context_length_exceeded` before
compaction can fire. Inherent to baking any window and disclosed by the opt-in, so not an ask
against this PR.

camjac251:
Pushed as `0dc363f`. All five addressed, including the two you flagged as yours. Suite is 1881/1884 locally, the three being the known host-specific timeouts.

## MINOR 1: the boundary now reaches every launch path

You were right that the type made it impossible to pass. Rather than add the field to a literal that could lose it again, I extracted the wrapper's route construction into `singleModelProxyRoute` and pinned it in `tests/proxy-route-composition.test.ts`. Deleting the `pricingBoundary` line turns it red; nothing observed it before.

Also removed `ProxyRoute.maxOutputTokens`. I traced it the same way you did: written by `src/catalog.ts`, read by nothing, and it came in with this branch, so it is mine to take back out.

## MINOR 2: fixed, and the diagnosis was exactly right

The saved-only rule was correct for the patch config and wrong to inherit on the metadata surface. `buildDesiredPatchConfig` now takes a `sessionStops` option, off by default. The patch path keeps saved-only; `buildModelMetadata` passes it on, because that surface reports what the current run resolves to rather than what is baked into a binary.

Pinned by a test asserting stderr and parsed stdout agree, and that nothing is written to preferences without `--save`.

## MINOR 3: fixed, plus the help suppression you noted

Launch-scoped assignments now name `clodex models --context <model=stop> --save`. The report moved after the help and version short-circuits, and I went one further than asked: the whole context block is now skipped when `--help` or `--version` is set. Your parenthetical about an unknown target suppressing help entirely was the case that forced it, since that one failed at parse time, upstream of where I had moved the report.

## MINOR 4: taking the pattern you actually meant

Fair, and the distinction is worth having stated plainly: the help pins prove a flag is documented, not that anything is behind it. `tests/models-context-command.test.ts` drives `main()` and `runModelsCommand` against an isolated `CLODEX_HOME` with a seeded registry. Six deletions, each red:

| Mutation | Result |
| --- | --- |
| `buildModelMetadata` returns `[]` | 3 red |
| context-stop block deleted from `main()` | 5 red |
| metadata path reverts to saved-only stops | 1 red |
| launch footer reverts to `Add --save` | 1 red |
| stop report moves back ahead of the help short-circuit | 1 red |
| wrapper drops the pricing boundary | 1 red |

Your note on `tests/model-metadata.test.ts` is correct and I left it alone deliberately: it is a formatter test whose fixture is a literal, and the new file covers `buildModelMetadata` against real state. Happy to fold the two together if you would rather not carry a fixture that does not match production shape.

## MINOR 5: corrected, and thank you for not letting it pass as a label

This one was worth catching. I re-ran the resolver with the enforcement gate modelled properly and get `unknown-model`:

```
sol, nothing set   : {"window":828400,"configured":828400,"source":"unknown-model"}
sol, baked 258400  : {"window":258400,"configured":258400,"source":"unknown-model"}
```

The cause was a stubbing error on my side, and it is a trap worth recording: that rung is guarded by a **zero-argument** gate alongside argument-taking predicates, so a blanket `() => false` catch-all switches the gate off and falls through to `auto`. Stubbing zero-arg calls true fixes it. Every number was unaffected, and as you say the correct rung is the one that makes the removal argument hold, since `auto` is the value that disables enforcement. The PR body now says `unknown-model` and explains why the distinction matters.

## On your two retractions

Noted and appreciated, particularly re-running the headroom mutation both ways rather than filing against my count. I have left that row at 5.

I would rather land this than iterate further, but I am not asking you to skip anything: everything in this round is done and verified, so approve or re-review as you prefer. `src/patch-transforms.ts` is still byte-identical to the merge base, so #133 and its version 7 remain untouched.

On the stubbed-usage caveat: agreed, and worth stating plainly for the record. I have not driven a real provider payload across the boundary at this head either. The `onPromptTokens` seam is exercised and the arithmetic checks out, but the end-to-end "a real request crossed and the user saw the line" remains unobserved by both of us.

bman654 (approved):
# APPROVE — all five round-2 minors fixed, each verified by execution

Round 3 is a clean close-out. I re-reviewed only `2f84928..0dc363f`, and every fix in it holds up
under execution rather than reading. Three lenses plus my own checks; nothing blocking survived.

**Landing check.** `main` has moved twice since my last review (now `8c7007c`, carrying #135's
2.1.239 patch fix). I trial-merged it into your head: **no conflict, typecheck clean, 1914/1914,
build success.** `src/patch-transforms.ts` is still untouched by this PR, so #133/#135's
`PATCH_TRANSFORMS_VERSION = 7` is unaffected. Nothing for you to rebase.

## The five, confirmed

- **MINOR 1.** The extraction is faithful, which is the part worth proving rather than assuming. I
  captured the old inline literal by rewriting only its `return startProxyCatalog([{` into a
  capturable form, then deep-compared it against `singleModelProxyRoute` over four argument shapes
  (fully-populated `sdk`, `sdk: undefined`, no `contextWindow`, `[1m]`-suffixed id):
  `{"onlyOld":[],"onlyNew":["pricingBoundary"],"valDiff":[]}` in all four. No key dropped, renamed,
  retyped or re-defaulted. The `aliasId` dual-use survives too — the new `startProxyCatalog([route],
  route.aliasId, debug)` second argument equals the old `clientModelId` local in every shape.
- **MINOR 2.** The invariant that mattered is intact. With a session stop set, the patch config hash
  is byte-identical (`cd105253…8b8301` before and after), the seeded manifest reports
  `stale_lines=0`, and preferences stay empty; with `--save` the hash moves to `6da00c2a…cd678` and
  the launch path reports the binary stale. stdout and stderr now agree at 828,400.
- **MINOR 3.** Confirmed by executing the real CLI, including the case that forced your extra step:
  `claude --context nope=max --help` prints help and exits 0. A control-flow matrix across root,
  `models`, `server`, `patch`, `providers` and `claude` — run against both the old and new ordering
  — shows **no path that newly reports, and none that lost a report on real work**. The only
  differences are the two intended ones.
- **MINOR 4.** Six mutations reproduce. The distinction you restated is the right one.
- **MINOR 5.** Corrected, and the note on why the rung matters is now in the body.

**`maxOutputTokens` removal is in scope and safe.** `git show bb59873:src/proxy.ts` has no hit, so
it does not predate the branch; `git log -S` puts it in `ba6cc99` and out in `0dc363f`. Every repo
hit of the identifier is a `LocalProviderModel`/registry/patch-meta read or comes from
`body.max_tokens` — no destructure, no index access, nothing reading it off a `ProxyRoute`. The one
route spread (`src/http-proxy/routes.ts:88`) copies it and nothing downstream reads the copy.

## One theme left, and it is coverage rather than correctness

Four one-line invariants this PR establishes are unpinned — each survives the **full** 1914-test
suite when deleted, with `typecheck` clean:

| Deletion | Full suite |
| --- | --- |
| both `pricingBoundary: selectedModel.pricingBoundary` lines in `src/cli.ts` | 1914/1914 green |
| `sessionStops` defaulted to `true` (patch path goes session-aware) | 1914/1914 green |
| the `!parsed.showHelp && !parsed.showVersion` validation skip | 1914/1914 green |
| the non-zero propagation from the moved launch report | 1914/1914 green |

The first is the one I would actually fix. **It is the exact line that was MINOR 1**: the composition
test pins the wrapper carrying `sdk.pricingBoundary` onto the route, but nothing observes `cli.ts`
putting it into `sdk` in the first place — which is where the field was missing. A regression there
reproduces the original defect and stays green.

The second has the same shape one level up: the saved-only rule round 1 asked for now has an option
that can switch it off, and no detector fires when it does. Cheapest pin, as one of my lenses put
it: hash `buildDesiredPatchConfig().config` either side of a `--context` run in the test you already
wrote, and assert equality.

Related and pre-existing: `singleModelProxyRoute` pins `pricingBoundary` and `providerId`; three more
(`authType`, `compatibility`, `headers`) are caught by `tests/proxy.test.ts`; **nine survive suite-wide**
— `npm`, `baseURL`, `oauthAccountId`, `providerData`, `supportedParameters`, `reasoning`,
`interleavedReasoningField`, `useResponsesLite`, `preferWebSockets`. That is the failure mode the
function's own doc comment describes, and the comment reads as though exporting the function closed
it. A single all-fields assertion would.

None of this blocks the merge — the behaviour is proven correct, the feature is opt-in, and I would
rather land it than hold it for regression pins. Taking them as a follow-up.

## Demotions, stated so you do not re-derive them

- A lens filed the `sessionStops` gap as MAJOR. **Demoted to MINOR**: I executed the patch path both
  ways and the hash does not move on a session stop, so the exposure is a future regression, not a
  live defect.
- A lens filed that the new hint is not copy-pastable because the displayed line ends in a period
  (`--save.` → `Unknown models option`). **Dropped**: that is ordinary sentence punctuation, and the
  hint it replaced ended the same way.
- Your body still says `1870/1873` while your comment says `1881/1884`. Not filing it — the comment
  carries the current head and I matched it there.

## Still unobserved, by both of us

Nobody has driven a real provider payload across the boundary at this head. The `onPromptTokens`
seam is exercised and the arithmetic checks out, but "a real request crossed and the user saw the
line" remains stubbed. Unchanged from last round, and stated again so it does not quietly become
assumed.

Merging. Thanks — three rounds, and each one closed what was asked plus the thing behind it.

PR #141, opened 2026-08-22
feat(auth): add browser sign-in for workspaces that disable device codes

### LINKED ISSUE #140: OAuth sign-in fails when the ChatGPT workspace disables device code authorization
## Problem

`clodex providers auth openai` only offers the device-code flow. Some ChatGPT workspaces have **"Enable device code authorization for Codex"** turned off in org settings (OpenAI recommends caution with it because device codes can be phished). In such a workspace the sign-in fails with a message like "Please contact your workspace admin to enable device code authentication", which makes the `openai-oauth` provider unusable entirely.

### The Disabled Org setting(screenshot is from my personal account)

Image

## Proposal

Add an opt-in browser sign-in fallback:

```
clodex providers auth openai --browser
```

A standard PKCE flow through the user's browser, like the Codex CLI's `codex login` (same OAuth client), so it works even when device codes are disabled. Device code stays the default since it works over SSH/VPS where a local browser isn't available.

I have a working implementation and can open a PR.

dgokcin:
## Summary
Some ChatGPT workspaces disable device code authorization for Codex, which fully blocks the
ChatGPT/Codex-plan provider in clodex (`clodex providers auth openai` fails outright). This PR
adds an opt-in browser sign-in as an alternative, plus an interactive picker so users can choose
between device code and browser sign-in from the wizard and provider menus.

## Changes
- Add a PKCE browser-based OAuth flow for OpenAI (`src/oauth/openai.ts`) as an alternative to
  device-code auth
- Extend the local callback server with configurable ports, path, and hostname
  (`src/oauth/callback-server.ts`)
- Add a `--browser` flag to `clodex providers auth openai` and wire `method: 'browser'` dispatch
  through `authenticateProvider`
- Add an interactive `promptOAuthMethod()` picker (device code vs. browser), wired into the
  first-run wizard, `providers add`, the provider detail view, and both hub sign-in flows
- Update hints, README, and internal docs to describe both sign-in options
- Add tests for the browser flow, callback server, CLI flag parsing, and the new method picker

## Additional Notes
- Closes #140
- Device code stays the default — it works over SSH/VPS where a local browser isn't available
- `pnpm typecheck && pnpm test && pnpm build` are green; the one failing test
  (`credential-helper.test.ts`) is a pre-existing timing flake tracked in #138, unrelated to this
  change
- Verified end-to-end against a real ChatGPT workspace with device code authorization disabled

### DIFF
--- a/.claude/docs/registry-and-server.md
+++ b/.claude/docs/registry-and-server.md
@@ -7,7 +7,11 @@
 `src/registry/`. The shipped registry has three provider templates in `src/provider-templates.ts`:
 `openai` (API key), `openai-oauth`, and `opencode-go`. `provider-auth.ts` implements the OpenAI
 OAuth sign-in (device code by default; browser PKCE via `--browser` for workspaces that disable
-device codes); `refresh-models.ts` fetches the model list (3-tier fetch for OAuth).
+device codes). Interactive entry points that don't take flags — the providers hub's auth and
+add-account actions, the provider detail menu, `providers add`'s OAuth path, and the first-run
+wizard — ask instead through `promptOAuthMethod` (`providers-command.ts`), a device-code/browser
+picker whose Enter default is device code; the chosen method is forwarded to
+`authenticateProvider`. `refresh-models.ts` fetches the model list (3-tier fetch for OAuth).
 Materialization (`materialize.ts`) turns registry providers into `LocalProvider`s with per-model
 `npm`/`baseUrl`/`upstreamModelId`.
 
--- a/.claude/docs/registry-and-server.md
+++ b/.claude/docs/registry-and-server.md
@@ -6,7 +6,8 @@
 
 `src/registry/`. The shipped registry has three provider templates in `src/provider-templates.ts`:
 `openai` (API key), `openai-oauth`, and `opencode-go`. `provider-auth.ts` implements the OpenAI
-device-code OAuth flow; `refresh-models.ts` fetches the model list (3-tier fetch for OAuth).
+OAuth sign-in (device code by default; browser PKCE via `--browser` for workspaces that disable
+device codes); `refresh-models.ts` fetches the model list (3-tier fetch for OAuth).
 Materialization (`materialize.ts`) turns registry providers into `LocalProvider`s with per-model
 `npm`/`baseUrl`/`upstreamModelId`.
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ clodex claude                  # 5. launch Claude Code on an OpenAI model
 ```
 
 1. **Install** — puts the `clodex` command on your PATH.
-2. **Sign in** — opens a device-code OAuth flow for your ChatGPT/Codex plan; the token is stored in your OS credential store. (API-key users: `clodex providers add` instead.)
+2. **Sign in** — opens a device-code OAuth flow for your ChatGPT/Codex plan; the token is stored in your OS credential store. If your workspace admin has disabled device code authorization, add `--browser` to sign in through your browser instead. (API-key users: `clodex providers add` instead.)
 3. **Pick models** — an interactive manager for favorites (max 20) and short aliases like `sol` so you do not need to type the long names. Favorites drive the `/model` switch menu, proxy-mode routing, and the patcher.
 4. **Patch** *(optional but recommended for proxy mode)* — bakes your favorites and aliases into the Claude Code binary so they pass model validation, appear in `/model`, and report their real context windows. Re-run after each `claude` update; `clodex patch --restore` undoes it. This step is required if you want to use clodex-routed models as subagents via the Agent tool.
 5. **Launch** — starts Claude Code bridged to the model you choose.
@@ -273,7 +273,7 @@ Two things worth knowing about the numbers:
 | --- | --- |
 | *(none)* | Provider hub wizard |
 | `add` | Add OpenAI or OpenCode Go with an API key, or sign in with ChatGPT |
-| `auth openai` | Sign in with ChatGPT/Codex-plan OAuth (device code) |
+| `auth openai` | Sign in with ChatGPT/Codex-plan OAuth (device code; `--browser` for workspaces that disable device codes) |
 | `list` | Show configured providers |
 | `remove <id>` | Remove a provider by id |
 | `refresh-models [id]` | Update cached model lists |
--- a/src/first-run.ts
+++ b/src/first-run.ts
@@ -4,7 +4,7 @@ import pc from 'picocolors';
 import * as p from '@clack/prompts';
 import { printWelcomePanel } from './ui.js';
 import { loadRegistry } from './registry/io.js';
-import { runProvidersAdd, runProvidersAuth } from './providers-command.js';
+import { promptOAuthMethod, runProvidersAdd, runProvidersAuth } from './providers-command.js';
 
 export type FirstRunResult = 'continue' | 'cancel';
 
@@ -24,7 +24,7 @@ export async function runFirstRunWizard(_trace = false): Promise<FirstRunResult>
       {
         value: 'oauth',
         label: pc.cyan('Sign in with ChatGPT (Plus/Pro plan)'),
-        hint: 'OAuth device code — uses your ChatGPT/Codex plan',
+        hint: 'OAuth (device code or browser) — uses your ChatGPT/Codex plan',
       },
       {
         value: 'apikey',
@@ -38,9 +38,14 @@ export async function runFirstRunWizard(_trace = false): Promise<FirstRunResult>
     return 'cancel';
   }
 
-  const code = choice === 'oauth'
-    ? await runProvidersAuth('openai')
-    : await runProvidersAdd();
+  let code: number;
+  if (choice === 'oauth') {
+    const method = await promptOAuthMethod();
+    if (method === null) return 'cancel';
+    code = await runProvidersAuth('openai', method);
+  } else {
+    code = await runProvidersAdd();
+  }
   if (code !== 0) return 'cancel';
 
   if ((await needsFirstRunSetup())) return 'cancel';
--- a/src/oauth/callback-server.ts
+++ b/src/oauth/callback-server.ts
@@ -3,7 +3,7 @@
 // This is only used when running `clodex providers auth <provider>` without the GUI.
 
 import http from 'node:http';
-import { listenTcpServer } from '../listener-ready.js';
+import { listenTcpServer, waitForTcpListener } from '../listener-ready.js';
 
 export interface CallbackParams {
   code: string;
@@ -45,6 +45,16 @@ const FAILURE_HTML = `<!DOCTYPE html><html><head><meta charset="utf-8"><title>Si
 <p style="color:#666">Return to the terminal for details.</p>
 </div></body></html>`;
 
+const LOOPBACK_PROBE_TIMEOUT_MS = 250;
+
+async function isLoopbackPortTaken(port: number): Promise<boolean> {
+  const probes = await Promise.all(['127.0.0.1', '::1'].map(host =>
+    waitForTcpListener(host, port, LOOPBACK_PROBE_TIMEOUT_MS)
+      .catch(() => false),
+  ));
+  return probes.some(Boolean);
+}
+
 export async function startCallbackServer(options: CallbackServerOptions): Promise<CallbackServer> {
   let codeResolve: ((p: CallbackParams) => void) | undefined;
   let codeReject: ((e: Error) => void) | undefined;
@@ -77,6 +87,15 @@ export async function startCallbackServer(options: CallbackServerOptions): Promi
   let address: Awaited<ReturnType<typeof listenTcpServer>> | undefined;
   let lastError: unknown;
   for (const port of ports) {
+    // 'localhost' binds one loopback family, but the browser may resolve the
+    // other; a foreign listener there would swallow the redirect while our
+    // bind "succeeds". Treat a port answering on either family as taken.
+    if (port !== 0 && redirectHost === 'localhost' && await isLoopbackPortTaken(port)) {
+      const busy: NodeJS.ErrnoException = new Error(`listen EADDRINUSE: address already in use localhost:${port}`);
+      busy.code = 'EADDRINUSE';
+      lastError = busy;
+      continue;
+    }
     try {
       address = await listenTcpServer(server, port, redirectHost);
       break;
--- a/src/oauth/callback-server.ts
+++ b/src/oauth/callback-server.ts
@@ -11,6 +11,15 @@ export interface CallbackParams {
   error?: string;
 }
 
+export interface CallbackServerOptions {
+  /** Fixed ports to try in order; default [0] (ephemeral). */
+  ports?: readonly number[];
+  /** Only accept this callback path; default '/callback' and '/oauth/callback'. */
+  path?: string;
+  /** Hostname used in redirectUri; default '127.0.0.1'. */
+  redirectHost?: string;
+}
+
 export interface CallbackServer {
   port: number;
   redirectUri: string;
@@ -26,13 +35,16 @@ const SUCCESS_HTML = `<!DOCTYPE html><html><head><meta charset="utf-8"><title>Au
 <p style="color:#666">You can close this tab and return to the terminal.</p>
 </div></body></html>`;
 
-export async function startCallbackServer(): Promise<CallbackServer> {
+export async function startCallbackServer(options: CallbackServerOptions = {}): Promise<CallbackServer> {
   let codeResolve: ((p: CallbackParams) => void) | undefined;
   let codeReject: ((e: Error) => void) | undefined;
+  // The browser can hit the callback before waitForCallback arms the promise.
+  let buffered: CallbackParams | undefined;
 
+  const acceptedPaths = options.path ? [options.path] : ['/callback', '/oauth/callback'];
   const server = http.createServer((req, res) => {
     const u = new URL(req.url ?? '/', 'http://localhost');
-    if (u.pathname !== '/callback' && u.pathname !== '/oauth/callback') {
+    if (!acceptedPaths.includes(u.pathname)) {
       res.writeHead(404); res.end(); return;
     }
     const code = u.searchParams.get('code') ?? '';
@@ -40,21 +52,41 @@ export async function startCallbackServer(): Promise<CallbackServer> {
     const error = u.searchParams.get('error') ?? '';
     res.writeHead(200, { 'Content-Type': 'text/html; charset=utf-8' });
     res.end(SUCCESS_HTML);
-    codeResolve?.({ code, state, error: error || undefined });
+    const params: CallbackParams = { code, state, error: error || undefined };
+    if (codeResolve) codeResolve(params);
+    else buffered ??= params;
   });
 
-  const address = await listenTcpServer(server, 0, '127.0.0.1');
+  const ports = options.ports?.length ? options.ports : [0];
+  let address: Awaited<ReturnType<typeof listenTcpServer>> | undefined;
+  let lastError: unknown;
+  for (const port of ports) {
+    try {
+      address = await listenTcpServer(server, port, '127.0.0.1');
+      break;
+    } catch (error) {
+      lastError = error;
+    }
+  }
+  if (!address) throw lastError ?? new Error('OAuth callback server could not bind');
+
+  const redirectHost = options.redirectHost ?? '127.0.0.1';
   return {
     port: address.port,
-    redirectUri: `http://127.0.0.1:${address.port}/callback`,
+    redirectUri: `http://${redirectHost}:${address.port}${options.path ?? '/callback'}`,
     waitForCallback(timeoutMs = 300_000) {
       return new Promise<CallbackParams>((resolve, reject) => {
-        codeResolve = resolve;
-        codeReject = reject;
-        setTimeout(
+        if (buffered) {
+          resolve(buffered);
+          buffered = undefined;
+          return;
+        }
+        const timer = setTimeout(
           () => reject(new Error('OAuth timeout — browser closed without completing sign-in')),
           timeoutMs,
         );
+        codeResolve = params => { clearTimeout(timer); resolve(params); };
+        codeReject = err => { clearTimeout(timer); reject(err); };
       });
     },
     close() { server.close(); codeReject?.(new Error('Server closed')); },
--- a/src/oauth/callback-server.ts
+++ b/src/oauth/callback-server.ts
@@ -14,10 +14,12 @@ export interface CallbackParams {
 export interface CallbackServerOptions {
   /** Fixed ports to try in order; default [0] (ephemeral). */
   ports?: readonly number[];
-  /** Only accept this callback path; default '/callback' and '/oauth/callback'. */
-  path?: string;
-  /** Hostname used in redirectUri; default '127.0.0.1'. */
-  redirectHost?: string;
+  /** The callback path registered with the provider; the only path accepted. */
+  path: string;
+  /** Loopback hostname used in redirectUri and bound by the listener. */
+  redirectHost: 'localhost' | '127.0.0.1';
+  /** Only accept callbacks carrying this OAuth state. */
+  expectedState?: string;
 }
 
 export interface CallbackServer {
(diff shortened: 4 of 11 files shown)


bman654 (changes requested):
**Changes requested — three test-only asks. No production-code change is blocking.** A four-lens
panel drove the local callback, CLI, and credential flow end to end (real Chrome, real ports, real
CLI under a pty) and found no live defect in that code. We could not repeat the contributor's
vendor-backed end-to-end run because we do not have a workspace with device-code authorization
disabled. What the panel did find is that several properties this PR is *for* can be deleted with
the suite staying green, including the interactive picker's method forwarding.

Reviewed head `8e2ca5b345d88c845dac1bbafce9a1964585f465`, base `8350dd6`.

## What this does well

- **It unblocks a total outage, not a nicety.** On a workspace with device-code authorization
  disabled the ChatGPT/Codex-plan provider is simply unusable today. Opt-in with device code still
  the default is the right shape.
- **The default really is untouched.** Verified by execution, not by reading: `clodex providers auth
  openai` with no flag goes straight to the device-code panel with no new prompt, and the picker's
  `initialValue: 'native'` means Enter keeps today's behaviour at every interactive site.
- **The PKCE construction is correct.** 64-char verifier over a 64-character unreserved alphabet —
  `b % 64` with `b ∈ [0,256)` is exactly four whole cycles, so no modulo bias — S256 challenge,
  32-byte CSPRNG state. Executed against a stubbed token endpoint: `S256(verifier) === challenge`,
  and `redirect_uri` is identical across the authorize URL and the token exchange.
- **Validation order is right and fails closed.** error → missing code → state → exchange. On the
  state-mismatch, provider-error and missing-code paths we measured zero `fetch` calls, nothing
  persisted, and the port free again immediately.
- **The `clearTimeout` wiring in `waitForCallback` is sound callback-server cleanup.** In an
  isolated harness it lets the event loop drain 82 ms after a successful callback; reverting it to
  `main`'s shape leaves a live 300 s timer. I am not treating that as a demonstrated CLI hang,
  because the CLI explicitly exits when `main()` resolves (`src/cli.ts:1792-1794`).
- **Port fallback 1455 → 1457 actually works,** including the subtle part: re-`listen()` on the same
  `http.Server` after an `EADDRINUSE` rejection, and after a readiness-failure close. Two reviewers
  initially believed that was broken; instrumenting it showed the code is right and they were
  looking at their own leftover listeners.
- `--account <name>` composes with `--browser` and leaves the default credential slot alone; the
  browser credential is identically shaped to the device-code one, `accountId` included, so it flows
  to the `ChatGPT-Account-Id` header through the same code.
- Nothing logs, prints or persists the code, the verifier or the tokens.
- **The PR title is one of the better summary lines in this repo** — user-visible outcome, why it
  matters, no internal vocabulary, 72 chars.
- **Your note about `credential-helper.test.ts` is accurate and nobody should hold the PR on it.**
  We checked both sides: the branch is based on `8350dd6`, which predates #138's fix; with `main`'s
  version of that test applied the failure disappears (8/8 vs 5/8). Merging `main` moots it, and per
  our own convention that is our job, not yours — don't rebase on our account.

## Blocking

### 1. `tests/oauth-openai.test.ts:136-148` binds the real fixed ports 1455 and 1457

This is the only test that binds predetermined, usable fixed ports; the rest use ephemeral ports,
a previously allocated ephemeral port, or an intentionally invalid port. These are also the ports
used by this browser sign-in flow and Codex login.

Reproduced during synthesis, on an otherwise idle box with both ports verified free first:

```
BASELINE                              Tests  18 passed (18)
with one squatter on 127.0.0.1:1455   × runOpenAiBrowserFlow > reports busy ports instead of a raw bind error
                                        → listen EADDRINUSE: address already in use 127.0.0.1:1455
                                      Tests  1 failed | 17 passed (18)
```

A second panel member reproduced it independently for both ports. So `pnpm test` goes red — with a
setup error, not an assertion — for anyone running the suite while `codex login` is open, or while
an earlier `clodex providers auth openai --browser` they cancelled is still holding the port (a
cancelled sign-in keeps it until its 300 s timeout).

Same test, second defect: `const blockers = await Promise.all(...)` sits **outside** the `try`. If
one bind fails, `blockers` is never assigned, the `finally` never runs, and the server that *did*
bind leaks for the rest of the vitest worker's life. Reproduced: `Promise.all rejected: EADDRINUSE
1455 / created 1457: listening=true`.

**Smallest fix:** bind two ephemeral blockers (`listen(0)`) and pass them through the `ports` option
your own function already exposes — `runOpenAiBrowserFlow(vi.fn(), { ports: [p1, p2] })`. The thrown
message is hard-coded to `BROWSER_CALLBACK_PORTS` regardless of `opts.ports`, so the
`/1455 and 1457 are in use/` assertion still holds unchanged. For robust cleanup, create and retain
the server objects before awaiting their listens, then close every object in `finally`; merely
moving the existing `Promise.all` inside `try` would still lose the partially bound server.

### 2. Nothing pins that the method the user picks reaches `authenticateProvider`

`src/providers-command.ts:669-671`, `:840-843`, `:1002-1005`, `:1022-1025`, and
`src/first-run.ts:43-45`.

Reproduced during synthesis: I replaced all five forwarded values with `undefined` — the picker
still renders, the user's answer is silently discarded — and ran the three files that could
plausibly cover it:

```
BASELINE   tests/providers-command.test.ts + provider-auth.test.ts + first-run.test.ts   81 passed (81)
MUTATED    (all five method values -> undefined)                                          81 passed (81)
```

Two panel members reproduced the same at full-suite scale (1921/1922 both sides, the only failure
being the #138 flake above). That deletes the picker commit's user-visible effect with no test noticing,
and the failure it leaves open is silent and lands on exactly the users this PR exists for: they
pick "Browser", and they get a device code.

The one test line the picker commit adds — `.mockResolvedValueOnce('native')` at
`tests/providers-command.test.ts:982` — is queue alignment so an unrelated cleanup test keeps
working now that an extra prompt exists. It asserts nothing about the method. The same gap leaves
`initialValue: 'native'` unpinned: flipping it to `'browser'`, i.e. inverting the "Enter keeps the
device-code default" promise in the function's own doc comment, is also green.

**Smallest fix:** one test in `tests/providers-command.test.ts` that drives the hub `auth-menu` with
the existing select mock returning `'browser'` and asserts `authenticateProviderMock` was called
with `{ method: 'browser' }`. That kills the discard mutation at that call site. I am not asking for
one test per entry point; if you also want to pin the Enter default, assert the picker is configured
with `initialValue: 'native'` separately — browser/native return mirrors do not catch that mutation.

### 3. The PKCE verifier→challenge binding is not asserted

`src/oauth/openai.ts:204`, `tests/oauth-openai.test.ts:119`.

Reproduced during synthesis: I changed the exchange to send `challenge` instead of `verifier` — the
classic way PKCE breaks — and the file stayed **18/18 green**. The only assertion on it is
`expect(body.get('code_verifier')).toBeTruthy()`, which any non-empty string satisfies: the
challenge, a fresh random verifier, a constant.

Current behaviour is correct — this is regression protection for the one property that makes the
flow safe, and the failure mode of losing it is 100% of browser sign-ins failing at the vendor with
an opaque `OpenAI token exchange failed (400)`.

**Smallest fix**, in `exchanges the callback code for tokens`:

```ts
const verifier = body.get('code_verifier')!;
expect(createHash('sha256').update(verifier).digest('base64url'))
  .toBe(new URL(seenUrl).searchParams.get('code_challenge'));
```

A panel member ran exactly this assertion: green on your code, red on the mutation.

## Non-blocking notes

1. **Every callback-server failure is reported as "Ports 1455 and 1457 are in use"**
   (`src/oauth/openai.ts:189-193`). The bare `catch {}` also swallows `listenTcpServer`'s other throws —
   `TCP server did not bind to a network address` and `TCP listener did not become reachable within
   1000ms` (the readiness probe, reachable on a loaded machine or with loopback-filtering security
   software) — plus `EACCES`/`EADDRNOTAVAIL`. The user is then sent to close a `codex login` that
   isn't running. `catch (err)` and append `err.message`, or only substitute the friendly text when
   `code === 'EADDRINUSE'`.

2. **The token exchange hand-rolls `fetch`** (`src/oauth/openai.ts:148-167`) where
   `postOAuthRefresh` (`src/oauth/refresh-http.ts:13-57`) already provides a 30 s timeout and consumes
   or cancels non-ok bodies. A stalled token endpoint can therefore leave this flow waiting without
   a local timeout, and a failed exchange currently reports only the status. Consider reusing the
   helper for timeout and cleanup. We did not observe a real failing `auth.openai.com` token response,
   so before exposing its body, please confirm what it contains and whether it is safe and actionable.
   (The pre-existing device-code path at `src/oauth/openai.ts:90-104` has the same raw-fetch shape, so
   this is consistency with an established helper rather than a regression from the sibling path.)

3. **The browser says "Authentication successful" even when sign-in failed.**
   `src/oauth/callback-server.ts:53-54` writes the success page before any validation. Measured off
   the wire: state mismatch, `error=access_denied`, and a missing code all render the green check
   while the terminal reports the error. Render an error page when `error` is present.

4. **`close()` doesn't drop live sockets** (`src/oauth/callback-server.ts:92`). A socket that is
   connected but has not completed a request at that instant (browser preconnect, loopback scanner,
   endpoint-security agent) keeps the event loop alive with no timer to reap it — measured still
   alive at 330 s, unbounded. Four real Chrome runs never triggered it, so this is hardening. If you
   address it, stop accepting first and then reap existing sockets: `server.close();
   server.closeAllConnections();` (Node ≥ 18, fine under `engines.node >= 22`).

5. **Address-family mismatch between the free-port check and the advertised redirect.** The check
   binds `127.0.0.1` (`src/oauth/callback-server.ts:65`) while the redirect advertises `localhost`
   (`src/oauth/openai.ts:187`), which resolves to `::1` first on the reviewed macOS host. There, a
   process on `[::1]:1455` or a coexisting `[::]:1455` listener keeps the IPv6 side and receives the
   callback while clodex's IPv4 bind "succeeds" and
   then times out for five minutes — reproduced with real Chrome. Two things bound the severity, and
   they are why this isn't blocking: PKCE means the other listener cannot redeem the code (this is a
   hang, not a credential leak), and the concrete competitor your error message names does not
   trigger it — we ran `codex-cli 0.149.0 login` and it binds `127.0.0.1:1455`, which your fallback
   correctly detects and skips to 1457. If you want to harden it: treat a reachable listener on
   `[::1]:<port>` as taken (`waitForTcpListener` already exists), or bind both loopback families.

6. **`promptOAuthMethod` prints a session-closing "Cancelled." from inside loop callers.**
   `src/providers-command.ts:419` calls `p.cancel('Cancelled.')` unconditionally, but three of the
   five callers don't end the session — they `continue` the hub loop or `return 'back'`. Real
   terminal bytes show the program printing its closing `└ Cancelled.` bar and then immediately
   redrawing the menu; every pre-existing prompt in that same loop cancels silently. Keep the shared
   picker silent and render cancellation only from a top-level path that actually exits. The shared
   `providers add` flow is also reachable from the hub, so it needs caller context rather than an
   unconditional `p.cancel` of its own.

7. **The Browser option's explanation is invisible until you arrow onto it.** Clack renders the hint
   only for the active option, so the first frame is `● Device code (works everywhere, including
   SSH/VPS)` and a bare `○ Browser` — the sentence that tells the affected user this is their option
   isn't on screen. Consider folding it into the label.

8. **`--browser --native` together is accepted, last-wins.** Defensible, but
   `Use either --native or --browser, not both` would be clearer.

9. **Small doc/help gaps.** `clodex providers --help`'s Usage block doesn't carry the `--browser`
   line that `clodex providers auth` does; `.claude/docs/registry-and-server.md:9-10` describes the
   flag but not the picker at five call sites, and `promptOAuthMethod` appears nowhere in `.claude/`.

10. **Three more hunks red nothing when deleted** — the port-fallback loop (`ports.slice(0,1)` is
    green), the `clearTimeout` fix itself, and the `buffered` early-arrival fast path. The first two
    are worth a test if you're already in the file. For `buffered`: nobody could construct a
    production path that reaches it, since `runOpenAiBrowserFlow` arms `waitForCallback` in the same
    tick as `onAuthorizeUrl` and `openBrowser` doesn't await — either reach it in a test or note in
    a comment that it's deliberate belt-and-braces.

## Could not verify — asking rather than prescribing

The comment at `src/oauth/openai.ts:13-15` says `http://localhost:{1455|1457}/auth/callback` are the
only redirect URIs registered for `app_EMoamEEZ73f0CkXaXp7hrann`, and that anything else is rejected
by `auth.openai.com`. **We could not check that.** Every direct probe of the authorize endpoint
returns a Cloudflare interstitial (`HTTP 403`, "Just a moment...") — for a valid port, the fallback
port, and a deliberately bogus port alike — so the endpoint gives no signal either way. The best
independent evidence we found is OpenAI's own Codex source at `rust-v0.149.0`
(`codex-rs/login/src/server.rs`), which uses the same client id, builds
`http://localhost:{port}/auth/callback`, and labels 1457 as the fallback that must stay in sync with
the redirect allow-list. That corroborates your values; it does not establish the negative claim.

Two questions, and we'd rather encode your observation than guess:

1. **Did your end-to-end run ever exercise 1457, or only 1455?** The first free port wins, so a clean
   machine takes 1455 — which would mean the entire busy-port fallback rests on the vendor claim
   rather than on an observed success.
2. **Did you actually see `auth.openai.com` reject a `127.0.0.1` redirect URI, or is `localhost`
   chosen to match Codex?** That answer decides whether the hardening in note 5 is free or
   vendor-constrained. If you have the rejection, paste it.

## Checked and cleared

Stated out loud so you don't re-derive them:

- **`clodex providers auth openai` gains no prompt** — driven for real under a pty, straight to the
  device-code panel. Enter is device code at every picker.
- **Cancel semantics are correct at all five sites,** traced to the shell exit code. Ctrl-C behaves
  as Esc; exit 0 on user cancel matches this file's existing convention, and the typed account name
  in the `auth-account` path is correctly discarded because nothing was persisted.
- **No new non-TTY exposure.** Every picker site is already preceded by an existing clack prompt on
  the same path, so `p.select`'s `ERR_TTY_INIT_FAILED` would already have thrown before the new
  prompt is reached; the one non-interactive path never prompts.
- **Port fallback works,** including re-`listen()` on the same `http.Server` after a rejected bind
  and after a readiness-failure close, with no leaked `error` listeners.
- **No partial credential on any failure path;** every error path closes the server and releases the
  port; the browser credential is shape-identical to the device-code one.
- **`--account` + `--browser` doesn't touch the default slot** (executed).
- **Duplicate, out-of-order and non-matching callback hits don't poison the flow;** a 404 path
  returns before touching `buffered`; `close()` rejecting an already-settled promise is a no-op
  (checked under `--unhandled-rejections=strict`).
- **A local process can't inject a code** — no `state`, no exchange. For the same reason, the
  handler not checking the `Host` header is defence-in-depth only, not an exploit.
- **No OIDC `nonce` is needed** — the `id_token` arrives over the back channel.
- **The callback timer is cleared after success.** An isolated callback-server harness drains
  promptly; the CLI itself also has an explicit `process.exit` after `main()` resolves, so we did not
  establish a user-visible 300 s CLI hang from the old timer shape.
- **The `timeoutMs: 50` test is not a flake risk** — 15/15 serial, plus 20 earlier and 6 concurrent
  runs. Every failure we saw in that file was the fixed-port test.
- **The picker is unreachable for `opencode-go`** and behaves correctly on the API-key `openai`
  provider's pre-existing "Sign in again (OAuth)" action.
- **README, help text and the SSH caveat all match the code as written.**
- **`pnpm typecheck` and `pnpm build` clean**, and the change writes nothing into `CLODEX_HOME`.

dgokcin:
Thanks @bman654 for the review. Pushed 7 commits (head `0debdc7`) addressing all three blocking asks, plus non-blocking notes 1, 2, 3, 5, and 9.

**Blocking**

1. Busy-port test: now binds two ephemeral blockers and passes them through `ports`. Server objects are created before their listens are awaited and closed in one `finally`, so a partial bind cannot leak. The suite no longer touches 1455/1457.
2. Picker forwarding: new tests drive the hub `auth-menu` and assert `authenticateProvider('openai', { method: 'browser' })`, and pin `initialValue: 'native'`. Your discard mutation now fails 2 tests; flipping the default fails 1.
3. PKCE binding: the exchange test asserts `sha256(code_verifier)` (base64url) equals the advertised `code_challenge`. Sending the challenge instead of the verifier goes red.

**Non-blocking taken**

- Note 1: only `EADDRINUSE` gets the busy-ports message (now naming the ports actually tried); other bind errors propagate with the real message and `cause`.
- Note 2: the token exchange reuses `postOAuthRefresh` (30s abort, body cleanup). It does not expose the response body, per your caution.
- Note 3: error or codeless callbacks render a failure page instead of the green check.
- Note 5: before binding a fixed port for a `localhost` redirect, both loopback families are probed with the existing `waitForTcpListener`; a listener on `[::1]:<port>` counts as taken and the flow falls through to the next port. Pinned by a test that squats the loopback family the bind does not take.
- Note 9: `providers --help` Usage carries `--browser`, and `.claude/docs/registry-and-server.md` documents the picker and `promptOAuthMethod`.

Also on this branch: mismatched-state callbacks no longer settle the flow. The callback server rejects them with 400 and keeps waiting, so a stale tab or concurrent sign-in cannot kill a legitimate login. The in-flow state check stays as defense in depth.

**Your two questions, both now observed**

1. Yes, 1457 is exercised end to end. With 1455 held on both `127.0.0.1` and `::1`, the flow bound 1457 and `auth.openai.com` delivered a real authorization code to `http://localhost:1457/auth/callback?code=ac_...&state=...`. Sign-in completed.
2. Yes, `auth.openai.com` rejects a `127.0.0.1` redirect URI. The same authorize URL with only the redirect host changed to `127.0.0.1` returns `{"error": {"message": "Invalid authorize request", "type": "invalid_request_error", "code": "invalid_authorize_request"}}`, while the `localhost` variant renders the consent page. The error is generic rather than naming `redirect_uri`, but the host was the only variable changed. So `localhost` is vendor-constrained, which is why note 5 is handled by probing both families instead of changing the advertised host.

Verification: focused file 21/21, full suite 1927/1927 (the `credential-helper` flake passed this run), `pnpm typecheck` and `pnpm build` clean. Environment: macOS (darwin 24.6.0), Node 24, no proxy vars, throwaway `CLODEX_HOME` for the live run.

bman654 (approved):
**Approved.** All three round-1 blocking asks are closed, and we verified each one by mutation
rather than by reading your diff. The unasked callback-state filter and your voluntary fix for
round-1 note 5 are correct; deleting either core behavior makes a named test go red. Nothing in the
delta blocks. One non-blocking note about what the terminal says after a stale-tab callback, and one
test-coverage nit.

Reviewed head `0debdc7411700ac6fe55122ff513a4b7b1b0feca`, delta `8e2ca5b3..0debdc74`. Gates re-run
here on a clean tree at that head: `pnpm typecheck` clean, `pnpm test` **1927 passed (102 files)**,
`pnpm build` clean — your reported numbers match ours exactly.

## Round 1 asks — closed

Each verified by breaking the production code and watching a named test go red, so you know we
checked rather than trusted.

**1. The busy-port test binding the real 1455/1457 — closed.** We re-ran the round-1 reproduction
against both heads under an identical squat: four listeners held on `127.0.0.1:1455`, `[::1]:1455`,
`127.0.0.1:1457`, `[::1]:1457`.

```
head 8e2ca5b3 (round-1 head)   × reports busy ports instead of a raw bind error
                                 → listen EADDRINUSE: address already in use 127.0.0.1:1455
                               Tests  1 failed | 17 passed (18)
head 0debdc74 (this head)      Tests  21 passed (21)
```

`grep -rn '1455\|1457' tests/` now finds only the authorize-URL input and matching expectation,
plus your own comment — nothing binds them. The partial-bind leak is closed structurally, and we
proved
it rather than reading it: replicating your setup shape with the second `listen` forced to fail
(`EADDRNOTAVAIL`), both retained server objects are closed by the single `finally`
(`listening = [false, false]`) and the process exits 0 with no live handle.

**2. Picker forwarding — closed.** We re-applied the exact round-1 mutation: `method` replaced with
`undefined` at all five picker call sites (`providers-command.ts:672`, `:844`, `:1006`, `:1026`,
`first-run.ts:45`).

```
mutation "discard the user's pick"                → Tests  2 failed | 46 passed (48)
                                                     × forwards the picked browser method to authenticateProvider
                                                     × keeps device code as the picker default ...
mutation initialValue 'native' -> 'browser'       → Tests  1 failed | 45 passed (46)
```

That matches your report (2 tests / 1 test). The silent failure that round 1 was worried about —
user picks Browser, gets a device code — can no longer ship green.

**3. PKCE verifier→challenge binding — closed.** Sending the challenge in place of the verifier
(`openai.ts:213`) now reds the exchange test on the assertion itself, not on a downstream symptom:

```
× exchanges the callback code for tokens
  → expected 'lrh1fxIqIhIHBpoP-…' to be 'QxnDXZFMxfk6AY7gDk_2…'   Tests  1 failed | 20 passed (21)
```

**Bonus, unasked:** round-1 note 10's first item is closed as a side effect. `ports.slice(0, 1)` —
deleting the port-fallback loop — used to be green and now reds
`treats a port held on the other loopback family as taken and falls back`.

## Your two vendor answers

We asked because we could not measure it; you went and measured it, and that settles both.

1. **1457 is exercised end to end.** With 1455 held on both families, the flow bound 1457 and
   `auth.openai.com` delivered a real code to `http://localhost:1457/auth/callback`. The busy-port
   fallback no longer rests on the redirect-allow-list claim alone — it rests on an observed vendor
   round trip. That was the specific gap in round 1's evidence and it is now filled.
2. **`localhost` is vendor-constrained.** The same authorize URL with only the redirect host changed
   to `127.0.0.1` returns `invalid_authorize_request` while the `localhost` variant renders consent.
   We accept the generic-error caveat you flagged; the host was the only variable. That answer is
   what makes the dual-family probe the right shape instead of "just bind the literal", and we would
   have prescribed the wrong fix without it.

## What the new work adds

The state filter was unasked; the loopback probe voluntarily closes round-1 note 5. Both are the
right call, and we tried to break them.

**Correction to our round-1 review.** We said non-matching callbacks did not poison a valid flow and
praised the old error → missing-code → state validation order. That was wrong: at `8e2ca5b3`, the
first mismatched callback settled the wait and terminated the flow before a valid callback could
arrive. The new server-side, state-first filter fixes that. We are changing our earlier assessment
explicitly rather than silently reversing it.

**Callback-state filter (`callback-server.ts:73-77`).** A callback whose `state` does not match is
answered 400 and the server keeps waiting, so a stale tab or a concurrent sign-in can no longer kill
a legitimate login — round-1's head died on the first forged hit. This matches OpenAI's own client:
`codex-rs/login/src/server.rs` at `rust-v0.149.0` validates state **before** it looks at `error`,
answers `400 State mismatch`, and keeps serving, with no overall wait cap at all. Your in-flow check
at `openai.ts:210` is genuinely unreachable now, and the comment saying so is accurate — that is
fine as defense in depth.

**Dual-family loopback probe (`callback-server.ts:48-56`, `:93-98`).** This is the fix for round-1
note 5, and we checked specifically that it fixes the case rather than moving it. Old vs new in one
process across the four foreign-listener layouts: there is no layout where the new code is worse.
The `[::1]:1455`-occupied case went from a five-minute silent hang (we bound v4, Chrome went to v6)
to a clean fallback to 1457; the `codex login` shape your error message names behaves exactly as
before, 250 ms slower; the both-ports-occupied case went from a hang to an immediate, accurate
error. A real Chrome reaches the new `[::1]` bind through `localhost` and delivers the callback. The
~250 ms happy-path cost sits under the "Opening your browser..." spinner and is not visible.

**`postOAuthRefresh` reuse.** Byte-identical user-visible message (`OpenAI token exchange failed
(400)`), `includeBody` correctly absent per round-1 note 2's caution — against a real endpoint
returning a 200 KB error body, nothing from that body reaches the message — and the 30 s abort now
covers the authorization-code exchange, which previously could hang forever.

## Blocking

None.

## Non-blocking notes

1. **After a stale-tab callback the terminal says the wrong thing for five minutes.** The
   keep-waiting semantics are right; the message that eventually appears is not. Measured on the
   real CLI: with a mismatched-state callback delivered, the browser gets `400 Invalid OAuth state`,
   the spinner never changes, and 302 s later the flow dies with `OAuth timeout — browser closed
   without completing sign-in` — which is false, the browser was never closed. Reproduced here at
   library level (400, then reject at the timeout, `fetch` never called). Reachability is narrow:
   run `--browser`, cancel, run it again, then finish the **older** consent tab. Round-1's head
   failed that in ~0 s with an accurate message; this head hangs. We are not asking you to reverse
   the trade — a stale tab killing a live login is the worse failure, and Codex makes the same
   choice. The cheap fix is to let the timeout name what happened: count the rejected callbacks in
   the handler and build the message when the timer fires
   (`callback-server.ts:119`), e.g. `OAuth timeout — ignored N callback(s) carrying a different
   sign-in state; you probably completed an older browser tab. Run the command again and use the
   newest tab.` Serving `FAILURE_HTML` for the 400 instead of bare text is optional — Codex serves
   plain `State mismatch` there too.

2. **The other-family test pins one direction per machine.** Your test derives the squatted family
   from `dns.promises.lookup('localhost')`, which is what makes it adaptive and correct — on any
   host it reds the one-family regression that is actually harmful *there*. What it cannot see is
   the mirror image: we deleted the `::1` probe and got `21/21` on this host, then re-ran the same
   mutation under `--dns-result-order=ipv4first` and it went red. So a future one-family regression
   ships green on whichever resolver order makes that probe redundant, and CI only runs one order.
   If you want to close it: run that test twice, forcing `ipv4first` and `ipv6first` and restoring
   the order in `finally`. Optional — the code is correct on both orders today.

Round-1 notes 4, 6, 7 and 8 (socket reaping on `close()`, the unconditional `p.cancel('Cancelled.')`
inside loop callers, the Browser hint only rendering when focused, `--browser --native` last-wins)
remain unchanged. They were non-blocking then and they are non-blocking now; we are not re-asking,
and none of them needs to hold this PR.

Your response comment already describes this head and its verification accurately, so we are not
asking you to duplicate that addendum into the PR body. Also, the body's `credential-helper` note
is not stale merely because the flake passed this run: this branch still predates #138's fix. Our
round-1 statement that the diagnosis was accurate still stands.

## Checked and cleared

- **Gates at this head, first-hand:** `pnpm typecheck` clean, `pnpm test` 1927/1927 (102 files),
  `pnpm build` clean, with a throwaway `CLODEX_HOME`. Focused file 21/21.
- **Each principal new production path reds a named test when deleted or disconnected:** delete
  the state filter →
  `ignores a mismatched callback before accepting a valid one`; return the forged params instead of
  ignoring them → same test; invert `failed` → `shows the browser a failure page when the provider
  denies access`; delete `isLoopbackPortTaken` → `treats a port held on the other loopback family
  as taken and falls back`; drop the `EADDRINUSE` discrimination → `reports busy ports instead of a
  raw bind error`; drop `includeStatus` → `throws on a failed token exchange`; drop `expectedState`
  at the call site → the mismatch test. One test per property, each red alone.
- **Regression sweep over what round 1 cleared.** Re-executed, not assumed: duplicate, out-of-order
  and 404-path callbacks still cannot poison a valid flow under the new filter; a local process
  still cannot get a code exchanged (zero `fetch` calls on a forged state); port fallback still
  works, including re-`listen()` on the same `http.Server` after a real `EADDRINUSE` and after the
  new probe-skip `continue`; every failure path still leaves no listener behind (`lsof` clean, and
  the probe-skip path never binds); `--account work --browser` still leaves the default slot alone
  and the browser credential is still shape-identical to the device-code one; the no-flag CLI path
  still goes straight to the device-code panel with zero method prompts; cancel semantics at all
  five picker sites still behave, driven under a real pty; nothing logs or persists the code,
  verifier or tokens — a real 200 KB error body from the exchange endpoint never reaches the
  message.
- **Flake risk of the reworked tests:** 40 focused runs (24 serial, 16 concurrent) and four full-suite
  runs, 44/44 green. The `dns.setDefaultResultOrder('ipv6first')` in the busy-port test is
  process-global but restored in `finally` before any other file can observe it, and the test derives
  its blocker address from the machine's own resolution, so it does not assume a family. The
  other-family test returns early and cleanly on a single-family host.
- **Making `path` and `redirectHost` required breaks nothing.** One production caller, zero test
  callers, and the dropped `'/callback'`/`'/oauth/callback'` defaults have no consumer in this fork.
- **No XSS or reflection on the callback pages.** Success, failure and 400 bodies are constants with
  no interpolation; probed with `code`/`state`/`error`/path all set to `<script>alert(1)</script>`.
- **Considered and not filed:** a security product that answers every loopback connect would make
  the probe report both ports busy where the old code would have bound. We have no evidence such a
  product behaves that way on loopback, and `listenTcpServer`'s pre-existing readiness probe already
  assumes it does not, so we are not turning speculation into an ask.
- **Bind-the-literal is not a hidden gap.** Mutating the bind back to `'127.0.0.1'` while still
  advertising `localhost` stays green under both resolver orders — but with the probe in place it is
  also behavior-preserving in practice (a real browser falls through the unbound family in ~0.3 s),
  so we are not asking for a test that pins it.

PR #144, opened 2026-08-25
fix(patch): restore custom model aliases on Claude Code 2.1.242 and newer

bman654:
`clodex patch` stopped working on Claude Code 2.1.242 and 2.1.243. Every platform, every install:
it printed `required patch failed: PATCH 1: Agent tool model enum` and exited 1, so nobody on a
current Claude Code could get their custom model aliases, per-model context windows or effort
settings baked in. This restores it, and it keeps working on older Claude Code releases exactly as
before.

## What broke

Claude Code 2.1.242 code-split its bundle. Through 2.1.241 the whole ~28 MB of JavaScript was **one**
Bun module; since 2.1.242 the entry module is a ~19,950-byte ESM stub whose body is a list of
`import{…}from"/$bunfs/root/chunk-<hash>.js"` lines, and the code lives in ~1,374 sibling chunk
modules — 1,391 modules in the blob against 15 before.

tweakcc's `readContent` returns only the module it recognizes **by name**, which is that stub. So
`clodex patch` was matching its anchors against 19.9 KB of import statements. `PATCH 1` is the first
required site, so it aborted there and the other ten were never attempted — which is why the canary
reported a single failing site and made this look like one drifted anchor.

It is not an anchor drift. Every clodex anchor is still present in 2.1.243 and still occurs exactly
once across the bundle; they are just spread over six different chunks, none of them the entry.

The hourly patch canary caught this on 2.1.243. 2.1.242 carries the same split and would have failed
identically.

`src/patcher.ts` read the bundle with `readContent` and wrote it with `writeContent`, both of which
address exactly one module. Any user running `clodex patch` against Claude Code 2.1.242 or later hit
this — it is the only supported way to get aliases into the binary, and the failure is total, not
partial.

## The change

**Reading.** `readClaudeBundle` (new `src/bun-bundle.ts`) reads every module Bun will execute as
JavaScript — selected by Bun's **loader id**, so vendored `.js` assets like `mermaid.min.js` and the
`*.node` helpers stay out; feeding megabytes of foreign JavaScript to anchors that must match
exactly once is a way to invent ambiguity and refuse a release that is perfectly patchable — and
joins them with a `/*clodex:module-boundary*/` line. `applyClodexPatches` still receives one document
and **no patch transform changed shape**; `splitBundleSource` cuts it back up afterwards and refuses
a document whose boundary count moved.

Two consequences worth stating plainly, because both are easy to get wrong:

- **A module's position in the blob table is not its position in the bundle**, so every repoint is
  addressed by table index. On every build shipped so far the non-JavaScript modules happen to sort
  last, which makes the two accidentally equal — the test fixture deliberately interleaves them so
  they don't.
- **The read corpus grew for OLD releases too, not just split ones.** A pre-split blob carries five
  ~2 KB loader-1 helper stubs (`image-processor.js`, `audio-capture.js`, `url-handler.js`,
  `computer-use-*.js`) beside the bundle, which tweakcc's read never returned. Measured on a real
  2.1.232: none of them contains any anchor, any `process.env` read, or anything PATCH 5's and
  PATCH 10's whole-bundle count guards look at, so no built-in behaves differently. A **local**
  patch that appends or prepends by position rather than matching an anchor would now land in a
  different module on those releases; that is documented in README where local-patch authors see
  it, and the earlier README wording — which claimed the joined document only applied from 2.1.242 —
  was wrong and is corrected here.

**Writing** is the hard half. `writeContent` puts one buffer into one module, and calling it once per
changed module is not an option: every ELF repack relocates the whole ~290 MB blob to the end of the
binary, so six repacks would leave a multi-gigabyte `claude` and take minutes. (Mach-O and PE assign
in place, so that cost is invisible on macOS and Windows — the same asymmetry that hid the restore
sweep bug.) Instead there is **one repack plus a pointer edit**: every changed module's patched source
is concatenated, NUL-separated, into the single buffer tweakcc writes, and then
`repointBunModuleContents` rewrites the 8-byte `{offset, length}` pair in each module's struct to
address its own slice. Nothing moves and nothing grows — the bytes are already in the blob. After the
repack, `applyBundleWritePlan` refuses if the module tweakcc wrote is not the one the plan was built
around or does not hold exactly the bytes that were written, then reads every repointed module back
and compares it to the source it should be.

**When the only changed module is the one tweakcc writes, the plan degrades to the single call clodex
has always made, with no pointer edit at all** — which is every release up to 2.1.241. Verified: on
2.1.241 the join changes only the `cli` module, so the write is byte-for-byte the old path.

Several refusals and one design change came out of adversarial review rather than the original
design, and each closes a way this could have published something subtly wrong:

- **The separator carries punctuation.** Every built-in anchor bounds its wildcards with a character
  class — `` [^`]*? `` in PATCH 4, `[^{}]*` in PATCH 6, `[^;{}]` in PATCH 5, `[^"]+` in the alias
  arrays, `[^\]]*` in PATCH 7 — and a separator those accept is one they can match ACROSS, binding
  a patch to a span belonging to two modules at once. The boundary COUNT check cannot see that; it
  only sees a boundary added or removed. The marker therefore carries one of each character, inside
  a comment where none of them mean anything to JavaScript. It is
  ``/*clodex:module-boundary`;{}"]*/``.

- **UTF-8 round-trip.** Every module is decoded and re-encoded as UTF-8, so a module stored in any
  other encoding would come back as replacement characters — and nothing downstream could see it,
  because the read-back check and the probe both decode the same way and would agree with each
  other. Measured pure ASCII on every module of 2.1.232 and 2.1.243; this refuses rather than
  repairs.
- **Exactly one module tweakcc would write to.** `writeContent` puts its buffer into *every*
  recognized module, not the first; a second would ship holding the whole concatenation and the
  read-back loop could not see it, because a module nobody repointed is a module nobody checks. No
  published build has two, and the entry-module shim declines to fire when one already exists — so
  this is a refusal, not a repair. It runs **before** the "nothing to repoint" early return, because
  that is the common shape on every release up to 2.1.241 and is not exempt from the problem.
- **The single-module fallback says so.** When the Bun table cannot be read on a native binary,
  clodex falls back to tweakcc's one-module read — and on a split build the result is a
  first-required-site failure that reads exactly like a drifted anchor. A one-line warning turns a
  future canary mystery into a one-glance diagnosis.
- **Order-sensitive `modules-intact` in the probe.** It compared the module lists as sets, so a
  repack that kept every module but shuffled the table read as green. That never mattered before;
  now that modules are addressed by table position, it does.

**Two independent PATCH 4 drifts, in the same family.** 2.1.242 also started concatenating a
conditional sentence onto the Agent tool's `model` description
(``describe(`…`+(fn()?"…":""))``), and both the anchor in `patch-transforms.ts` and its
postcondition proof in `built-in-patch-proofs.ts` required the template to be followed immediately by
`)`. The anchor silently dropped every custom model from that description; the proof turned every
local patch into "could not capture built-in postconditions". Both now end at the template's own
closing backtick, which `[^`]*?` cannot cross anyway.

`PATCH_TRANSFORMS_VERSION` 8 → 9, so existing installs read as `stale-config` and repatch.

`scripts/probe-patch-mechanism.mjs` and `scripts/extract-cc-bundles.mjs` read and write the bundle the
same way now. The extractor mattered independently: it is the tool `.claude/skills/pr-verification`
tells everyone to use before making a claim about Claude Code's behaviour, and on 2.1.242+ it was
silently writing out 0.05% of the code.

**Deliberately not done:** no new binary-format code. All container work — Mach-O segment extension,
the ELF `.bun` relocation, PE, code signing — stays in tweakcc. clodex only writes 8 bytes per
patched module into a table it already parses.

## Discriminating tests

New `tests/bun-bundle.test.ts` (14 cases) and a new end-to-end `applyPatch` case in
`tests/patcher.test.ts` driving a **code-split** fake binary — real Bun blob, entry module
deliberately away from table index 0, anchors spread across three chunks, non-JavaScript modules
interleaved between them, and the tweakcc stand-in reproducing the "whole buffer into every
recognized module" behaviour that makes the repoint necessary. The shared fixture's `describe(` now
carries 2.1.242's spelling.

Mutations run, full-file (not `-t`):

| Removed | Turns red |
| --- | --- |
| multi-module read (`readClaudeBundle` → null) | 7 (all of `bun-bundle.test.ts` + the e2e) |
| the JavaScript-loader filter | 7 — `PATCH 1` becomes an ambiguous anchor and the patch refuses |
| join/split boundary count check | `refuses a patched document that lost a module boundary` |
| the write plan's slice layout (order, NUL, changed-only filter) | 4, 1 and 2 respectively |
| the module-table repoint (`applyBundleWritePlan` → no-op) | 3, incl. the e2e: chunk 0 still holds the pristine enum |
| the "bytes are not what I planned around" refusal | `refuses to repoint when the bytes…` |
| the "more than one recognized module" refusal | `refuses when the blob has more than one module…` |
| the read-back verification loop | `refuses when a repointed module does not read back…` |
| the payload bounds check | `reports no bundle when a payload range runs past the end…` |
| the UTF-8 round-trip check | `reports no bundle when a module is not valid UTF-8` |
| the boundary-already-present check | `reports no bundle when a module already contains…` |
| computing the writable module before the shim instead of after | 2 `applyPatch` cases |
| the separator's punctuation (back to a plain marker) | `keeps the module boundary out of reach…` |
| the "more than one recognized module" refusal for a zero-repoint plan | `refuses more than one module … even when nothing needs repointing` |
| the replacement-character half of the UTF-8 check | `reports no bundle when a module is not valid UTF-8` |
| re-signing after the repoint instead of before it | the e2e: the bytes captured at `codesign` time still hold the pristine enum |
| the single-module fallback warning | `says so when it can only read the one module tweakcc names` |
| PATCH 4 anchor fix | 12 across `patcher.test.ts` and `probe-patch-sites.test.ts` |
| PATCH 4 postcondition proof fix | `captures every successful built-in postcondition…` |

A first pass of this table had **eleven** NONE rows. They are gone because the blob fixture can now
express a non-JavaScript module (`loader`), which is what lets a test tell a table position from a
bundle position — and because the vendored asset in the code-split fixture deliberately carries a
copy of the enum anchor, so removing the loader filter makes `PATCH 1` ambiguous instead of silently
patching a module Bun never runs.

The `PATCH_TRANSFORMS_VERSION` digest pin is **not** the discriminating test for any of these; each
has a behavioural test that fails on its own.

Two mutations deserve calling out because the obvious test for them does NOT discriminate. The
UTF-8 check needs a **truncated** four-byte sequence (`f0 90 80`), which decodes to one replacement
character that re-encodes to the same three bytes — `ff fe` is caught by the byte-length half and
leaves the other half unpinned. And "exactly one `codesign` call" is not an ordering assertion: the
e2e now captures the file's bytes at the moment `codesign` runs and asserts they already carry the
repointed modules, because signing before the pointer edit leaves the count unchanged and the
signature invalid.

`pnpm typecheck && pnpm test && pnpm build` green (103 files, 1946 tests), and green again under
`CLODEX_HOME="$(mktemp -d)"`. Node v24.14.1, `CLAUDE_CODE_ENTRYPOINT=cli`.

## Runtime evidence, on real binaries

All eight published 2.1.243 builds, through the canary's own reproduction harness, against a pristine
copy of each release:

| Platform | Leg | Result |
| --- | --- | --- |
| darwin-arm64 | real `clodex patch` on this Mac | 11 applied, 0 skipped, 0 failed |
| linux-arm64, linux-arm64-musl | real `clodex patch` in a native-arch container | 11 applied, 0 failed; **patched binary starts and reports 2.1.243** |
| darwin-x64, linux-x64, linux-x64-musl, win32-x64, win32-arm64 | bundle probe | every check green, including byte-for-byte readback of ~36 MB across ~1,385 modules, `modules-intact`, and a valid Mach-O signature |

Before this change all eight failed at `PATCH 1`.

Three things I checked rather than assumed, because the whole approach depends on them:

- **Bun runs the module source, not the bytecode beside it.** Overwriting eleven bytes of one
  chunk's payload in place on a real 2.1.243 darwin-arm64 build (`Your prompt` → `YOUR_PROMPT`),
  re-signing, and running `claude --help` printed the edited string.
- **A chunk-level patch survives the whole pipeline into the running binary.** A local patch that
  rewrites a string living in a chunk — read, joined, transformed, split, one repack, repointed,
  re-signed — shows up in `claude --help` output on a real patched 2.1.243.
- **Re-patching does not accumulate.** Patched a real 2.1.243, changed the alias set, patched again:
  11 applied, identical file size, binary still starts.
- **Local patches still work, on a chunk.** A local patch rewriting a string that lives in a chunk —
  read, joined, transformed, split, one repack, repointed, re-signed — shows up in real
  `claude --help` output; a second one injecting a startup `console.error` into the entry prints
  ahead of `--version`.

## Deliberately not done

One reviewer proposed having `readClaudeBundle` join only the entry module when the entry is not a
stub, restoring the exact pre-2.1.242 local-patch contract on older releases. I did not: it means a
"is this a stub" heuristic and two reader modes to keep honest forever, to protect a local patch
that edits by position rather than by matching an anchor — which is trusted, opt-in code whose
author is now told, in README, exactly where a positional edit lands. Documenting the contract is
the smaller and more durable answer.

## What I could not verify

- **No authenticated end-to-end run.** The one thing left unproven is that a routed request with a
  patched alias actually reaches a provider on 2.1.243. Everything reachable without credentials is
  identical between this build and a 2.1.241 patched by `main` — including the pre-existing
  `[claude-code:unrecognized_model]` diagnostic, which a known-good patched 2.1.241 prints for its
  own aliases too, so it is not a regression. The smoke test
  (`clodex claude -- --model <alias> -p …`) still wants a human run.
- **x64 ELF is bytes-only by design** — no leg executes one. The x64 reasoning is checked against
  the linux-arm64 container leg, which does start its patched binary.
- **2.1.242 itself was never patched end to end.** Its blob has the same shape and module count as
  2.1.243 (verified by parsing it), but only 2.1.243 was run through the matrix.
- **A build with two tweakcc-recognizable module names has never been seen**, so the refusal that
  covers it is reasoned and unit-tested, not observed. Same for a module that is not valid UTF-8:
  every module of every binary measured (2.1.208–2.1.232 and all of 2.1.243) is pure ASCII.

## Failure behaviour

Every new refusal happens while the work is still on a temporary candidate; the live binary is only
renamed into place at the very end, and the candidate directory is removed on any error. A binary
whose blob cannot be parsed falls back to tweakcc's single-module read — which is the pre-2.1.242
behaviour, and on a split binary that means `PATCH 1` fails loudly rather than a partial patch being
published. `clodex patch --restore` is untouched.

## Land order

PR #104 touches `patcher.ts`, `patch-transforms.ts`, `built-in-patch-proofs.ts` and the same test
files. This one should land first — 2.1.243 is currently unpatchable for every user — and #104 rebase
on top.

### DIFF
--- a/.claude/docs/claude-code-internals.md
+++ b/.claude/docs/claude-code-internals.md
@@ -28,10 +28,12 @@ skipping any it has already extracted. It uses clodex's own tweakcc dependency f
 `readClaudeBundle` for the JavaScript, which is why it has to run from inside this repo, where
 `node_modules` resolves.
 
-**Since 2.1.242 the file it writes is a JOIN of every JavaScript module**, separated by
-`/*clodex:module-boundary*/` lines — the same document `clodex patch` matches its anchors against.
-Do not go back to tweakcc's `readContent` for this: it returns only the module it recognizes by
-name, which since 2.1.242 is a ~20 KB stub holding none of Claude Code's behaviour.
+**The file it writes is a JOIN of every JavaScript module**, separated by
+``/*clodex:module-boundary`;{}"]*/`` lines — the same document `clodex patch` matches its anchors
+against. That is true of every version, not only the split ones: a pre-2.1.242 binary carries the
+bundle plus five ~2 KB helper modules, so its extract is a six-module join. Since 2.1.242 it is
+roughly 1,380. Do not go back to tweakcc's `readContent` for this: it returns only the module it
+recognizes by name, which since 2.1.242 is a ~20 KB stub holding none of Claude Code's behaviour.
 
 Extract from a **pristine** `.orig` backup, never a patched live binary: a patched claude reports
 its version perfectly well, so the only thing distinguishing them is the patch marker.
--- a/.claude/docs/claude-code-internals.md
+++ b/.claude/docs/claude-code-internals.md
@@ -23,9 +23,15 @@ export REVIEW_BUNDLE_DIR=<that directory>     # the real-bundle harnesses read t
 ```
 
 That script walks every pristine `*.orig` backup in `~/.tweakcc` (override with
-`TWEAKCC_CONFIG_DIR`) and writes one ~23 MB `.js` per version, skipping any it has already
-extracted. It uses clodex's own tweakcc dependency — `tryDetectInstallation({ path })` then
-`readContent` — which is why it has to run from inside this repo, where `node_modules` resolves.
+`TWEAKCC_CONFIG_DIR`) and writes one `.js` per version (~23 MB before 2.1.242, ~36 MB after),
+skipping any it has already extracted. It uses clodex's own tweakcc dependency for detection, and
+`readClaudeBundle` for the JavaScript, which is why it has to run from inside this repo, where
+`node_modules` resolves.
+
+**Since 2.1.242 the file it writes is a JOIN of every JavaScript module**, separated by
+`/*clodex:module-boundary*/` lines — the same document `clodex patch` matches its anchors against.
+Do not go back to tweakcc's `readContent` for this: it returns only the module it recognizes by
+name, which since 2.1.242 is a ~20 KB stub holding none of Claude Code's behaviour.
 
 Extract from a **pristine** `.orig` backup, never a patched live binary: a patched claude reports
 its version perfectly well, so the only thing distinguishing them is the patch marker.
@@ -34,15 +40,45 @@ Then **enumerate every branch of the function yourself.** Shipped bugs have come
 gates that happened to be visible, and from grepping one syntactic form of a comparison — grep the
 *value*, then trace every hit.
 
+## The bundle was code-split in 2.1.242
+
+Through 2.1.241 the whole ~28 MB bundle was ONE Bun module. 2.1.242 split it: the entry module is
+now a ~20 KB ESM stub whose body is `import{…}from"/$bunfs/root/chunk-<hash>.js"` plus a `main()`
+call, and the code lives in ~1,374 sibling `chunk-*.js` modules — 1,391 modules in the blob, against
+15 before. Chunk names are content-hashed, so they differ between releases AND between platform
+builds of the same release.
+
+| Version | Modules in the blob | Entry module payload |
+| --- | --- | --- |
+| 2.1.232 | 15 | 26,504,651 bytes — the whole bundle |
+| 2.1.241 | 15 | 28,252,504 bytes — the whole bundle |
+| 2.1.242, 2.1.243 | 1,391 | ~19,950 bytes — an import stub |
+
+(Measured on darwin-arm64. 2.1.242 and 2.1.243 have the same module count and differ only in their
+chunk hashes and version string.)
+
+Consequences worth knowing before you touch any of this:
+
+- **The anchors did not move; the reader did.** Every clodex patch site still occurs exactly once
+  across the bundle — they are just in six different chunks on 2.1.243, none of them the entry.
+- **Bun executes the chunk SOURCE, not the chunk bytecode.** Verified on 2.1.243 darwin-arm64 by
+  overwriting eleven bytes of one chunk's payload in place (`Your prompt` → `YOUR_PROMPT`),
+  re-signing, and running `claude --help`: the edited string is what printed.
+- **A module's payload can be repointed without moving it.** `{ offset, length }` in the module
+  struct is the only thing that says where a module's source is, so several modules can be patched
+  in ONE repack by concatenating their new sources into the buffer tweakcc writes and then pointing
+  each module at its own slice. That is what `src/bun-bundle.ts` does.
+
 ## The entry module's name is not stable (renamed in 2.1.229)
 
-The bundle lives in a Bun data blob as one module among ~15, and tweakcc finds it **by name**:
+The bundle lived in a Bun data blob as one module among ~15 (see the code-split section above for
+what changed in 2.1.242), and tweakcc finds it **by name**:
 `/claude`, `claude`, `/claude.exe`, `claude.exe`, `/src/entrypoints/cli.js`, `src/entrypoints/cli.js`.
 
 | Version | Entry module |
 | --- | --- |
 | 2.1.224, 2.1.226, 2.1.228 | `/$bunfs/root/src/entrypoints/cli.js` |
-| 2.1.229, 2.1.231–2.1.234 | `/$bunfs/root/cli` |
+| 2.1.229, 2.1.231–2.1.234, 2.1.241–2.1.243 | `/$bunfs/root/cli` |
 
 (2.1.228 is the last release clodex's pinned tweakcc 4.3.0 — and clodex's own mirrored copy of its
 name list — can discover, and 2.1.230 was never published for any platform package, so 2.1.229 is
@@ -56,8 +92,9 @@ packaging fault described in `patcher.md` and is not it. tweakcc carried the old
 
 Two things that are easy to assume wrongly about the blob:
 
-- **The entry module also carries ~190 MB of Bun bytecode** (`// @bun @bytecode @bun-cjs`), and
-  repacking preserves it verbatim. It does **not** win over the patched source — a canary injected
+- **Every JavaScript module also carries Bun bytecode** (`// @bun @bytecode`) — ~190 MB on the
+  single-module releases, a few hundred KB per chunk since the split — and repacking preserves it
+  verbatim. It does **not** win over the patched source — a canary injected
   into the JS prints at startup on a repacked 2.1.231 (verified twice on macOS arm64 — once by
   hand and once through clodex's own local-patches feature, both printing the canary on stderr
   ahead of `--version`). Bytecode has been present since at least 2.1.226, so this was never the
--- a/.claude/docs/patcher.md
+++ b/.claude/docs/patcher.md
@@ -87,13 +87,23 @@ the whole-bundle count guards in PATCH 5 and PATCH 10 look at.
   module, because a pre-split blob carries five loader-1 helper stubs beside it — verified on 18
   binaries from 2.1.208 to 2.1.241, all six-module, entry first. For a built-in that is inert
   (measured on a real 2.1.232: no anchor and no count-guard text in any of the five), but a LOCAL
-  patch that appends or prepends by position rather than matching an anchor now lands in a
-  different module than it used to. README says so where a local-patch author will see it.
+  patch that APPENDS by position rather than matching an anchor now lands in a different module
+  than it used to — the end of the joined document is the last helper stub, not the end of the
+  entry. Prepending is unchanged on those releases, because they are all entry-first; on a split
+  build neither end is the entry. README says so where a local-patch author will see it.
 - **A repoint is a write AFTER tweakcc's repack, which signs on its way out.** Restoring the
   entry-module name re-signs and covers the normal path; `resignMachOBinary` covers the binary that
   needed no shim. Skip either and macOS refuses to start the result.
 - Bun runs the module SOURCE, not the bytecode that sits beside it — verified on a real 2.1.243
   build, see `claude-code-internals.md`. Without that, none of this would have any effect.
+  **That guarantee has no automated coverage and cannot get any**: 1,377 of the 1,391 module
+  structs on a real darwin-arm64 2.1.243 carry a non-empty bytecode range (and a matching
+  `moduleInfo` and `bytecodeOriginPath`), while `repointBunModuleContents` rewrites only the
+  `contents` pair — so after a patch every chunk points at new source beside its pre-patch
+  compiled form. The blob fixture writes empty bytecode ranges for every module, so there is
+  nothing stale for a test to trip over. This is the one place where fixture and implementation
+  share an assumption the fixture cannot falsify; re-check it with the probe on a real binary
+  whenever the write path changes.
 
 ## The entry-module shim (`bun-entry-module.ts`)
 
--- a/.claude/docs/patcher.md
+++ b/.claude/docs/patcher.md
@@ -1,20 +1,22 @@
 <!-- Read when changing src/patcher.ts, patch-transforms.ts, patch-backup.ts, local-patches.ts,
-     built-in-patch-proofs.ts, bun-entry-module.ts, or anything about `clodex patch`. -->
+     built-in-patch-proofs.ts, bun-entry-module.ts, bun-bundle.ts, or anything about
+     `clodex patch`. -->
 
 # Patcher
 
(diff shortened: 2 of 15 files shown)


bman654 (commented):
## APPROVE

This restores `clodex patch` on a Claude Code that is currently unpatchable for every user, and the
core mechanism holds up under every proof I could construct against real binaries. Two independent
passes — one reviewer and one lens, on separate copies — agree. I have pushed the review follow-ups
to the branch rather than asking for them back (516e713); everything below reflects the branch as it
now stands.

### Verified on real binaries, not fixtures

Against pristine copies of the canary's real 2.1.243 builds, in a sandboxed
`CLODEX_HOME`/`TWEAKCC_CONFIG_DIR`:

| Check | Result |
| --- | --- |
| Real `clodex patch` on 2.1.243 darwin-arm64 | 11 applied, 0 skipped, 0 failed |
| Every module compared against pristine | 1,376 of 1,382 byte-identical; exactly the 6 expected chunks changed; no table drift, no replacement characters |
| All eight platform builds | every probe check green, including byte-exact readback across ~1,385 modules |
| Binary health | starts, reports 2.1.243, `codesign -v` passes, `--help` runs — so chunk code genuinely executes after the repoint |
| `--restore` | returns the file byte-identical to pristine |
| Re-patching | three consecutive patches produce an identical sha256, no size growth |
| Publication order | every new refusal fires on the private candidate before `renameSync`; a throw mid-repoint stays on the candidate and `finally` removes it |

The root cause reproduces exactly as described: a real 2.1.243 reads as 1,382 JavaScript modules
against 6 on 2.1.232, and the entry is **not** at table index 0 — so addressing repoints by table
index rather than bundle position is load-bearing, not defensive.

The PATCH 4 drift is real and total: the old anchor matches **zero** times on every 2.1.243 build,
and the new one matches exactly once with its span inside the Agent tool schema, on all 24 real
bundles checked (2.1.208–2.1.232 plus all eight 2.1.243 builds).

### The mutation table is honest

Audited by running the mutations rather than reading them, full-file. **No row overstates
discrimination; four understate it.** The decoy experiment is the one worth calling out: planting a
genuine competing PATCH 3 anchor in a second module produces `ambiguous anchor: PATCH 3`, a refusal
— not a silent wrong bind. The fixture is not co-designed to hide that.

One claim in the body was backwards: the UTF-8 note says `ff fe` is caught by the byte-length half
"and leaves the other half unpinned". Measured, it is the reverse — deleting the replacement-character
half turns a test red, deleting the byte-length half leaves all 1,946 green, and `ff fe` trips both.
The byte-length half is in fact redundant by construction (`byteLength` is the length of the very
bytes `source` was decoded from), which the code now says so nobody writes a test claiming to pin it
alone.

### What I pushed

1. **`scripts/extract-cc-bundles.mjs` no longer started at all** — a regression against `main`. It
   statically imports `bun-bundle.ts`, which imports `./bun-entry-module.js` the TypeScript way, so
   `node scripts/extract-cc-bundles.mjs` died with `ERR_MODULE_NOT_FOUND` before parsing arguments.
   Nothing in `typecheck && test && build` could see it: `tsconfig.json` includes only `src`, and no
   test loads the script. It matters more than a normal script break, because bare
   `node scripts/extract-cc-bundles.mjs` is the only invocation given anywhere — including the
   canary's own incident runbook, so the next patch-canary page would have handed its responder a
   broken tool. Fixed with the `registerHooks` pattern `probe-patch-mechanism.mjs` already uses.

2. **PATCH 4 was fail-open on an escaped backtick.** Dropping the trailing `)` was right — that
   constraint is exactly what drifted — but `[^`]*?` stops at the first backtick *byte*, which on a
   description containing `` \` `` is mid-escape. Splicing there turns the escaped backtick into a
   live terminator: a syntax error in a module Bun loads at startup. The old anchor merely skipped.
   The wildcard now skips backslash pairs, which keeps the corner fail-closed **without**
   constraining what follows the template. A real 2.1.243 patches to a byte-identical binary either
   way, and the span is identical on all 24 real bundles — this only closes the corner.

3. **Three unpinned guarantees now have behavioural tests**, each verified to go red on its own
   mutation: the module-table order refusal, the separator's punctuation, and the escaped-backtick
   corner. The table-order one is worth a note — it is the only guard that can see a repack which
   preserves every module but reorders the table, and the read-back loop provably cannot substitute
   for it, since it compares each index *after* writing that index. Reachability is currently zero
   because the pinned tweakcc preserves order, so this is calibration; it is also precisely the
   guard a tweakcc bump would silently disarm.

4. **Three wrong or overstated claims corrected.** The separator comment said every built-in anchor
   bounds its wildcards with a character class — PATCH 10's anchor and proof use a negative lookahead
   and *can* cross it (PATCH 10 is safe for its own reason, which is now what the comment says).
   `claude-code-internals.md` scoped the joined extract to "since 2.1.242" and quoted the separator
   without its load-bearing punctuation — the join applies to every version, and I measured a
   pre-split 2.1.232 extracting as a six-module join. That is the same wrong scoping this PR
   corrects in README, left behind in the sibling doc. And `patcher.md`'s "appends or prepends" is
   half wrong: pre-split blobs are all entry-first, so only appending moved.

### Checked and cleared

- **Unchanged modules survive the repack.** The read-back loop only verifies repointed modules, so
  this was the biggest open risk. It holds: every unchanged JavaScript module and every
  non-JavaScript payload is byte-exact against pristine after a real patch.
- **Re-signing order.** The repack signs on its way out, so a repoint after it would invalidate the
  signature. Both paths sign after the pointer edit, and the e2e's capture of the bytes at
  `codesign` time is a genuine ordering assertion, not a count.
- **`restoreEntryModuleName`'s `machO = resign` change** is equivalent — `resignMachOBinary` does its
  own `isMachO` check and no-ops off-Mach-O.
- **The loader filter** selects exactly what Bun executes, verified by enumerating every module and
  loader id on real builds across all three OS families.
- **Old releases don't regress.** Applying the built-ins to the old one-module source and to the new
  six-module join produces an identical patched `cli`, with all five helper stubs byte-unchanged.
- **The 8→9 bump** touches no backup or manifest state; mismatch yields `stale-config` and a repatch.
- **Local patches moving text across a boundary** while preserving the count is real, and is already
  documented as an accepted limit for trusted opt-in code.

### Follow-ups, not blockers

- The no-shim re-sign branch (`patcher.ts`, `else if (repointed) resignMachOBinary`) is unreached by
  any test and dead in production today — no real build carries a tweakcc-recognized name before the
  shim. Correct, but pinning it needs a second code-split e2e whose entry is already named
  `/$bunfs/root/claude`; not worth blocking a total-outage fix.
- "Bun runs the source, not the adjacent bytecode" has no automated coverage and cannot get any —
  1,377 of 1,391 real module structs carry a bytecode range and the fixture writes none. Now stated
  in `patcher.md` so it gets re-checked with the probe whenever the write path changes.

### Still unverified

The authenticated end-to-end run. I have it scripted and will run it after merge.

PR #149, opened 2026-08-26
fix(patch): restore patching on Claude Code 2.1.246, where the patch was lost or crashed claude

bman654:
Patching Claude Code 2.1.246 was broken on every platform, in two different ways. On macOS and
Windows `clodex patch` refused and exited 1. On Linux it reported "11 applied" and left you with a
`claude` that crashed the moment you ran it. And underneath both of those, on this release a
patched module would have gone on running its **pre-patch** code even if the crash were fixed — a
patch that reports success and changes nothing. This restores patching on all eight published
builds and makes the patched code actually run.

## Lead with the second failure, because it is the worse one

Had this PR fixed only the crash, `clodex patch` would report 11 sites applied, the binary would
start, print the right version, carry every `ccpatch:` marker — and run its old code. Nothing clodex
owns would have caught that. The blob fixture writes empty bytecode ranges for every module, so no
test in the suite could ever trip on it, and `--version` cannot tell a working patch from a silent
no-op.

`.claude/docs/patcher.md` has said since 2.1.243 that "Bun runs the module SOURCE, not the bytecode
that sits beside it", that the guarantee has no automated coverage and cannot get any, and that it
must be **re-checked on a real binary whenever the write path changes**. This is the first time
that instruction has mattered. It was re-checked, and it had stopped being true.

Three legs on one pristine 2.1.246 darwin-arm64 (sha256 `7b09f01c…`), each re-signed with
`codesign -s - -f` and run as `claude --version`:

| Leg | What was done to it | `'2.1.246'` in the file | `'2.1.XXX'` | modules with bytecode | prints |
| --- | --- | ---: | ---: | ---: | --- |
| **P** | nothing; re-signed only | 1660 | 0 | 1404/1582 | `2.1.246` |
| **A** | all 1,659 occurrences in the JavaScript rewritten to the same-length `2.1.XXX`; bytecode, module info and source hashes untouched | **1** | 1659 | 1404/1582 | **`2.1.246`** |
| **B** | the same edit, plus each touched module's bytecode, module info and source hash cleared — exactly what this PR does | 1 | 1659 | 0/1582 | **`2.1.XXX`** |

Leg A prints a version that **does not exist in its JavaScript**. The single surviving copy of
`2.1.246` anywhere in that file sits at blob offset 104,504,864 — inside the shared bytecode string
table (103,812,984 … 113,691,148), the constant pool of the compiled form. The binary is reading
its version out of bytecode that no longer corresponds to its own source. Leg P rules out the
re-signing being the cause; the census rules out the edit having missed a copy.

Why it changed: Bun 1.4.1 records the source hash JSC keys its code cache on, instead of computing
it from the source that is actually there, so a patched module no longer invalidates its own cache.
JSC still compares the source **length**, which rejects every built-in clodex patch site — running
the real transforms over a real 2.1.246 bundle changes six chunks and all six grow, at every model
count measured — but not a same-length local patch, and not the general guarantee the docs asserted.
(The per-site byte deltas scale with the configured model count, so they are not quoted.)

This was independently reproduced by a second reviewer with a separately written parser, matching
every number, including the 1,409 non-zero entries in the source-hash array being exactly the
JavaScript module count.

## The first failure, and the single root cause

Claude Code 2.1.245 ships Bun 1.4.0; **2.1.246 ships Bun 1.4.1**. Bun 1.4.1's blob carries three
things no per-module range points at, announced by the blob's `flags` word — **15 on 2.1.245, 255
on 2.1.246**:

| Flag bit | What it announces | On a real 2.1.246 darwin-arm64 |
| --- | --- | --- |
| 5 `HAS_SOURCE_HASHES` | `[u32; modules]` after the module table | 6,328 bytes |
| 6 `HAS_BUILTIN_BYTECODE` | `u32 count`, then `count` × `{id, offset, length}` | 4 bytes (count 0) |
| 7 `HAS_BYTECODE_STRING_TABLE` | one `{offset, length}` after that… | 8 bytes |
| | …pointing at the shared bytecode string table, written among the payloads | **9,878,164 bytes** |

The tail arithmetic lands exactly on `byte_count`: 6,328 + 4 + 8 + 1 = 6,341 bytes from the end of
the module table. Verified against Bun's own source (`src/standalone_graph/StandaloneModuleGraph.rs`
— the `Flags` bitflags and `to_bytes`) and byte-for-byte on real 2.1.245 and 2.1.246 binaries.

tweakcc's repack rebuilds the payload region from the six per-module ranges and the exec-argv
string, carrying `flags` across as a scalar. So it published a blob whose flags promised three
structures whose bytes were gone:

- **linux-arm64, linux-arm64-musl** — the only legs that execute the binary they patch. `11 applied,
  0 skipped, 0 failed`, then exit 139: `panic(main thread): Segmentation fault`, `oh no: Bun has
  crashed`.
- **darwin-arm64, darwin-x64** — dropping ~9.9 MB while adding ~7.4 MB made the blob ~2.5 MB
  *smaller*, so the previous blob's trailer survived *above* the new one. clodex locates a blob by
  scanning back from EOF and taking the first candidate that validates, so it parsed the **stale**
  module table, saw the pristine 20,605-byte entry payload where its own buffer should be, and
  refused: `holds 20605 bytes where the 7414833 bytes clodex wrote should be`.
- **win32-x64, win32-arm64** — the write completed and the binary shrank by ~2.49 MB. Nothing here
  executes a PE, so what that binary does on Windows is **not measured**.
- **linux-x64, linux-x64-musl** — byte-level probe passed. Those legs never start the binary, so
  they could not see the crash.

All 11 patch sites matched on all 8 builds. No anchor drifted; `src/patch-transforms.ts` has no
transform change in this PR.

## The fix: stop rebuilding the blob

tweakcc's repack is still what resizes the container's Bun section — the part that needs node-lief
and knows Mach-O from ELF from PE, and calling it once per changed module is not an option because
each ELF repack relocates the whole ~290 MB blob. But what it rebuilds is now discarded. It is
handed a **placeholder** sized so the section it produces has room, and then:

- `planBundleWrite` appends each patched module's source past the end of the pristine blob, 8-byte
  aligned and NUL-terminated, and repoints that module's struct in place. **Nothing already in the
  blob moves**, so the tail structures, the 128-byte bytecode alignment JSC requires, and whatever
  Bun adds next survive by construction rather than by enumeration.
- A patched module's cached bytecode and module info are cleared and its source hash zeroed — the
  legs above are why.
- `SOURCE_TEXT_CONTIGUOUS` is cleared, because the appended sources make it false. Bun reads it in
  exactly one place, a `madvise(MADV_DONTNEED)` hint.
- The publish **refuses if the repack put the blob back at a different offset modulo 128**. Every
  unpatched module keeps its cached bytecode where it was, and JSC decodes that in place at a
  128-byte boundary; Bun writes each range at `120 mod 128` from the blob's start, which only lands
  correctly because the section base is page-aligned. Measured on six real 2.1.246 builds:
  `blobAt % 128 == 8` and every bytecode range at `120 mod 128`, on all of them. Two independent
  reviewers flagged that nothing else in the verification could see a repack that broke it.

### Why not something smaller

Keeping tweakcc's rebuild means clearing the flags and every module's bytecode — disabling Bun's
compiled-code cache for all ~1,400 modules, slowing every launch for every user — and it would
still be wrong the next time Bun adds a structure. Publish-verbatim-and-append is the smaller
long-term risk precisely because it does not enumerate what it preserves. A concrete instance:
Bun's `main` already adds a fourth tail record, `HAS_STARTUP_MODULE_COUNT`, and this write survives
it for free.

### What it costs

The pristine copy of every patched module stays in the blob as dead space. **On Mach-O and PE**,
where the repack assigns the section in place, that is the whole cost: **+7.5 MB on 2.1.246**
(230,824,016 → 238,390,496 on darwin-arm64), and +28 MB on a pre-split release where the one module
IS the bundle. **On ELF the published binary is ~1.75x pristine** (247,389,632 → 434,470,336 on
linux-arm64 2.1.246) — but that is tweakcc relocating the whole blob to the end of the file and
stranding the original, which predates this change and predates the entry-module shim; what this
change adds on top is the same appended sources. The candidate is renamed into place on every
format, so on Linux that ratio is what the user ends up with, and always has been.

None of it compounds — every patch is built from the pristine backup, so repeated patches settle at
the same size. Peak memory during a patch rises by one copy of the blob's data region (~164 MB on
2.1.246, ~290 MB pre-split); `patcher.md` records it beside the existing ELF figures.

`PATCH_TRANSFORMS_VERSION` goes 9 → 10. Note what that does **not** do, because it would be easy to
claim otherwise: it cannot rescue an install whose patched binary no longer starts. `clodex patch`
resolves the version by *executing* the binary, so a crashing one fails at `version-unknown` before
the staleness check is reached — `--restore` is dispatched earlier and reads the version from the
manifest instead, so those users need `clodex patch --restore` and then a fresh patch. Measured,
that is the two arm64 Linux builds; macOS refused rather than publishing, so it left nothing broken,
and what a 2.8.1-patched Windows binary does is unmeasured. The bump is for every install that still
starts: a patch written the old way carries stale bytecode beside patched source.

## Tests, and the mutations that prove they discriminate

`tests/bun-bundle.test.ts` goes from 19 to 30 tests. The fixture now models Bun 1.4.1's tail
structures, per-module bytecode and `bytecodeOriginPath`, and carries `flags` and `structBytes`
through its simulated repack — `flags` in particular, because carrying it while dropping the
structures is what makes the loss lethal rather than merely wasteful. `ENTRY_ID` moved from 0 to 2,
because at 0 "publish `entryPointId`" and "publish nothing" are the same bytes.

The headline test asserts the red state explicitly: after the stand-in repack it checks the flags
still promise all three structures and that the shared string table is **gone**, then publishes and
checks it is back at the same offset with the same bytes.

An adversarial review ran 50 mutations against the first version of this change and **16 survived**.
All are now caught:

| Mutation | Now caught by |
| --- | --- |
| skip the append alignment padding / `APPEND_ALIGNMENT = 1` | `appends 8-byte aligned, NUL-terminated payloads` |
| append without the NUL terminator | same |
| `repackedBlobBytes` returns 0, or 2×, or slack set to 0 | `sizes the placeholder from the repack it is about to get` |
| drop the module-table names/order check | `refuses to publish when the repack reordered the module table` |
| drop the published `byteCount`/`entryPointId`/`flags`/names read-back | `refuses when the published blob does not read back as the one that was planned` |
| drop the "every patched module came back as JavaScript" count | `refuses when a module it planned to move is not one Bun executes` |
| don't publish `entryPointId` / don't publish the exec-argv pointer | `publishes the offsets struct the plan built` |
| drop the `writableIndex` / `patchedSources.length` guards | two new plan-time tests |
| repoint, or clear bytecode/module info for, the module **next to** the patched one | `leaves every module it did not patch exactly as it found it` |
| disable the modulo-128 refusal | `refuses to publish when the repack moved the blob off its 128-byte residue` |
| loosen the truncation refusal by 4 KiB | boundary test now inflates by exactly slack + 1 |
| disable the overlapping-ranges refusal | `refuses to size a placeholder when the module table double-counts` |

The one surviving mutation is deleting the trailer write, and it survives because it is a **provable
no-op** — the published blob is the same length as the repacked one, so it writes a trailer over an
identical trailer. It is kept so the published blob is written here in full rather than partly
inherited from the repack, and the comment says so.

The reorder test previously passed on the *wrong guard*: rotating the table moved the placeholder
off `writableIndex`, so the placeholder-length check fired first and the names check was never
reached. It now swaps two modules that are not the one tweakcc wrote, and asserts the exact message.

`scripts/probe-patch-mechanism.mjs` gains **`blob-sized-as-planned`**, which is the only place
`repackedBlobBytes` — clodex's arithmetic mirror of code it does not own — meets a real repack on a
real module table. Drift is 0 on every build.

## Runtime evidence

Environment: macOS 26.6.2 arm64, Node v24.14.1, pnpm 10.34.5. Every run under a throwaway `HOME`,
`CLODEX_HOME`, `TWEAKCC_CONFIG_DIR` and `TWEAKCC_CC_INSTALLATION_PATH`; no `ANTHROPIC_*`, `CLODEX_*`
or proxy variables set in the shell.

**All eight 2.1.246 builds, on pristine copies:**

| Build | Leg | Result |
| --- | --- | --- |
| darwin-arm64 | real `clodex patch` on this Mac | 11 applied; patched binary prints `2.1.246` |
| linux-arm64 | real `clodex patch` in `node:24-bookworm` (linux/arm64) | 11 applied; prints `2.1.246`, exit 0 — **was SIGSEGV** |
| linux-arm64-musl | real `clodex patch` in `node:24-alpine` (linux/arm64) | 11 applied; prints `2.1.246`, exit 0 — **was SIGSEGV** |
| darwin-x64 | probe | PASS, 18 checks |
| linux-x64 / linux-x64-musl | probe | PASS, 17 checks each |
| win32-x64 / win32-arm64 | probe | PASS, 16 checks each |

**The discriminating control.** `--version` proves nothing — a pristine binary prints it
identically. The patched and pristine 2.1.246 binaries were run side by side against a local mock
Anthropic endpoint and their request bodies diffed:

```
patched:   "enum": ["sonnet","opus","haiku","fable","luna","sol","terra"]   (+ the clodex description)
pristine:  "enum": ["sonnet","opus","haiku","fable"]
```

The method, in case a future reviewer needs it: point both binaries at a stand-in Anthropic endpoint
that logs the request body and answers `/v1/messages` with a one-token SSE reply, run
`claude --model <alias> -p hi` against each, and diff the `Agent` tool's `model` schema. `--version`
and `--help` are byte-identical between the two, and the model id on the wire is passed through
unvalidated by both, so neither discriminates.

That is `PATCH 1` and `PATCH 4`, both in code-split chunks, executing at runtime out of the
published binary — which also proves clearing those chunks' bytecode did not break their loading.
**It covers those two sites at runtime; the other nine are covered statically** (marker presence,
and every patched module read back through the same parser Bun's loader agrees with). On this
release that control is the only evidence we have that a patched binary runs patched code at all.

**Published blob invariants, on the real patched 2.1.246:** exactly one trailer in the file; `flags`
255 → 239 (only bit 4 cleared); the string-table pointer still `{103812984, 9878164}` with those
bytes unchanged; every surviving bytecode range still `120 mod 128`; 1,582 modules in the same
order; entry module still `/$bunfs/root/cli`.

**Older releases** — the write changed for every version, not just 2.1.246. Probes pass on 2.1.245,
2.1.241 and 2.1.232; real patch-and-run passes on darwin-arm64 2.1.245 and 2.1.241; container
patch-and-run passes on linux-arm64 2.1.245 and 2.1.241. 2.1.241 and 2.1.232 are pre-split, so they
exercise the single-module path where the appended source is the whole 28 MB bundle.

**Upgrade path.** A 2.1.245 binary patched by shipped clodex 2.8.1, then patched by this branch:
it re-patches rather than reading as current, the result runs, and a third run settles to
`already patched with the current model config`.

**Idempotency.** 2.1.246 patched three times: 230,824,016 → 238,341,248, then `already patched`
twice, size unchanged. `--restore` then patch again: 238,341,248 again. (That figure is from the
three-favourite canary config; the probe legs quoted above configure more sites and land at
238,390,496.)

**Rescue path.** `clodex patch --restore` returns the binary to bytes byte-identical to the release
(sha256 `7b09f01c…` before and after) and the restored binary runs.

`pnpm typecheck && pnpm test && pnpm build` all pass — 103 files, 1,962 tests — and the suite also
passes under a throwaway `CLODEX_HOME`.

## Deliberately not in this PR

- **No tweakcc bump.** This is against the pinned **4.3.0**. I read 4.3.3's shipped repack: it still
  rebuilds from the six per-module ranges (`r === 52 ? i.push(c, u, d, f, p, m) : i.push(c, u, d, f)`)
  and still copies `flags` verbatim, so it would publish the same broken blob. A bump on top of a
  mechanism rewrite doubles what a reviewer holds, and 4.3.3's #915 moves ELF placement — the legs
  that were already crashing. Tracked separately. If a future tweakcc does break the 128-byte
  residue, the new refusal makes it a loud failure with an untouched install.
- **A separate, currently-unreachable defect, found on the way.** 2.1.246 is the first build to ship
  modules with `Encoding::Utf16` — 76 of them. None is JavaScript; they are the embedded `.md`/`.txt`
  files Bun does not execute, so clodex never reads or writes one. But `readClaudeBundle`'s UTF-8
  round-trip guard would not catch one if it appeared: a UTF-16LE payload of ASCII has the same byte
  length decoded as UTF-8 and contains no U+FFFD, so it passes the guard and would be written back
  as mojibake. Verified in isolation. A three-line refusal on an unrecognised encoding closes it;
  that belongs in its own PR.

## What I could not verify

- **No leg executes an x64 ELF or a PE binary.** linux-x64, linux-x64-musl, win32-x64 and
  win32-arm64 are bytes-only by design; their runtime behaviour is inferred from the linux-arm64
  container legs, which do run. In particular, what the 2.8.1-patched Windows binary actually did on
  Windows is unmeasured — it completed the write and shrank, and the flags/structures mismatch is
  the same one that crashes on Linux, but that is inference.
- **tweakcc's legacy ELF-overlay repack path is unexercised.** Every Claude Code ELF build measured
  carries a `.bun` section, so it is not reachable from a real install. The publish writes exactly
  `repacked.byteCount + 48` bytes from `repacked.blobAt`, which is the repacked blob's own
  self-description, so it is in-bounds by construction there — but that is arithmetic, not a
  measurement, and it is the one path with no guarantee that the blob keeps its 128-byte residue.
- **`repackedBlobBytes` mirrors arithmetic clodex does not own.** A tweakcc whose repack lays the
  blob out differently produces a section that is too small and the patch **refuses** rather than
  publishing a truncated blob. The probe's `blob-sized-as-planned` check catches drift on a real
  binary per format; 64 KiB of slack absorbs drift too small to be worth a release.
- **The single-module fallback still uses tweakcc's rebuild.** When `readClaudeBundle` returns null,
  `patcher.ts` calls `writeContent` directly. That is correct for an npm `cli.js` install, and on a
  native 2.1.242-or-later binary it cannot publish anything — the module tweakcc recognises is the
  ~20 KB stub, so `PATCH 1` fails first and the patch aborts before any write. Documented rather
  than changed.

## Failure and rollback behaviour

Everything happens on a candidate copy in a temp directory; the live binary is only touched by the
final rename. `applyBundleWritePlan` refuses — leaving the install untouched — if the module tweakcc
wrote does not hold exactly the placeholder (which is also what proves the located blob is the one
the repack just wrote, not a stale trailer above it), if the module table's names or order changed,
if the blob came back at a different offset modulo 128, if the section is too small, if the
published blob does not read back as the one that was planned, or if any patched module does not
read back as its patched source. Reverting this PR restores the previous behaviour exactly; binaries
patched by it are undone by `clodex patch --restore` from the same pristine backup as before.


bman654:
# Review — PR #149 · `fix(patch): restore patching on Claude Code 2.1.246, where the patch was lost or crashed claude`

**Verdict: approve, with notes.** This is correct, and it is correct for the reason the description
gives. I checked the dangerous claim first — that the published binary *runs the patched code* rather
than merely looking patched — and it holds on the metal: a real `clodex patch` against a pristine
2.1.246 darwin-arm64 binary produced a binary that starts, verifies its signature, and serves an
Agent-tool model enum of `["sonnet","opus","haiku","fable","luna","sol","terra"]` where the pristine
control serves `["sonnet","opus","haiku","fable"]`. The "publish the pristine blob verbatim" claim is
literally true byte-for-byte: of the 164,222,846 pristine blob bytes, exactly 40 runs differ, all of
them inside the module table or the source-hash array, and the payload region, the tail, and the
9.88 MB shared bytecode string table are byte-identical. Six module structs change; six source
hashes zero; 1,398 of 1,404 cached-bytecode ranges are left exactly where Bun put them, still at
`120 mod 128`. The refusals are real refusals that leave the live install untouched, `--restore`
reproduces the release sha256 exactly, and the tests discriminate — deleting the bytecode clearing
reddens one named test, deleting the repoint reddens nine across two files.

The restructuring is justified. The mechanism it replaces cannot be made correct: tweakcc's rebuild
reaches only the six per-module `{offset,length}` ranges, and neither the source-hash array nor the
string table is reachable from any of them, so no amount of checking on top of a rebuild would have
helped. I confirmed a dependency bump was not an alternative — the pinned 4.3.0 and the current
4.3.3 carry the identical rebuild expression.

The remaining findings are MINOR; two test/documentation concerns are explicitly DEMOTED. None
should hold the release. I checked the GitHub body and comments at head `f9b52ff` (there are no
comments), so none of the notes below alleges that the author skipped evidence reported elsewhere.

**Release decision:** I found no defect here that is worse than today's outcome for a 2.1.246 user
who runs `clodex patch`. The measured darwin-arm64 and Linux arm64 paths now start, and the
darwin-arm64 control executes patched code; every forced failure refused before the candidate
replaced the live install. Today those users get a refusal on macOS, a crashing binary on the
measured Linux builds, or an unmeasured Windows result. PE runtime after the fix remains inference,
as the PR already discloses; no observed result is worse. The
surviving concerns below are documentation accuracy and future-regression coverage, not a shipped
failure in this head.

---

## Concerns I checked and am explicitly dropping

Recording these so nobody re-derives them.

- **Silent no-op** (the failure mode that would be worse than a crash) — disproven at runtime by the
  enum diff above, read out of the running binary over MCP.
- **Over-broad bytecode clearing.** Exactly 6 of 1,582 modules lose their bytecode
  (`withBytecode` 1404 → 1398) and exactly 6 source hashes zero (173 → 179 zeroes). Everything else
  keeps its range *and* its bytes.
- **Flag damage.** `flags` goes 255 → 239 — `SOURCE_TEXT_CONTIGUOUS` cleared, nothing else. An
  unknown high bit is preserved (there is a test for it).
- **The 128-byte refusal firing spuriously.** It does not fire on any format: Mach-O and PE repack
  in place, and the ELF relocation moves the blob by a multiple of 128. Forcing a misalignment
  produces a loud failure and a byte-identical install.
- **Atomicity.** A forced throw at the entry to `applyBundleWritePlan` — after the repack, the most
  dangerous moment in the new path — left the live binary sha, the manifest, and the candidate
  directory count all unchanged.
- **`--restore` fidelity.** Restored file is `7b09f01c…`, byte-identical to the release, and runs.
- **Code signature.** `codesign -v` reports valid on disk and satisfying its designated requirement
  after publish + re-sign.
- **`PATCH_TRANSFORMS_VERSION` 9 → 10 reaching existing installs.** It does: the version is the
  first element the config hash digests, and a manifest carrying the version-9 hash triggers a full
  11-site repatch seeded from the pristine backup, not "nothing to do".
- **On-disk growth** (+7.5 MB Mach-O/PE, ~1.75× on ELF) — real, but disclosed in `patcher.md` and
  inherent to the repack, not new here.
- **The legacy 36-byte module-struct write path has no committed test.** You disclose this yourself
  at `tests/bun-bundle.test.ts:545`, and I confirmed a mutation there leaves the committed suite
  green. A review-only `structBytes: 36` fixture passes on HEAD and catches the mutant. The measured
  18-binary 2.1.208–2.1.241 corpus and the real 2.1.245/2.1.246 binaries use 52-byte structs, so no
  supported release in the evidence reaches this gap. Per the reachability rule it is ungraded
  hardening, not a finding; `tests/bun-blob-fixture.ts` already accepts `structBytes: 36`.

---

## MINOR 1 — the stated reason for clearing the bytecode is not the reason it works today, and the real reason is stronger

`src/bun-bundle.ts` (the `planBundleWrite` doc comment) and `.claude/docs/patcher.md` say that on
2.1.246 the stale bytecode *wins*, and that "an empty bytecode range leaves Bun no choice but to
compile the source that is there."

Under this PR's own append-and-repoint mechanism, that is not what is doing the work. I removed
**all** of the clearing — bytecode, moduleInfo, bytecodeOriginPath and the source hash — rebuilt,
and ran a real patch on a pristine 2.1.246 copy. The published binary kept every one of them
(`withBytecode` 1404, `zeroSourceHashes` 173, i.e. byte-identical to pristine in those fields) and
still ran the patched code:

```
CONTROL-A Agent model enum -> "enum":["sonnet","opus","haiku","fable","luna","sol","terra"]
```

This does not contradict the same-length experiment that settled the question — it localises it.
JSC's cache key carries the source's hash *and its length*; the settled experiment used same-length
edits, so the key matched and the stale bytecode won. Every built-in transform lengthens its module,
so the key misses on its own. I also ran the missing repoint control during cross-check: I moved the
20,605-byte `/$bunfs/root/cli` source to a different blob offset, changed every `2.1.246` in that
module to the same-length `2.1.XXX`, and left its cached bytecode, module info, and source hash
intact. The re-signed binary still printed `2.1.246`. Thus repointing alone does not rescue a
length-neutral patch; the clear remains required for that supported local-patch shape.

**Please do not delete the clearing** — but the current wording is falsifiable in two minutes, and
the next person who runs that experiment will conclude it is dead code. The stronger justification
is one you can also point at: `~/.clodex/local-patches.mjs` (`README.md:202`) lets a user run an
arbitrary transform, and a length-neutral one is entirely ordinary. Without the clear, that user's
patch is a silent no-op. Suggested restatement: *the cached bytecode is keyed on the source's hash
and its length; the built-in transforms all lengthen their module, but a local patch need not, so
clodex clears rather than relying on that.*

## DEMOTED NOTE — the low four `flags` bits are not pinned, but HEAD preserves them

`tests/bun-bundle.test.ts:375-428` builds its 2.1.246 miniature with
`SOURCE_TEXT_CONTIGUOUS | HAS_SOURCE_HASHES | HAS_BUILTIN_BYTECODE | HAS_BYTECODE_STRING_TABLE`
= 240. Real blobs are 15 (≤ 2.1.245) and 255 (2.1.246) — in both cases all four low bits are set.
In Bun those are `DISABLE_DEFAULT_ENV_FILES`, `DISABLE_AUTOLOAD_BUNFIG`, `DISABLE_AUTOLOAD_TSCONFIG`
and `DISABLE_AUTOLOAD_PACKAGE_JSON`.

I changed both production flag expressions (`src/bun-bundle.ts:339` and `:363`) to additionally
apply `& ~0x0f` and ran the **entire** suite:

```
Test Files  103 passed (103)
Tests  1963 passed (1963)
```

Nothing reddens, and `scripts/probe-patch-mechanism.mjs` has no flags assertion at all, so the
real-binary path would not catch it either. The implementation is right and a review-only
`flags: 15` fixture both passes on HEAD and catches the mutant. Bun names these bits for disabling
autoload of default env files, `bunfig.toml`, `tsconfig.json`, and `package.json`, so preserving
them is a useful contract to pin. But the only trigger is a deliberately broken mutant: no shipped
path in this PR loses the bits. Per the reachability rule this is ungraded future-regression
hardening, not a MINOR. One committed fixture with `flags: 15` closes it.

## MINOR 3 — the documented incremental memory estimate understates the measured increase

`patcher.md` describes the write as adding "the blob's data region held in memory across the repack
(~164 MB on 2.1.246)". That sentence is an incremental estimate, so absolute peak RSS alone does
not refute it. The differential ELF probe does: main measured 3,406,315,520 bytes and this PR
3,958,833,152, an increase of about 552 MB rather than 164 MB. A real `clodex patch` of the pristine
2.1.246 darwin-arm64 binary measured this absolute peak:

```
/usr/bin/time -l …  clodex patch: 11 applied, 0 skipped, 0 failed
      2580578304  maximum resident set size      (2.58 GB)
      2484422856  peak memory footprint
```

The ELF probe peaked at roughly 4 GB. The executed measurements establish both the larger delta and
the absolute peaks; they do not establish which allocator owns every byte, so I am not attributing
the difference to V8 or `Buffer`. Worth correcting the incremental estimate and stating the measured
peaks so constrained environments are not planned around 164 MB.

## MINOR 4 — the published-size figure is the probe's number, not a real patch's

`patcher.md` gives "230,824,016 to 238,390,496 on darwin-arm64". The probe does produce exactly
238,390,496; a real `clodex patch` with a three-favorite config produces **238,308,416**, because
the published size depends on the alias and display strings in `~/.clodex/config.json`. Either say
which config produced it or round it — a precise number invites exactly the reproduction that then
looks like a discrepancy.

## MINOR 5 — three new doc lines break the files' own wrap, one by splicing into an unrelated sentence

`.claude/docs/patcher.md:161` (128 chars), `.claude/docs/patcher.md:297` (140), and
`.claude/docs/claude-code-internals.md:101` (114). On `main` neither file has a line over 108, so
these three are new. The 140-char one is the one to fix: a new sentence about
`blob-sized-as-planned` is jammed onto the same line as the pre-existing platform-independence
sentence, and the two now read as one thought. A reflow of the three paragraphs is all it needs.

## DEMOTED NOTE — the 2.1.246 UTF-16 discovery does not falsify the reader comment

I tried to establish the proposed contradiction against `src/bun-bundle.ts:113-118`. The real
2.1.246 census is:

```
2.1.246: {"enc1/loader1":1409,"enc1/loader13":88,"enc0/loader10":5,"enc0/loader5":4,"enc2/loader13":76}
```

All 76 UTF-16 records are loader 13. `readBunJavaScriptModules` filters to loader 1 before the
round-trip guard, so none is among the "modules" that comment describes; the sentence also scopes
its historical measurement explicitly to 2.1.232 and 2.1.243. The new release therefore does not
make the statement false, and there is no finding. Recording the 2.1.246 census beside it would make
the current boundary clearer, but that is optional documentation upkeep.

## MINOR 7 — name the six builds behind the alignment measurement

`.claude/docs/claude-code-internals.md:107` records the 128-alignment invariant as checked "on
2.1.245 and on all six 2.1.246 builds". The PR body separately and accurately reports all eight
published builds in its runtime table; this note does not allege skipped platform validation. The
doc simply does not identify which six supplied this particular alignment measurement. Naming them
costs a clause and saves the next reader the reproduction attempt.

## MINOR 8 — still-true read-path knowledge went out with the obsolete write bullet

The removed `patcher.md` bullet carried two separable halves. The write half is genuinely obsolete.
The read half is not, and this PR does not change the read: that the pre-split shape was verified on
18 binaries from 2.1.208 to 2.1.241, entry module first, and therefore that **a local patch which
appends by position now lands in the last helper stub rather than at the end of the entry module**.
`README.md:210` keeps the user-facing half, which is why this is minor — but the provenance and the
entry-first fact are gone from the place a contributor would look. As a current executed spot check,
the real 2.1.232 blob has six JavaScript modules: `/$bunfs/root/cli` is both entry id 0 and first,
while `computer-use-input.js` is last, so positional append does land in the helper rather than the
entry. I did not independently recreate the deleted 18-binary census and treat that number as
historical provenance being removed, not as a new measurement.

---

## Praise, specifically

- **The tests genuinely discriminate**, which is the thing that was historically missing here — the
  old fixture wrote empty bytecode ranges for every module, so nothing automated could ever trip on
  stale bytecode. The new one carries non-empty bytecode, module info, origin paths, source hashes
  and the unpointed shared string table, and all of it is load-bearing. My own mutations:
  removing the bytecode clear reddens `clears the cached bytecode of a patched module and keeps
  every other module's`; removing the repoint reddens 7 tests in `tests/bun-bundle.test.ts` plus
  both code-split integration cases in `tests/patcher.test.ts`.
- **Disclosing the one mutation that survives** (the trailer write, which is idempotent today)
  rather than quietly leaving it out.
- **The refusals are placed where the information is.** The `wanted < empty` check in particular
  explains why clamping to zero instead would not be caught downstream — that is the kind of
  reasoning that stops a later "simplification".
- **The description is precise about which legs executed and which are bytes-only**, and says so
  again under "What I could not verify". That is the standard this repo asks for, and it made this
  review a matter of re-running your claims rather than reconstructing them.

PR #153, opened 2026-08-26
fix(openai): prevent nested agents from being forced into worktrees

KristijanJovanovski:
Nested Claude Code agents routed through OpenAI could be forced into temporary Git worktrees even when their Agent input did not request isolation. This change preserves optional tool arguments so those agents stay in the caller's checkout unless isolation is explicitly selected.

## Root cause

Clodex passed the original Anthropic tool JSON Schema to the AI SDK without setting the function tool's `strict` flag. OpenAI Responses may normalize a tool with an omitted `strict` value toward strict mode, where every property must be required. For Claude Code's Agent schema, that made optional properties such as `isolation` mandatory and forced the model to choose `worktree` or `remote`.

The official OpenAI Codex serializer explicitly sends `strict: false` for Responses function tools: https://github.com/openai/codex/blob/21ff2e802cd74eee8b25db6b17d9bd4891a42e38/codex-rs/tools/src/responses_api.rs

## Change

- Set `strict: false` only for tools translated through `@ai-sdk/openai`.
- Preserve the original schema and its `required` list unchanged.
- Leave the strict setting unset for other providers.
- Keep the existing tool-input sanitization and OAuth continuation normalization unchanged.

This deliberately avoids converting optional properties into nullable required properties. Such a conversion would alter Claude Code's tool contract and would require recursive handling of nested objects, unions, references, `additionalProperties`, and unsupported schema keywords.

## Verification

- Added a wire-level Responses request test proving the Agent tool is serialized with `strict: false`, while `description` and `prompt` remain required and `isolation` remains optional.
- Added a provider-scope assertion proving OpenAI-compatible tools remain unchanged.
- Mutation check: replacing the scoped `strict: false` assignment with `undefined` makes the full adapter test file fail exactly the new regression test; restoring it passes all 73 adapter tests.
- `corepack pnpm typecheck`
- `corepack pnpm test`: 103 files and 1,964 tests passed.
- `corepack pnpm build`
- `git diff --check`

A runtime reproduction used a fresh temporary Git repository, a clean Claude configuration, endpoint bridge mode, the ChatGPT OAuth `sol` route, and no session persistence. Official v2.8.2 placed the nested Explore agent under `<fixture>/.claude/worktrees/agent-*`; the patched build reported `<fixture>` itself. Both runs exited successfully, and automatic cleanup left the fixture's Git worktree list unchanged.

Environment: Darwin arm64, Node.js 24.16.0, repository-pinned pnpm 10.34.5.

## Boundaries

`strict: false` prevents the Responses backend from making optional fields mandatory. It does not prevent a model from voluntarily requesting isolation when that behavior is appropriate. Native Claude Code background-session isolation is separate from this translation path.

### DIFF
--- a/src/sdk-adapter.ts
+++ b/src/sdk-adapter.ts
@@ -420,12 +420,16 @@ function toolRequiredProps(tools?: SdkCallParams['tools']): Map<string, Readonly
   return map;
 }
 
-export function translateTools(anthropicTools?: AnthropicTool[]): Record<string, ReturnType<typeof tool>> | undefined {
+export function translateTools(anthropicTools?: AnthropicTool[], npm?: string): Record<string, ReturnType<typeof tool>> | undefined {
   if (!anthropicTools?.length) return undefined;
   const tools: Record<string, ReturnType<typeof tool>> = {};
   for (const t of anthropicTools) {
     if (!t.name || !t.input_schema) continue;
-    tools[t.name] = tool({ description: t.description ?? '', inputSchema: jsonSchema(t.input_schema) });
+    tools[t.name] = tool({
+      description: t.description ?? '',
+      inputSchema: jsonSchema(t.input_schema),
+      strict: npm === '@ai-sdk/openai' ? false : undefined,
+    });
   }
   return Object.keys(tools).length ? tools : undefined;
 }
@@ -546,7 +550,7 @@ export function translateRequest(
       ...translateMessages(messages, npm, supportsExplicitOpenAiCaching),
     ],
     allowSystemInMessages: true,
-    tools: translateTools(upstreamTools.length ? upstreamTools : undefined),
+    tools: translateTools(upstreamTools.length ? upstreamTools : undefined, npm),
     toolChoice: compactRequest ? 'none' : translateToolChoice(body.tool_choice),
     maxOutputTokens: options?.openAiOAuth ? undefined : body.max_tokens,
     temperature: body.temperature,
--- a/tests/sdk-adapter.test.ts
+++ b/tests/sdk-adapter.test.ts
@@ -43,10 +43,72 @@ describe('translateTools', () => {
   it('builds client-side tools (no execute) keyed by name', () => {
     const tools = translateTools([
       { name: 'Read', description: 'read a file', input_schema: { type: 'object', properties: { path: { type: 'string' } } } },
-    ]);
+    ], '@ai-sdk/openai-compatible');
     expect(tools && Object.keys(tools)).toEqual(['Read']);
     expect(tools!.Read.execute).toBeUndefined();
+    expect(tools?.Read.strict).toBeUndefined();
   });
+  it('serializes optional properties as optional OpenAI function arguments', async () => {
+    const requestBodies: unknown[] = [];
+    const provider = createOpenAI({
+      apiKey: 'synthetic-test-key',
+      fetch: async (_input, init) => {
+        requestBodies.push(JSON.parse(String(init?.body)));
+        return new Response(JSON.stringify({
+          id: 'resp_synthetic',
+          model: 'gpt-5.6-sol',
+          output: [],
+          usage: {
+            input_tokens: 1,
+            input_tokens_details: { cached_tokens: 0 },
+            output_tokens: 0,
+            output_tokens_details: { reasoning_tokens: 0 },
+          },
+        }), { status: 200, headers: { 'content-type': 'application/json' } });
+      },
+    });
+    const params = translateRequest({
+      model: 'gpt-5.6-sol',
+      messages: [{ role: 'user', content: 'delegate this task' }],
+      tools: [{
+        name: 'Agent',
+        description: 'Launch an agent',
+        input_schema: {
+          type: 'object',
+          properties: {
+            description: { type: 'string' },
+            prompt: { type: 'string' },
+            isolation: { type: 'string', enum: ['worktree', 'remote'] },
+          },
+          required: ['description', 'prompt'],
+        },
+      }],
+    }, '@ai-sdk/openai', { openAiOAuth: true });
+
+    await generateAnthropicResponse(
+      provider.responses('gpt-5.6-sol'), params, 'gpt-5.6-sol');
+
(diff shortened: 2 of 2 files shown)


bman654:
Thanks for the submission.  Claude is taking a look.  We battled what sounds like a similar issue to this before so I'm curious why our fix didn't solve your problem and also if your fix can let us retire the the existing fix we have.

bman654 (approved):
Thanks for this — it's a well-scoped change with an unusually good PR body, and it holds up. I ran a
three-lens panel plus adversarial refuters against it, and independently reproduced the symptom on the
real ChatGPT/Codex OAuth route.

## The premise is now measured, not inferred

The one thing the PR rests on that the body could only argue by analogy to `openai/codex` — that
Responses normalizes an omitted `strict` toward strict mode — is the thing I most wanted evidence for.
It reproduces cleanly. I instrumented the WebSocket send path and the raw tool arguments, and ran
`gpt-5.6-luna` through the real Codex backend on `d5f019a` and on `68de7b8`:

| observation | base `d5f019a` | head `68de7b8` |
| --- | --- | --- |
| `strict` on the OAuth wire | absent on all 9–10 tools | `false` on all 9–10 tools |
| `Agent` supplied `isolation:"worktree"` unbidden | 3/3 runs | 0/3 runs |
| `WebSearch` supplied `allowed_domains`+`blocked_domains` | 10/10 calls | 0/10 calls |
| …of those, a **non-empty** domain list | 5 of 7 distinct calls | 0 of 5 |
| Anthropic `400 … domains are not accessible to our user agent` | observed twice | never |

So the mechanism is real, it is not confined to the `Agent` tool, and the fix reaches the wire on the
OAuth route specifically — I confirmed the WS transport re-parses the SDK-built body rather than
constructing its own `tools`, so `strict: false` survives to `socket.send`.

## Why this matters more than the PR claims

clodex already had a downstream repair for this symptom class: `sanitizeToolInput` in
`src/tool-input-sanitize.ts` strips `null` and **empty** arrays from optional tool params, added
because a model emitting `allowed_domains: []` made Claude Code 400 against Anthropic's server-side
WebSearch API.

That fix never covered the dominant shape. In 5 of 7 base WebSearch calls the model did not emit an
empty list — it invented a plausible one (`["fifa.com","espn.com","bbc.com","reuters.com"]`), which the
sanitizer cannot strip and which is exactly what produced the 400 above. Your change removes the
compulsion that generates both shapes. Worth saying in the body: this doesn't just un-force `isolation`,
it closes a WebSearch failure that has been reported against clodex before.

## Notes — no action needed from you

None of these block the merge, and I'm folding the two test-level ones into a follow-up PR of my own
rather than sending you back around for them. Recording them so the reasoning is on the record:

**The exact-tools-array assertion is coupled to the vendor's serializer.**
`expect(requestBodies).toEqual([expect.objectContaining({tools: [ ...exact object... ]})])` pins five
keys, of which only `strict` is clodex-controlled. `@ai-sdk/openai`'s `prepareFunctionTool` already
emits a second conditional field (`defer_loading`) in that same object, so an SDK bump can redden this
test for a reason unrelated to clodex. Scoping the tool entry to
`expect.objectContaining({ type: 'function', name: 'Agent', strict: false, parameters: {...} })` keeps
the discrimination without the coupling.

**The test name claims a little more than the diff does.** Base and head serialize byte-identically
except for the added `"strict": false` — the `required: ['description','prompt']` and
`isolation`-stays-optional assertions pass unchanged on base too. "serializes optional properties as
optional OpenAI function arguments" describes the status quo; "sends `strict: false` for OpenAI
function tools" names what the fix actually does.

**One thing worth adding to Boundaries if you touch the description again** (purely optional — your
commit summary is already the right shape for our changelog): `strict: false` turns off constrained
decoding for *every* tool on *every* `@ai-sdk/openai` route, API-key as well as OAuth. clodex validates
nothing of its own — `translateTools` builds `jsonSchema(t.input_schema)` with no `validate` fn, and
the AI SDK's `safeValidateTypes` short-circuits to success when `validate == null` — so a call missing
a required property or carrying a bad enum is forwarded to Claude Code verbatim. Claude Code's own
validator is the backstop, so the cost is a wasted turn rather than a crash, and this is the same
posture `openai/codex` takes. It's the most consequential fact about the change and a reader
shouldn't have to derive it.

**`src/openai-adapter.ts` is a second tool-synthesis path this doesn't reach** — it builds
`tool({description, inputSchema})` itself and serves `POST /openai/v1/chat/completions` in endpoint
mode, resolving the same `@ai-sdk/openai` models to `.responses()`, so that surface keeps the
default-strict behavior. Confirmed pre-existing (identical wire result on `d5f019a`) and off the
Claude Code path. Ours to track, not yours.

## Checked and explicitly *not* a problem

- **Should the `npm` guard be wider?** No — I had a lens argue for dropping it and refuted that. The
  field is serialized through to third-party servers on `@ai-sdk/openai-compatible`, and vLLM rejects
  `tools[*].function.strict` with `400 extra_forbidden`. Your guard is co-extensive with the routes
  that can reach a Responses endpoint. Keep it.
- **`@ai-sdk/azure` excluded** — correct; it isn't a dependency and no supported config reaches it.
- **Prompt caching** — `prompt_cache_key` is byte-identical base vs head and the WS prompt-partition
  key excludes `tools`; the only effect is one cold prefix at upgrade.
- **Does this make our sanitizer dead code?** No, and I checked hard because it looked like it might.
  `strict: false` removes compulsion, not volition (your head still emits optional `run_in_background`
  when it wants it), `@ai-sdk/openai-compatible` routes never get the flag, and the sanitizer
  separately defines the shape the OAuth continuation head snapshot must match. Nothing to remove.
- **`strict` is a declared AI SDK field** (`@ai-sdk/provider-utils` `BaseFunctionTool`), forwarded only
  when non-null — not an undeclared property that happens to survive.
- **Mutation discipline** — reproduced both directions: deleting the fix reddens exactly your new test
  in a full-suite run; widening it to all providers reddens exactly the modified one. All five numbers
  in your Verification section check out.

Approving. Nice piece of work — the runtime repro in your description is what made this quick to
confirm, and the restraint in not rewriting the schema to nullable-required was the right call.

PR #159, opened 2026-09-01
fix(patch): restore Claude Code patching on 2.1.257, broken on every platform

bman654:
Claude Code 2.1.257 could not be patched at all. `clodex patch` failed on every published build —
macOS, Linux and Windows, Intel and ARM — so anyone who upgraded lost their custom model aliases,
their per-model context windows and their effort settings until they went back to an older Claude
Code. This restores patching on 2.1.257, verified against a pristine copy of all eight builds.

It was three separate breaks stacked on top of each other; each one had to be fixed before the next
became visible.

## 1. PATCH 7 — the context-window anchor hardcoded minified parameter names

**Symptom.** `clodex patch: required patch failed: PATCH 7: per-model context window`, "anchor not
found", on all eight builds. Every other site reported OK; PATCH 7 is required, so the patch aborted
there and PATCH 8–10 never ran.

**Cause.** The anchor wildcards every identifier in the resolver's body *except* the two parameter
names, which it spelled `(e,t)`:

```
2.1.252  function Lp(e,t){let r=Vw();if(r!==void 0)return r;if(KSn(e,t))return g$;return Gw(e,t)}
2.1.257  function Np(e,n){let r=AL();if(r!==void 0)return r;if(nAn(e,n))return XN;return bL(e,n)}
```

Same function, and the minifier named the second parameter `n` instead of `t`. Parameter names churn
exactly like the rest of the identifiers; the anchor treated them as if they did not.

**Change.** The anchor binds on shape — two parameters, both threaded unchanged into the two calls,
around the same three-statement body — with both names wildcarded and back-referenced. Measured on
the real extracted bundles, it matches exactly once on 2.1.252 and exactly once on 2.1.257, and
binds to the same function tweakcc-independent inspection identifies as the resolver.

The injected lookup is now built around whichever name the build gave the *model* parameter. That
also closes a latent bug on the refresh path: it rewrote the snippet with a hardcoded `e`, so
re-patching a 2.1.257 binary in place would have left a lookup on an identifier not in scope.

## 2. PATCH 10 — the child-env fingerprint required a name upstream stopped spelling

**Symptom.** Only reachable once PATCH 7 applied: `clodex patch: child network environment target
validation failed`, again on all eight builds.

**Cause.** The anchor bound to the right function; validation rejected it. The check required the
matched body to contain all five of `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_SUBSCRIPTION_TYPE`,
`CLAUDE_BG_PTY_AUTH`, `"OTEL_"` and `CLAUDE_CODE_OTEL_DIAG_STDERR`. 2.1.257 replaced a run of
per-name `process.env.CLAUDE_BG_*!==void 0` reads with a set-membership test and stopped spelling
`CLAUDE_BG_PTY_AUTH` anywhere in the builder.

**Change.** Those names were never the identity proof — the whole-bundle uniqueness of the
`)return process.env;let X={` passthrough early-out, the `CLAUDE_CODE_REMOTE` ternary in the head,
and the brace walk that proves the match ends at the function's own closing brace are. Review
measured that directly: with the five names ignored entirely, the full anchor still has exactly one
candidate on both real bundles, so they never choose among candidates and all a high floor can do is
refuse a target that was already correctly identified — which is this outage. They are now a floor
of one: a body that scrubs nothing is not the builder and is refused, but it takes four independent
upstream refactors rather than one to trip. 2.1.252 matches five, 2.1.257 matches four.
`{...process.env` stays required outright, because the rewrite depends on it structurally.

Every refusal now names the check that produced it. Diagnosing the bare "target validation failed"
from a canary report meant extracting a 32 MB bundle by hand to learn which of four conditions fired.

## 3. tweakcc's ELF repack could no longer find Bun's blob pointer

**Symptom.** Only reachable once both patch-site failures were fixed and the patch got as far as the
repack: `Could not find original BUN_COMPILED location in binary (searched for 0x5541000)` on all
four ELF builds. macOS and Windows were unaffected throughout.

**Cause.** A Bun standalone ELF keeps its `.bun` section's virtual address in an 8-byte global.
`repackELFSection` moves the section, so it has to rewrite that global — and it finds it by scanning
the first writable `PT_LOAD` **only at 16384-aligned addresses**. Measured: on 2.1.246 and 2.1.252,
x64 and arm64, glibc and musl, that global sat at `vaddr % 16384 == 0`. On 2.1.257 it does not, on
any of the four: `0x534f758` linux-x64, `0x4d9af08` linux-x64-musl, `0x5310758` linux-arm64,
`0x4c46d10` linux-arm64-musl. The alignment was luck, not a promise. tweakcc 4.3.3 (latest) carries
the identical scan, so bumping the pin does not fix it — worth knowing for #45.

**Change.** New `src/bun-compiled-pointer.ts`, in the same shape as the existing entry-module shim:
plant a copy of the value the scan looks for at an address it does visit, let the repack rewrite that
copy, then move the address it wrote onto the real global and restore the eight bytes the copy
displaced. It is a strict no-op wherever tweakcc already works — it reproduces tweakcc's own scan
first and stands aside if that succeeds — and it identifies the real global rather than guessing it:
the one 8-byte occurrence of `.bun`'s address inside the writable segment and outside `.bun`'s own
payload. Matches inside the payload are coincidences in compressed JavaScript (dozens on some
builds). Anything other than exactly one candidate is refused, and the restore verifies that the
repack rewrote the stand-in, that it agrees with where `.bun` actually ended up, and that both writes
read back.

## Deliberately not done

- **No tweakcc bump.** 4.3.3 has the same scan; the workaround is needed either way. #45 stays
  independent.
- **No rewrite of the ELF repack in clodex.** Doing the section move ourselves would replace a
  narrow, verifiable stand-in with ownership of the whole relocation, on the path that has already
  broken twice on ELF alone.
- **`PATCH_TRANSFORMS_VERSION` 10 → 11**, so installs patched by an older clodex read as stale and
  repatch rather than keeping narrower anchors forever.

## Verification

Everything below ran against a pristine 2.1.257 downloaded per platform from the npm registry, in a
throwaway sandbox with `HOME`, `CLODEX_HOME`, `TWEAKCC_CONFIG_DIR` and
`TWEAKCC_CC_INSTALLATION_PATH` redirected. Node 24.14.1, macOS 15 host, Docker Desktop for the
container legs.

| Platform | Leg | Result |
| --- | --- | --- |
| darwin-arm64 | real `clodex patch` on the host | 11 applied, 0 failed |
| linux-arm64 | real `clodex patch`, native arm64 container | 11 applied; patched binary starts and reports 2.1.257 |
| linux-arm64-musl | real `clodex patch`, native arm64 container | 11 applied; patched binary starts and reports 2.1.257 |
| darwin-x64 | `scripts/probe-patch-mechanism.mjs` | 18 checks pass |
| linux-x64 | probe | 17 checks pass |
| linux-x64-musl | probe | 17 checks pass |
| win32-x64 | probe | 16 checks pass |
| win32-arm64 | probe | 16 checks pass |

**What that does and does not prove.** The two container legs are the only ones that START the
patched binary, and they cover ELF arm64 only. The probe legs prove the anchors matched, the repack
round-tripped and the published bytes read back byte-for-byte carrying all six clodex patch markers —
they never execute the result. **No leg executes an x64 ELF or a Windows binary**; that coverage is
bytes-only by design, and the x64 ELF reasoning is checked against the arm64 container leg, which
exercises the identical code path on the identical executable format.

`pnpm typecheck && pnpm test && pnpm build` green (1999 tests), and green again under an isolated
`CLODEX_HOME="$(mktemp -d)"`.

**Mutation testing.** Every behaviour here was deleted or broken and the full affected test file
re-run. The `PATCH_TRANSFORMS_VERSION` source-digest pin reddens alongside some of these; it is a
tripwire, and in every row below behavioural tests are red beside it.

| Mutation | Behavioural tests red |
| --- | --- |
| PATCH 7 anchor back to the literal `(e,t)` | 3, with the exact production error |
| PATCH 7 snippet hardcodes `e` again | 2 (one a `ReferenceError` from executing the emitted resolver) |
| PATCH 7 anchor stops proving both arguments are threaded | 1 — the lookalike makes it ambiguous |
| Scrubbed-name floor raised to 5 (the old rule) | 7 |
| Floor lowered to 0 | 1 |
| Floor comparison `<=` instead of `<` | 1 — the boundary case |
| Brace-scan failure folded back into a distance | 1 |
| Anchor accepts any env var in the head ternary | 1 |
| `shimBunCompiledPointer` forced to return null | 7 |
| `restoreBunCompiledPointer` made a no-op | 3 |
| Either production call site deleted individually | 1 each |
| `fileOffsetOf` treats a virtual address as a file offset | 4 |
| Chunk sweep overlap `SCAN_CHUNK` instead of `SCAN_CHUNK - 7` | 2 |
| Two-segment refusal removed | 1 |
| `PN_XNUM` rejection removed | 1 |

**Three adversarial review rounds ran before this was pushed**, one per subsystem plus one on test
discrimination, and a fourth over the changes those produced. They found no wrong bind on either
real bundle and no production defect, and they found five things the tests could not see — the last
four rows above, plus the two individual call sites. All are fixed here. Two review notes are
recorded rather than acted on: PATCH 7's anchor is structural, so a future release could in
principle bind a lookalike silently (true of the anchor before this change too, and an ambiguous
match is refused), and the scrubbed-name floor cannot be made to discriminate anything on today's
bytes because the anchor already has exactly one candidate without it.

## Not verified

- No launch-path smoke test: nothing in `launch.ts`, `env.ts`, `claude-wrapper.ts` or the gateway is
  touched, and running one would have meant patching a real install.
- The "no address tweakcc scans is usable" refusal in the new module is covered by a synthetic ELF
  only. No observed build produces that layout.
- Whether Claude Code 2.1.258+ keeps either the resolver shape or the pointer alignment. Neither is
  a contract; the canary is what catches the next one.

**Unrelated, observed while verifying, not caused by this change:** one probe run in 24 against
linux-arm64-musl published a binary carrying no `ccpatch:` markers at all — the run fell to
tweakcc's single-module path, which on a code-split release patches a stub. The probe caught it and
failed. It did not reproduce in 24 further runs and the input binary was byte-stable, so the cause
is unproven; a plausible one is that `readBunModuleNames` treats a short `readSync` as "no module
table". Through `clodex patch` that same event is a loud failure at the first required patch site,
not a corrupt install, but the message would blame an anchor. Worth its own issue and its own
evidence rather than a speculative fix here.

### DIFF
--- a/.claude/docs/patcher.md
+++ b/.claude/docs/patcher.md
@@ -171,6 +171,46 @@ So the blob is published, not rebuilt:
   emits four. Do NOT use `--version` or `--help` (byte-identical between the two) or the model id on
   the wire (passed through unvalidated by both).
 
+## The Bun blob-pointer stand-in (`bun-compiled-pointer.ts`)
+
+ELF only, and a no-op on every release where tweakcc works unaided.
+
+A Bun standalone ELF keeps the virtual address of its `.bun` section in an 8-byte little-endian
+global; Bun reads it at startup to find the blob. `repackELFSection` moves `.bun` to a fresh page
+past the end of the file, so it has to rewrite that global — and it finds it by scanning the first
+writable `PT_LOAD` for eight bytes equal to `.bun`'s current address, **only at addresses that are
+multiples of 16384**. tweakcc 4.3.0 and 4.3.3 both do this, so bumping the pin does not help.
+
+Through Claude Code 2.1.252 the global happened to land on a 16 KiB boundary on every ELF build
+(measured on 2.1.246 and 2.1.252, x64 and arm64, glibc and musl — all at `vaddr % 16384 == 0`).
+2.1.257 moved it off one on all four: `0x534f758` linux-x64, `0x4d9af08` linux-x64-musl,
+`0x5310758` linux-arm64, `0x4c46d10` linux-arm64-musl. The strided scan steps over it and
+`repackELFSection` throws `Could not find original BUN_COMPILED location in binary`. **Alignment is
+not something Bun promises; it was luck, and it ran out.**
+
+So clodex plants a copy of the value the scan is looking for at an address the scan does visit,
+remembers the eight bytes it displaced, and after the repack copies the address tweakcc wrote onto
+the global's real home and puts the displaced bytes back. The published binary differs from an
+unaided repack in exactly one place: the global, holding the value it was always meant to hold.
+
+- **It never runs where tweakcc already works.** The module reproduces tweakcc's own scan first and
+  returns null if that finds anything, so an older release — or a future one that goes back to
+  being aligned — takes the path it always took, byte for byte.
+- **The real global is identified, not guessed:** the ONE 8-byte occurrence of `.bun`'s address
+  inside the writable segment and outside `.bun`'s own payload. Occurrences inside the payload are
+  coincidences in compressed JavaScript (dozens on some builds) and are excluded by range. Anything
+  other than exactly one candidate is refused rather than patched, because publishing a binary whose
+  Bun global still points at the old blob is a `claude` that dies inside Bun before it prints
+  anything.
+- **The restore verifies rather than assumes:** the stand-in must have changed, must agree with
+  where `.bun` actually ended up, and both writes must read back.
+- Mach-O and PE never reach this code — they assign the section in place and rewrite no pointer.
+  That asymmetry is why the failure was ELF-only, and why the probe must be run per format.
+
+`tests/bun-compiled-pointer.test.ts` covers the rules on hand-built ELF64 files. Only
+`scripts/probe-patch-mechanism.mjs` and the canary's container leg cover a real 300 MB build, and
+only the container leg starts the result.
+
 ## The entry-module shim (`bun-entry-module.ts`)
 
 tweakcc finds the module holding the bundle **by name**, and Claude Code 2.1.229 renamed it from
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -197,6 +197,12 @@ ELF build of Claude Code across two releases while macOS stayed green. Before ch
 `scripts/probe-patch-mechanism.mjs` against a build for each format — Mach-O, ELF and PE. It needs
 no Linux or Windows host, because it never executes the binary. See `.claude/docs/patcher.md`.
 
+**A green macOS/Windows run says nothing about ELF.** Claude Code 2.1.257 moved the global Bun
+reads its blob's address from off the 16 KiB boundary tweakcc's ELF repack scans for it, and
+`clodex patch` failed on all four ELF builds while Mach-O and PE stayed green — Mach-O and PE
+assign the section in place and rewrite no pointer. `src/bun-compiled-pointer.ts` works around it;
+see `.claude/docs/patcher.md`.
+
 **Patch anchors are not platform-independent either.** Claude Code 2.1.238 minified one helper's
 parameter differently on its arm64 builds, so `PATCH 5: model picker options` matched five builds
 and missed `linux-arm64`, `linux-arm64-musl` and `win32-arm64`. The probe now applies every patch
--- a/scripts/probe-patch-mechanism.mjs
+++ b/scripts/probe-patch-mechanism.mjs
@@ -66,6 +66,11 @@ import {
   restoreEntryModuleName,
   shimEntryModuleName,
 } from '../src/bun-entry-module.ts';
+// bun-compiled-pointer.ts imports nothing relative either, so a static import is safe here.
+import {
+  restoreBunCompiledPointer,
+  shimBunCompiledPointer,
+} from '../src/bun-compiled-pointer.ts';
 
 // `src/` spells its own imports the TypeScript way — `./model-aliases.js` for a file that is
 // really `./model-aliases.ts`. tsup and vitest both understand that; bare `node` does not, and
@@ -399,12 +404,19 @@ try {
     const writable = writableModuleIndex(scratch);
     if (writable === null) throw new Error('no module of this build carries a name tweakcc can write to');
     const plan = planBundleWrite(scratch, bundle, splitBundleSource(bundle, written), writable);
+    // Mirrors patcher.ts: without this the ELF repack cannot find the global Bun reads its blob's
+    // address from on Claude Code 2.1.257, and the probe reports a repack failure that has nothing
+    // to do with this build's anchors. See src/bun-compiled-pointer.ts.
+    const bunPointerShim = shimBunCompiledPointer(scratch);
     await writeContent(installation, plan.content);
+    if (bunPointerShim) restoreBunCompiledPointer(scratch, bunPointerShim);
     publishedPlan = plan;
     applyBundleWritePlan(scratch, plan);
     publishedBlob = true;
   } else {
+    const bunPointerShim = shimBunCompiledPointer(scratch);
     await writeContent(installation, written);
+    if (bunPointerShim) restoreBunCompiledPointer(scratch, bunPointerShim);
   }
   let restoreError = null;
   if (writeShim) {
--- a/src/bun-compiled-pointer.ts
+++ b/src/bun-compiled-pointer.ts
@@ -1,0 +1,327 @@
+// Keeps tweakcc's ELF repack able to find the pointer Bun uses to locate its embedded blob.
+//
+// Read `.claude/docs/patcher.md` first. This file exists for the same reason `bun-entry-module.ts`
+// does: tweakcc identifies something in the binary by a rule that a Claude Code release quietly
+// stopped satisfying, and the cheapest correct answer is a reversible byte-level stand-in around
+// the repack rather than a fork of tweakcc.
+//
+// THE RULE. A Bun standalone ELF stores the virtual address of its `.bun` section in an 8-byte
+// little-endian global — Bun reads it at startup to find the blob. `repackELFSection` moves `.bun`
+// to a fresh page past the end of the file, so it has to rewrite that global. It finds it by
+// scanning the first writable `PT_LOAD` segment for 8 bytes equal to `.bun`'s CURRENT virtual
+// address — but only at addresses that are multiples of 16384. tweakcc 4.3.0 and 4.3.3 both do
+// this; bumping the pin does not help.
+//
+// WHAT BROKE. Through Claude Code 2.1.252 that global happened to land on a 16 KiB boundary on
+// every ELF build (measured: 2.1.246 and 2.1.252, x64 and arm64, glibc and musl — all at
+// `vaddr % 16384 === 0`). Claude Code 2.1.257 moved it off one: 0x534f758 on linux-x64,
+// 0x4d9af08 on linux-x64-musl, 0x5310758 on linux-arm64, 0x4c46d10 on linux-arm64-musl. The
+// strided scan steps straight over it, `repackELFSection` throws "Could not find original
+// BUN_COMPILED location in binary", and `clodex patch` fails on all four ELF builds. Nothing about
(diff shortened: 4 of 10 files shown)


bman654:
**APPROVE.** Merging. Four MINORs below are follow-ups, not merge blockers — none of them can mis-patch a shipped build.

Reviewed at head `5972237` with a four-lens adversarial panel plus my own verification, all against **pristine 2.1.257 binaries downloaded fresh from the npm registry** rather than the canary's copies.

## The break and the fix reproduce exactly

Extracted all eight 2.1.257 bundles and ran both anchors:

| | old PATCH 7 anchor | new PATCH 7 anchor | PATCH 10 anchor | passthrough sites | scrubbed names spelled |
|---|---|---|---|---|---|
| all 8 builds | **0** | **exactly 1** | 1 | 1 | **4 of 5** |

So both site failures are real and both fixes are exact. Restoring the old five-name floor reddens all eight real builds with the exact production text.

The new PATCH 7 anchor binds `Nf(e,n)`; reading the surrounding bundle confirms it is the real resolver and that parameter 1 is the model (`Xc(e)`, `Ve(e)`, `xy(e)`) and parameter 2 a betas array (`n?.includes(iI.header)`). The **emitted resolver executes**: extracted out of the real patched bundle and run under stubs, 11/11 cases pass — alias, canonical id, uppercase, surrounding whitespace, unconfigured model, `""`, `undefined`, `null`, plus the preserved `wL()` override and `FEn()` branches. No competing candidate exists on any real bundle, established three independent ways. Backward compatibility measured, not assumed: `old=1 new=1 byteIdentical=true` on every 2.1.246 build and 2.1.252.

## Both gaps the PR body declared are now closed

Real execution now covers **all four ELF builds and darwin-arm64**:

| platform | leg | result |
|---|---|---|
| linux-x64 | real `clodex patch`, amd64 container | 11 applied, 0 failed; **patched binary starts, reports 2.1.257** |
| linux-x64-musl | probe-published binary under `alpine:3` amd64 | **starts, reports 2.1.257**, carries all 6 markers |
| darwin-arm64 | real `clodex patch` on the host | 11 applied; **starts**; published bytes carry the injected snippet on `Np(e,n)` |
| linux-arm64 / -musl | your container legs | as reported |

Byte-level on the published linux-x64: the real global holds the new `.bun` address (`0x10000000`), the stand-in slot is restored **byte-identical**, and each address resolves in exactly one `PT_LOAD` after the repack extends the segment. `scripts/probe-patch-mechanism.mjs` PASSes on all eight of my own downloads. Strict no-op confirmed on real 2.1.252 (linux-x64/arm64, darwin-arm64) and on all four Mach-O/PE builds.

Integration: trial-merged into current `main` (past #160 and #161, zero file overlap) — typecheck clean, **2021/2021**, build green.

## Findings

**[MINOR-1] The ELF no-op path stands aside without checking *which* occurrence tweakcc found.** `bun-compiled-pointer.ts:255` runs `tweakccWouldFind` *before* the candidate census at 261, so the file's own stated safety rule — "the ONE 8-byte occurrence … anything other than exactly one candidate is refused" — holds only on the engage path. A lens built a decoy at the first scanned address with the real global 0x758 above it: `shimBunCompiledPointer` returned `null` and wrote nothing, where the engage path would have thrown `2 candidates`. **Unreachable today**, measured: every occurrence in all four 2.1.257 writable segments is unaligned (14168, 12040, 1880, 11536 within the stride), and on 2.1.252 the single aligned hit *is* the real global — which is why standing aside is correct now. The ~5-line fix is to run the sweep first and use it to *validate* the no-op.

**[MINOR-2] Two comments and `patcher.md` assert a failure mode that does not occur.** `bun-compiled-pointer.ts:241` and `:302` and `patcher.md:203` justify the refusal by saying a stale global means a `claude` that "does not start" / "dies inside Bun before it prints anything". Single-variable disproof: I took the correctly-published linux-x64 binary and changed **only** the eight bytes at `0x534f758` back to the pre-repack address `0x5541000` — it still starts and reports 2.1.257. Reproduced on linux-x64-musl by the panel, and a second control with `restoreBunCompiledPointer` neutered entirely also still started. Bun finds its blob another way (the end-of-file trailer scan `bun-bundle.ts` already documents).

This does not weaken the change at all. The shim is still required — without it `repackELFSection` throws and the patch fails outright. And the restore is still required for the other half of its job: the stand-in slot is **live initialised data on every build** — `.data` on glibc (`0x5330000` inside `[0x532e980..0x5344aac)`), `.tdata`, the TLS initialisation image, on musl. That is a *stronger* argument than the one the comments give. Worth correcting because this repo's own rules note a wrong clodex comment has already propagated into a contributor PR.

**[MINOR-3] PATCH 10's new diagnostics misdescribe what they measured.** `scrubbedSeen` counts substring *spellings*, not scrubbing — a builder that deletes all five names through an external table is reported as `body scrubs only 0 of the 5 known child-env names`. And an overrun reads `match ends -49 characters from the function it started in`, with the direction hidden behind a sign. This is the accuracy of the hunk whose whole purpose is diagnostic accuracy. The adjacent "takes four independent upstream refactors, not one" is also unsupported — extracting the remaining literals into one shared table is one ordinary refactor, and is exactly the move 2.1.257 already made for one of the five.

**[MINOR-4] Both write-readback guards in the ELF restore are deletable with the unit file green** (`bun-compiled-pointer.ts:331-340`, 18/18 either way). `bun-compiled-pointer.ts` is not covered by the transform source digest either, so nothing notices at all.

Plus NITs: "dozens on some builds" is really 0, 0, 5, 3; `writableLoad` first-vs-last, `tweakccWouldFind`'s `<=` bound and the stand-in/pointer overlap guard are each unpinned (163/163 green when mutated, all mirror-fidelity or fail-safe today); `readAt`'s short-read contract turns into a raw `RangeError` at line 321; and the PR body's mutation table under-reports several rows in the favourable direction (`shimBunCompiledPointer` → `null` is 15 red, not 7).

## Demotions, stated out loud

- **A MAJOR was raised and refuted with measurement**: "PATCH 7's tests pin the enclosing function but not the matched span". Killed three ways — over all nine real bundles the narrow anchor and the greedy mutant match once each and produce **byte-identical** output, so no behavioural assertion can detect it by construction; the same mutation on `main` also leaves only the digest red, so the PR neither introduces nor widens the gap (it takes PATCH 7 from 134 to 145 tests and adds the site's first behavioural coverage); and with *all* back-references removed candidates is still 1 on every shipped build, with `applyOnce` failing closed on ambiguity.
- Two lenses said MINOR-1's consequence was a binary that will not start. The stale-global experiment above contradicts that. The finding survives; its stated failure mode does not.
- The un-`finally`'d restore in `patcher.ts` and the probe is **not** a leak: `candidatePath` lives in a `mkdtemp` dir that `applyPatch`'s `finally` `rmSync`s on every exit, `renameSync` runs only on the success path after the restore, and the pristine backup is published ~90 lines before the shim runs under a sha256 guard.
- `scanForNeedle`'s 7-byte chunk overlap is differentially identical to a naive whole-segment scan over four real ~130 MB segments, no duplicates — and both stride mutations are behaviourally caught.
- `braceScanFailed`'s sentinel `bound = 0` cannot mask a real violation: the `why` chain tests it first and `bound` has no other consumer.
- Dropping `CLAUDE_CODE_REMOTE` from `requiredBodyLiterals` really is dead code — the anchor's body group already requires it.

## Not verified

No `claude -p` round trip through a provider anywhere in this review; `--version` is the deepest execution reached. Windows PE and darwin-x64 remain bytes-only. Nobody reproduced or root-caused the one-in-24 probe anomaly you report, or checked that deferring it is safe. And Lens 1 (PATCH 7) died mid-run — its harnesses were recovered and re-run (14/14 green) so its *measurements* are in the record above, but PATCH 7 has no dedicated adversarial verdict beyond the refuter's work and my own.

PR #182, opened 2026-09-05
fix(auth): stop stalled OpenAI responses from hanging model refresh and sign-in

### LINKED ISSUE #168: Model-catalog refresh clears its 10s timeout before reading the response body
`src/registry/refresh-models.ts` (~202-227) arms a 10-second `AbortController` around the OpenAI OAuth catalog fetch, but clears the timer in a `.finally()` that runs when the response *headers* arrive — before `response.text()`/`response.json()` at ~220-223.

A server that returns headers promptly and then stalls the body leaves that read unbounded despite the timer. The timeout effectively covers time-to-headers only.

Both catalog tiers are affected (the Codex-specific model endpoint and the `/backend-api/models` fallback). Found during a sweep of upstream request budgets for #166; unrelated to that issue's fix.

The timer should stay armed until the body has been consumed — compare `src/registry/fetch-template-models.ts` and `src/registry/custom-endpoint.ts`, which clear theirs in an outer `finally` after parsing.

### LINKED ISSUE #169: Device-code sign-in can hang indefinitely: no timeout on any of its fetches
`src/oauth/openai.ts` device-code sign-in issues three requests with no `AbortSignal` and no timeout:

- `requestOpenAiDeviceCode` (~47-60)
- the device-token poll and the authorization-code exchange inside `pollOpenAiDeviceCodeToken` (~76-115)

The five-minute deadline (`deadline` ~73, checked by the `while` at ~75) is only evaluated **between** completed calls, so a single stalled request or stalled body read outlives it. `clodex` then sits in sign-in with no error and no way forward short of Ctrl-C.

By contrast the browser-PKCE flow's token exchange goes through `postOAuthRefresh` (`src/oauth/refresh-http.ts:13-58`), which does arm a 30s `AbortController` that stays live through `response.json()`. The device-code path should get the same treatment.

Found during a sweep of upstream request budgets for #166; unrelated to that issue's fix.

bman654:
Refreshing OpenAI's model list and signing in with a device code no longer wait forever when OpenAI starts a reply but does not finish it. Model refresh now falls back instead of hanging, while device-code sign-in retries a timed-out approval check and reports clear OpenAI-specific errors when setup or token exchange stalls.

## User-visible failure

An OpenAI model-list refresh could remain on its spinner indefinitely if either ChatGPT catalog endpoint returned headers and then stalled its body. The same failure affected the automatic refresh after OpenAI OAuth sign-in.

The default OpenAI device-code flow could also hang indefinitely in any of its three requests: requesting the user code, polling for approval, or exchanging the authorization code for tokens. The existing device-code deadline was checked only between completed polls, so it could not interrupt a stalled request or body read.

These paths are reachable through `clodex providers refresh`, the refresh performed after authentication, `clodex providers auth openai`, and first-run OpenAI setup. OpenAI currently omits `expires_in` from the device-code response; the 300-second approval window is clodex's own fallback, not a lifetime stated by OpenAI.

## Change

- Keep each catalog request's existing 10-second abort timer active through success and error body consumption. A failed Codex catalog request proceeds to the general ChatGPT catalog, then to the existing static/cached fallback behavior.
- Reuse one timeout helper that gives each device-code setup, polling, token-exchange, browser-exchange, and token-refresh request its own 30-second timer. Each timer remains active through response-body consumption.
- Clamp only an approval poll to the remaining device-code window. Once a poll returns an authorization code, the approval deadline has served its purpose, so token exchange receives its own full 30-second request budget.
- Retry an individual timed-out approval poll while the approval window remains. Non-timeout network failures still propagate immediately instead of being misreported later as approval expiry.
- Report attributed timeout messages for device-code setup and token exchange.

The change deliberately does not make these timeouts configurable, change browser sign-in behavior, alter stored credentials or registry formats, or introduce a different catalog timeout message from the sibling model-fetch implementations.

## Discriminating tests

The new tests resolve response headers immediately and leave body reads pending until the production abort signal fires.

- Moving catalog timer cleanup back onto the `fetch()` promise made both catalog body tests fail in the full test file. Changing the catalog budget from 10,000ms to 100ms also failed both tests at the new pre-deadline assertion.
- Removing each of the three OAuth abort signals independently failed the corresponding full-file test. Changing the OAuth budget from 30,000ms to 1,000ms failed the pre-deadline assertions.
- Restoring the device-code deadline clamp on token exchange failed the positive test where approval arrives 100ms before synthetic expiry and a healthy 400ms exchange completes afterward.
- Removing timed-out-poll recovery failed both the retry test and the attributed approval-expiry assertion.
- Deleting the non-timeout poll guard changed `TypeError: fetch failed` into `OpenAI device authorization timed out`; the new test failed.
- Collapsing all token-exchange errors into the timeout branch changed `OpenAI token exchange failed (400)` into `OpenAI token exchange timed out`; the new test failed.
- Removing either timeout-message wrapper exposed the bare `The operation was aborted due to timeout` text and failed the exact-message assertion.

Every mutation used a file snapshot and a full test-file run rather than test-name isolation.

## Runtime verification

Verification ran on macOS 26.6.2 arm64 with Node v24.14.1 and pnpm 10.34.5. The automated gate used an isolated `CLODEX_HOME` and `CLAUDE_CODE_ENTRYPOINT=cli`:

- `pnpm typecheck` passed.
- `pnpm test` passed: 107 files and 2,327 tests.
- `pnpm build` passed.

Manual smoke testing used the default proxy mode and covered `--model haiku` for Anthropic passthrough and `--model luna` for the OpenAI translated path.

## Verification boundaries

I did not induce a stalled response body from the live OpenAI service; the stall behavior was exercised with deterministic fetch responses whose bodies reject only when the real production signal aborts. No live credential or persisted provider data was changed by these tests.

The documentation and CLI-help sweep found no timeout claim made stale by this change. The source tree also has no remaining `finally(() => clearTimeout(...))` pattern that would clear one of these timers at response headers.

## Rollback

There is no data migration or persisted-format change. Reverting the two commits restores the previous request behavior; existing cached model lists and OAuth credentials remain readable, but stalled catalog bodies and device-code requests can once again wait indefinitely.

Closes #168
Closes #169

### DIFF
--- a/src/oauth/openai.ts
+++ b/src/oauth/openai.ts
@@ -3,7 +3,11 @@
 import { generatePkce, generateOAuthState, positiveSecondsToMs, sleepMs } from './pkce.js';
 import type { OAuthTokenResponse } from './types.js';
 import { VERSION } from '../constants.js';
-import { postOAuthRefresh } from './refresh-http.js';
+import {
+  OAUTH_REQUEST_TIMEOUT_MS,
+  postOAuthRefresh,
+  withOAuthRequestTimeout,
+} from './refresh-http.js';
 import { startCallbackServer } from './callback-server.js';
 
 const CLIENT_ID = 'app_EMoamEEZ73f0CkXaXp7hrann';
@@ -29,6 +33,10 @@ export interface OpenAiDeviceCodeData {
   expires_in?: number;
 }
 
+function isTimeoutError(error: unknown): boolean {
+  return error instanceof Error && error.name === 'TimeoutError';
+}
+
 export function extractOpenAiAccountId(tokens: OAuthTokenResponse): string | undefined {
   const token = tokens.id_token ?? tokens.access_token;
   if (!token) return undefined;
@@ -45,18 +53,30 @@ export function extractOpenAiAccountId(tokens: OAuthTokenResponse): string | und
 }
 
 export async function requestOpenAiDeviceCode(): Promise<OpenAiDeviceCodeData> {
-  const response = await fetch(`${ISSUER}/api/accounts/deviceauth/usercode`, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-      'User-Agent': `clodex/${VERSION}`,
-    },
-    body: JSON.stringify({ client_id: CLIENT_ID }),
-  });
-  if (!response.ok) {
-    throw new Error('Failed to initiate OpenAI device authorization');
+  try {
+    return await withOAuthRequestTimeout(async signal => {
+      const response = await fetch(`${ISSUER}/api/accounts/deviceauth/usercode`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'User-Agent': `clodex/${VERSION}`,
+        },
+        body: JSON.stringify({ client_id: CLIENT_ID }),
+        signal,
+      });
+      if (!response.ok) {
+        throw new Error('Failed to initiate OpenAI device authorization');
+      }
+      return await response.json() as OpenAiDeviceCodeData;
+    });
+  } catch (error) {
+    if (isTimeoutError(error)) {
+      throw new Error('OpenAI device authorization timed out while requesting a sign-in code', {
+        cause: error,
+      });
+    }
+    throw error;
   }
-  return response.json() as Promise<OpenAiDeviceCodeData>;
 }
 
 export function openAiDeviceCodeUrl(): string {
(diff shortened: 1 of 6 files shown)


bman654:
### CodeQL check: 3 pre-existing alerts, re-attributed — not introduced here

`CodeQL` reports "3 new alerts including 3 high severity security vulnerabilities"
(`js/insufficient-password-hash`) at `src/env.ts:1955`, `:2074` and `:2106`.

**This PR does not touch `src/env.ts`** — `git diff origin/main..HEAD -- src/env.ts` is empty.
All three alerts are already open on `main` today (along with ~7 more of the same rule in the
same file). They are attributed to this PR only because refactoring `postOAuthRefresh` changed
the shape of the dataflow path CodeQL walks to reach them, not because the PR added anything.

They are also false positives on the merits. All three sites are integrity checksums, not
password storage:

```ts
const shortDigest = createHash('sha256').update(key).digest('hex');
```

The credential is stored in the OS keychain; the sha256 exists only to detect whether a chunked
write round-tripped intact. `js/insufficient-password-hash` wants bcrypt/scrypt/argon2, which
applies to password *verification* material — a slow KDF here would defeat the purpose (a
corruption check) without adding protection, since the digest guards nothing an attacker holding
the keychain entry does not already have in plaintext.

No action taken in this PR. Flagging so a reviewer does not have to re-derive it, and because
cleaning up the whole rule in `src/env.ts` belongs in its own change.

bman654:
### Added: end-of-life abort contract

Per maintainer review, both AbortControllers this PR owns are now aborted unconditionally when the
operation settles — success or throw — in the same `finally` that already clears the timer:

- `src/oauth/refresh-http.ts` `withOAuthRequestTimeout` → `new Error('OAuth request completed')`
- `src/registry/refresh-models.ts` `fetchJsonWithAuth` → `new Error('OpenAI catalog request completed')`

**This is defence in depth, not a fix for an observed leak.** Two independent audits measured zero
retention on these paths after forced GC, on Node v24.14.1 and on the Node 22 engines floor. Node's
only module-level strong collection, `gcPersistentSignals`, admits a signal only when
`(this[kTimeout] || (this[kComposite] && this[kSourceSignals]?.size)) && ...` — a plain
`new AbortController()` signal has neither flag and can never be registered. What the unconditional
abort buys is deterministic firing of consumer abort listeners at end-of-life.

Ordering is load-bearing and pinned: the abort must land only after the response body has been fully
consumed, or it would truncate the very body-read this PR exists to protect. Mutations moving the
abort earlier redden the pre-existing stalled-body tests (3 failed / 9 passed, and 3 failed / 14
passed respectively); mutations deleting it redden the new completion tests.

Behaviour is unchanged: success values byte-identical, `OpenAI token exchange failed (400)` preserved
with the same constructor, and stalled bodies still reject with their original timeout reasons
(`TimeoutError` / `AbortError`), not the completion reason.

Gate: typecheck clean, 107 files / 2329 tests, build success. Smoke on default proxy mode:
`--model haiku` → HAIKU-OK, `--model luna` → LUNA-OK, catalog rendered both times.

PR #183, opened 2026-09-05
fix(server): cancel provider calls after clients disconnect to reduce wasted work

### LINKED ISSUE #170: clodex server keeps upstream requests running after the client disconnects
`clodex server` never creates a client abort controller, so a downstream disconnect cannot cancel the upstream request it started. `src/proxy.ts` does this correctly — it builds `clientAbort` (~396-403) and threads it into the relay (~558-566), the SDK observer (~701-713) and the generate options (~728-742).

The server-mode equivalents all omit it:

- `handleAnthropicMessages` passthrough → `relayAnthropicMessages` (`src/server/router.ts` ~356-380), no `signal`
- `count_tokens` (~630-640), no `signal`
- direct OpenAI passthrough (~695-709), no `signal`
- the translated SDK calls (~463-471, ~481-489) omit the observer's `abortSignal`

Impact is bounded rather than a hang — the translated paths still have their own idle/total timers, so an abandoned request dies on a timer instead of immediately. But a client that disconnects (Ctrl-C, agent killed, browser closed) leaves an upstream generation burning tokens and a provider connection until that timer fires, and the raw passthrough paths have no timer at all.

Found during a sweep of upstream request budgets for #166. Deliberately kept out of that fix: it is about abort propagation, not request budgets.

bman654:
If you run `clodex server` and the program talking to it goes away — you press Ctrl-C, an agent is
killed, a browser tab closes — clodex used to keep the request it had already sent to the model
provider running anyway. The answer nobody would ever read kept being generated. Now clodex asks the
provider to stop as soon as the client disconnects, whether or not the answer had started arriving.
Cancellation is cooperative: clodex stops relaying and requests cancellation, but it cannot force a
provider to stop, and nothing here proves a provider stops generating or billing — that is not
observable from our side.

Closes #170.

## The user-visible failure

`clodex server` never created a client abort controller, so a downstream disconnect could not cancel
the upstream request it had started. An abandoned request stayed in flight:

- **Translated (SDK) routes** eventually died, but only when their idle or total timer fired —
  120s and 10m by default.
- **Raw passthrough routes** had no timer at all and ran until the provider finished on its own.

`src/proxy.ts` already did this correctly. Only the server-mode equivalents in `src/server/router.ts`
omitted it.

## Root cause and production reachability

`handleAnthropicMessages`, `handleAnthropicCountTokens` and `handleOpenAIChatCompletions` each start
upstream work with no cancellation signal attached. Seven call sites, all in `src/server/router.ts`.
I swept `src/server/` for `relayAnthropicMessages`, `streamAnthropicResponse`,
`generateAnthropicResponse`, `streamOpenAiResponse`, `generateOpenAiResponse` and raw `fetch(`; every
hit is in `router.ts`. `index.ts`, `models.ts`, `auth.ts`, `prompts.ts`, `provider-select.ts`,
`catalog-filter.ts` and `vendor-mask.ts` initiate no upstream request.

**Reachability is narrower than the issue implies, and this matters for how to read the rest of
this PR.** Only standalone `clodex server` in **endpoint** mode executes `src/server/router.ts`.
The default proxy mode does not go through this file at all.

| # | Site | Reachable? |
| --- | --- | --- |
| 1 | Anthropic passthrough → `relayAnthropicMessages` | **hardening only** |
| 2 | translated Anthropic stream → `streamAnthropicResponse` | reachable |
| 3 | translated Anthropic non-stream → `generateAnthropicResponse` | reachable |
| 4 | count_tokens → `relayAnthropicMessages` | **hardening only** |
| 5 | direct OpenAI passthrough → `relayAnthropicMessages` | reachable |
| 6 | translated OpenAI stream → `streamOpenAiResponse` | reachable |
| 7 | translated OpenAI non-stream → `generateOpenAiResponse` | reachable |

Sites 1 and 4 need a catalog entry with `modelFormat: 'anthropic'` and a `baseUrl`, which no stock
provider (`openai`, `openai-oauth`, `opencode-go`) produces. They are fixed for consistency, not
because a supported configuration reaches them — calling them a user-facing fix would overstate it.
Sites 6 and 7 were **not** in the issue's list; they are the same defect on the
`/openai/v1/chat/completions` translated route, found by sweeping for siblings.

## The change

`watchClientDisconnect(res)` in `src/http-utils.ts` — the mechanism `src/proxy.ts` already used,
extracted so both share it rather than running parallel copies. It aborts on a response `close` that
did not finish writing, so it covers a disconnect **mid-stream** as well as one before any output,
and never a normal completion.

All seven sites now pass the signal to their upstream call: `signal` on the three raw relays,
`abortSignal` on the four SDK calls. `streamOpenAiResponse` and `generateOpenAiResponse` gained an
optional `abortSignal` (they had none), forwarded into their existing timeout controller with
`forwardAbortSignal` — the same helper `sdk-adapter.ts` already uses, which is now exported. That is
the only change to `sdk-adapter.ts`: one word.

Five `catch` sites became abort-aware, so a cancelled request writes no error response — the client
that would read it is the one that left. This mirrors proxy mode. Non-abort failures are rethrown
unchanged and still reach a connected client.

**The extraction has one deviation from the original:** it drops the `IncomingMessage` `'aborted'`
listener. I probed all three disconnect shapes and an unfinished `res` `close` fires in every one,
including the mid-upload case that `'aborted'` covers; a cold review reproduced this on Node
v24.14.1, v22.11.0 **and** v22.14.0 (the published `engines` floor), with 4–61 microseconds between
the two events. The listener was dead for this helper's callers, so keeping it meant an unpinnable
branch and a cleanup path that itself changed behaviour. `req` was dropped from the signature rather
than left as an unused parameter.

**Deliberately left out:** OAuth credential refresh (`resolveProviderCredential`) — a credential
fetch, not a billable generation, and cancelling mid-rotation could leave a half-rotated token;
`createLanguageModel` — provider construction, cached, no upstream request of its own; and the local
catalog/health routes, which make no upstream call. `src/proxy.ts`'s behaviour is otherwise
untouched, and `src/oauth/responses-websocket.ts` is not modified.

## Mid-stream cancellation was a test gap, not a behaviour gap

An earlier revision pinned only the pre-header window. A mutation that disabled cancellation *after*
headers were sent passed the entire suite — meaning the most realistic user action of all (stream a
few tokens, then Ctrl-C) was unpinned.

I checked by execution whether the behaviour was also missing. It was not: `writableFinished` only
becomes true after `res.end()`, so a streaming response is still "unfinished" and the abort fires
normally. Confirmed for both the raw relay (undici's signal cancels an already-resolved response's
body stream, not just the pre-header phase) and both translated SDK streams. The behaviour was
correct; only the tests were blind. It is now pinned **independently of** the pre-header window —
mutating one leaves the other green.

## Discriminating tests and mutations

+22 tests (2318 → 2340). Every run below is a **full-file** run, never `-t`; every mutation was
applied after committing and restored from a `cp` snapshot.

**Cancellation fires, per site** (`tests/server-router.test.ts`, real gateway, real client socket,
upstream that answers nothing so an unfinished close is proof the connection was torn down):

| Mutation | Red |
| --- | --- |
| drop `signal` from the Anthropic passthrough relay | `cancels the Anthropic passthrough upstream request` + its mid-stream sibling |
| drop `signal` from count_tokens | `cancels the count_tokens upstream request` |
| drop `signal` from the OpenAI passthrough relay | `cancels the direct OpenAI chat-completions upstream request` |
| drop the observer's `abortSignal` | `cancels the translated Anthropic streaming SDK call` + its mid-stream sibling |
| drop `abortSignal` from `generateAnthropicResponse` | `cancels the translated Anthropic non-streaming SDK call` |
| drop the options arg from `streamOpenAiResponse` | `cancels the translated OpenAI streaming SDK call` + its mid-stream sibling |
| drop `abortSignal` from `generateOpenAiResponse` | `cancels the translated OpenAI non-streaming SDK call` |

Each mutation reddens only tests anchored at the site it disables, and no test at any other site.

**Mid-stream window.** Mutating the guard to `if (!res.headersSent && !res.writableFinished)` —
cancellation disabled once output has started — reddens exactly four tests and nothing else:
`aborts when the client disconnects mid-stream, after headers are sent` (`http-utils`) and the three
`… after output has reached the client` router tests (5001–5005 ms each). The pre-header tests stay
green.

**Normal completion is never cancelled.** Two negatives: a passthrough SSE stream with a 50 ms gap
between chunks arrives whole; and a translated call's captured signal is asserted `aborted === false`
*after* the gateway's own `res` has closed (hooked via a second `server.on('request')` listener, not
a sleep). Removing the `writableFinished` guard so it aborts on every close reddens both.

**The abort guards are narrow.** Widening each of the three raw-relay guards to `return;`
unconditionally reddens exactly its own new negative — `still reports an unreachable upstream on the
{Anthropic messages, count_tokens, OpenAI chat completions} route`, each asserting a connected client
still gets a 500. The two translated `break` guards needed no new tests: widening them reddens 7 and
4 **pre-existing** tests respectively (prompt-too-long shape, retry-after clamping, both OAuth-401
retry paths, no-retry-after-output, non-streaming force).

**Feature deletion.** Removing the watcher entirely reddens 10 router tests, 2 `http-utils` tests,
and the **pre-existing** `tests/proxy.test.ts > translated request cancellation` — the extraction is
load-bearing for proxy mode too, which is what makes sharing safer than duplicating.

**Adapter plumbing.** `tests/openai-adapter.test.ts`: neutering the signal forwarding reddens all
three cancellation cases; deleting `stopForwardingAbort()` from `close()` reddens the listener test
(`getEventListeners(signal, 'abort')` length 1 instead of 0). The router tests mock the adapter and
cannot see this layer, so it is pinned separately.

## Runtime evidence

**Environment for all of it:** macOS 26.6.2 (darwin 25.6.0), Node **v24.14.1** (`.nvmrc` pin), pnpm
10.34.5, `CLODEX_HOME=$(mktemp -d)`, `CLAUDE_CODE_ENTRYPOINT=cli`, `HTTPS_PROXY` and
`ANTHROPIC_BASE_URL` both unset (no ambient bridge).

**Endpoint mode — the only configuration that reaches this code.** Isolated `CLODEX_HOME` seeded
from the real config, `node dist/cli.js server --endpoint --quick --port 17699 --no-discovery`,
against ChatGPT-OAuth GPT-5.6 Luna:

- Non-streaming translated request (site 3) → `{"type":"text","text":"ENDPOINT-OK"}`.
- **Mid-stream disconnect against the real provider:** streaming request killed at 4 s, after
  `message_start` and `content_block_start` had already reached the client. Gateway stayed healthy
  (`{"ok":true}`); the follow-up request returned `AFTER-CANCEL-OK`.
- **Zero** `error` / `unhandled` / `ECONNRESET` lines in the gateway log for the whole session.
- `~/.clodex/server-runtime.json` mtime unchanged and the isolated home never created one —
  `--no-discovery` kept shared discovery untouched, so concurrent local work was unaffected.

**Default proxy mode** (`--model haiku` Anthropic passthrough → `HAIKU-OK`; `--model luna` translated
→ `MODEL-OK`): this is **regression coverage for the extracted helper only**. Proxy mode does not
execute `src/server/router.ts`, so it is not evidence for this fix and should not be read as such.

**Gate:** `pnpm typecheck && pnpm test && pnpm build` — 107 files, **2340 tests passed**, build
success.

## What I could not verify

- **Whether any provider actually stops generating or billing on cancellation.** Not observable from
  our side. The claim this PR stands behind is that clodex stops relaying and requests cancellation.
- **No wire capture against a provider proving the upstream connection closed.** Teardown is proven
  at the socket level against a local upstream in tests; the real-provider run shows only that the
  gateway survives a mid-stream disconnect and keeps serving.
- **Sites 1 and 4 have no real-provider exercise**, because no stock configuration produces an
  Anthropic-format catalog entry to point at. They are covered by tests only.
- **Windows and Linux were not exercised at runtime.** No platform-specific code is touched here, but
  the disconnect-event characterisation was measured on macOS.
- **The `'aborted'`-listener characterisation is scoped to this helper's callers** — the inference and
  count_tokens handlers, which read the whole request body before starting upstream work. A handler
  that streams its request body upstream (e.g. `src/http-proxy/server.ts`, which does not use this
  helper and is not on an inference route) has not been characterised and the comment says so.

## Failure and rollback behaviour

The controller is created per request and aborts on exactly one condition. If it never fires, every
route behaves as it did before this PR — the existing idle and total timers are untouched and still
bound the request. If it fires spuriously, the affected request is cancelled and answered with
nothing, which the two negative tests exist to prevent. Non-abort failures are rethrown unchanged, so
upstream 500s, refused connections and malformed responses reach the client exactly as before (head
and base return identical results on all three raw routes). Reverting is a clean revert of one
commit: no persisted format, no `~/.clodex` state, and no migration is involved.

### DIFF
--- a/.claude/docs/registry-and-server.md
+++ b/.claude/docs/registry-and-server.md
@@ -52,6 +52,17 @@ every slot credential for deletion.
 and `/openai/v1` (via `src/openai-adapter.ts`). Wizard/quick-start settings persist to config;
 network mode requires a password; default port 17645 (`--port` overrides).
 
+Every inference and count_tokens request builds a client-disconnect controller (`watchClientDisconnect`
+in `src/http-utils.ts`, shared with `src/proxy.ts`) and passes its signal to the upstream call — the
+raw relays' `fetch`, and the SDK adapters' `abortSignal`. The controller is aborted at end of life on
+every path, so consumer abort listeners run at a deterministic point rather than at the garbage
+collector's convenience; the abort reason says which path it was. A response `close` before the
+response finished writing is a real disconnect (mid-stream as well as before any output) and aborts
+with `Client disconnected`; a response that finished writing aborts with `ResponseCompleted` after its
+last byte. Ask `clientDisconnected(signal)`, never `signal.aborted`, when deciding whether an error is
+worth reporting — a cancelled request writes no error response, because the client that would read it
+is the one that left, while non-abort failures still answer the client.
+
 Endpoint-mode request model resolution (`createGatewayModelCatalog` in `server/models.ts`) accepts,
 in precedence order: exact catalog id (and its gateway-discovery id) → unmasked gateway id when
 `--mask-gateway-ids` is on (`vendor-mask.ts`) → canonical `clodex:{provider}:{model}` id → saved
--- a/README.md
+++ b/README.md
@@ -362,7 +362,12 @@ clodex --version    # version
   avoid near-immediate termination; the 1h/6h ceilings allow deliberately long
   calls without leaving stalls attached indefinitely. At either deadline,
   clodex aborts the SDK call; cancellation is cooperative, so a provider
-  transport that ignores the abort signal can settle later. These are
+  transport that ignores the abort signal can settle later. clodex also asks
+  the provider to stop as soon as the client that made the request goes away —
+  a Ctrl-C, a killed agent, or a closed browser — instead of waiting for a
+  deadline, whether the answer had started arriving or not. The same
+  cooperative limit applies: clodex stops relaying and requests cancellation,
+  but cannot guarantee the provider stops generating. These are
   server-side limits, and callers may stop sooner. Claude Code currently
   defaults to about 180s of downstream byte silence in proxy mode and 300s in
   endpoint mode, with
--- a/src/http-utils.ts
+++ b/src/http-utils.ts
@@ -66,3 +66,82 @@ export function sendJson(res: ServerResponse, status: number, body: unknown): vo
   res.writeHead(status, { 'Content-Type': 'application/json' });
   res.end(json);
 }
+
+/**
+ * Brand carried by the abort reason for a response that finished normally.
+ * A registry symbol rather than a module-local one, so the brand still reads
+ * correctly if this module is ever instantiated twice.
+ */
+const RESPONSE_COMPLETED: unique symbol = Symbol.for('clodex.responseCompleted');
+
+/**
+ * Abort reason for a response that reached the end of its life normally, as
+ * opposed to a client that went away. Read it through `clientDisconnected`
+ * rather than by identity.
+ */
+export class ResponseCompleted extends Error {
+  readonly [RESPONSE_COMPLETED] = true;
+
+  constructor() {
+    super('Response completed');
+    this.name = 'ResponseCompleted';
+  }
+}
+
+/**
+ * True only when the abort came from the client going away before its response
+ * was complete. An end-of-life abort on a response that finished writing is
+ * not a disconnect, so callers deciding whether to stay silent about an error
+ * must ask this rather than reading `signal.aborted`.
+ */
+export function clientDisconnected(signal: AbortSignal): boolean {
+  if (!signal.aborted) return false;
+  const reason: unknown = signal.reason;
+  return !(typeof reason === 'object' && reason !== null && RESPONSE_COMPLETED in reason);
+}
+
+/**
+ * Request cancellation for a client that goes away before its response is
+ * complete. Callers pass the returned signal to the upstream work they start,
+ * so abandoned work can stop; whether it stops at once is up to the transport.
+ *
+ * The controller is aborted on every exit path, successful ones included: a
+ * response only ever closes once, and at that point nothing the request started
+ * has a reader left. Aborting unconditionally is defence in depth, not a fix
+ * for a measured leak — no controller or response was found retained after a
+ * forced GC without it. What it buys is that consumer abort listeners (undici's
+ * per-request one, for example) run at a deterministic point instead of waiting
+ * on the garbage collector.
+ *
+ * The two cases carry different reasons and must not be conflated:
+ * `ResponseCompleted` for a normal finish, and the long-standing
+ * `Client disconnected` error for a client that left. `clientDisconnected`
+ * is the only correct way to tell them apart — `signal.aborted` is now true
+ * for both.
+ *
+ * On a normal finish the abort always lands after the last byte: `res` emits
+ * `close` only once it is finished or destroyed, and the relay paths that
+ * return before their pipe drains still end `res` from that pipe. Measured over
+ * 300 streaming and 300 non-streaming requests: byte-identical output, no
+ * truncation, no unhandled error.
+ *
+ * An unfinished `close` is the only disconnect signal needed. Measured on Node
+ * 24.14.1 and on the Node 22 engines floor, it fires in all three disconnect
+ * shapes — before the response starts, after headers while a stream is in
+ * flight, and while the request body is still uploading — and every observed
+ * `IncomingMessage` `aborted` event arrived alongside one, microseconds apart.
+ * The deprecated `aborted` event therefore adds no case here. That was
+ * measured for this helper's callers, the inference and count_tokens handlers
+ * in `src/proxy.ts` and `src/server/router.ts`, which read the whole request
+ * body before starting any upstream work; a handler that streams its request
+ * body upstream has not been characterised and should not assume it.
+ */
+export function watchClientDisconnect(res: ServerResponse): AbortController {
+  const controller = new AbortController();
+  res.once('close', () => {
+    controller.abort(
+      res.writableFinished ? new ResponseCompleted() : new Error('Client disconnected'),
+    );
+  });
+  return controller;
+}
--- a/src/openai-adapter.ts
+++ b/src/openai-adapter.ts
@@ -2,7 +2,7 @@ import { tool, jsonSchema, streamText, generateText } from 'ai';
 import type { LanguageModel, ModelMessage } from 'ai';
 import { parseToolArguments } from './proxy-shared.js';
 import type { SdkCallParams } from './sdk-adapter.js';
-import { oauthServiceTier, reportUnsupportedServiceTier } from './sdk-adapter.js';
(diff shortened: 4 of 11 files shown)


bman654:
### Added: end-of-life abort contract

Per maintainer review, `watchClientDisconnect` now aborts its controller on **every** exit path,
including normal completion — previously `abort()` was never called on a successful request.

To avoid silently redefining what the existing guards mean, the two cases carry distinct reasons and
callers ask a predicate instead of reading `signal.aborted`:

```ts
const RESPONSE_COMPLETED: unique symbol = Symbol.for('clodex.responseCompleted');
export class ResponseCompleted extends Error { readonly [RESPONSE_COMPLETED] = true; /* … */ }
export function clientDisconnected(signal: AbortSignal): boolean { /* brand check */ }
```

All **8** call sites that previously read `clientAbort.signal.aborted` — 3 in `src/proxy.ts`, 5 in
`src/server/router.ts` — now call `clientDisconnected(clientAbort.signal)`, so their meaning is
unchanged. A registry `Symbol.for` brand is used rather than `instanceof` so a duplicated module
instance cannot reclassify a completion as a disconnect. The real-disconnect reason is byte-identical
to before.

**This is defence in depth, not a fix for an observed leak.** Two independent audits measured zero
retention on this path after forced GC (N=200/500/1000, real gateway and real MITM proxy: 0
controllers and 0 responses retained, flat heap slope, 1 keep-alive socket for 600 requests), on Node
v24.14.1 and the Node 22 engines floor. Node's `gcPersistentSignals` only admits timeout or non-empty
composite signals; a plain `new AbortController()` can never be registered.

Safety was measured, not assumed: 300 streaming + 300 non-streaming requests produced byte-identical
SSE output, 0 truncations, 0 non-200s, 0 unhandled errors. `relayAnthropicMessages` pipes upstream to
`res` without awaiting, so `res` `'close'` fires only after the pipe ends `res` — the completion abort
always lands after the last byte.

Mutations: making the completion abort conditional again → 2 red; `clientDisconnected` forced true →
**27 red** (every guard test); forced false → 3 red. Two characterization tests were updated rather
than deleted, and two new `proxy.ts` tests were added because guards 1 and 2 had no coverage at all.

Gate: typecheck clean, 107 files / 2342 tests, build success (sentinel verified present in the
bundle). Smoke on default proxy mode: `--model haiku` → HAIKU-OK, `--model luna` → LUNA-OK, 0 error
lines in either session log.

PR #188, opened 2026-09-06
fix(voice): restore dictation in proxy mode to prevent connection errors

FallDownTheSystem:

I couldn't use voice dictation through clodex in proxy mode, even with a native Claude model selected. It showed "Voice connection failed. Check your network and try again." This fixes that connection failure.

## Problem and root cause

Claude Code 2.1.263 opens `wss://api.anthropic.com/api/ws/speech_to_text/voice_stream` for dictation. That connection hits clodex's host-based interception, regardless of the selected model.

The proxy forwarded the request upstream but didn't handle the `101 Switching Protocols` response. The voice client never completed its WebSocket handshake. A local reproduction succeeded directly and through a generic CONNECT tunnel, but timed out through the intercepted Anthropic connection.

## The change

This adds a separate upgrade handler in `src/http-proxy/server.ts`. It reuses the existing upstream agent, certificate validation, and request-header filter. It forwards handshake headers and buffered first frames before piping data in both directions. Rejected upgrades remain HTTP responses with valid chunk framing.

Cleanup covers pending connections, half-open clients, and shutdown. Only upgraded sockets get separate ownership, so cleanup after a rejected handshake cannot destroy a connection another request has reused.

Existing HTTP request handling stays unchanged. No dependencies, binary patches, launch settings, or model-translation changes.

## Evidence

**I've only tested this on Windows 11. I haven't tested macOS or Linux.**

- Live dictation worked with Claude Code 2.1.263 and the local build linked through pnpm.
- All 18 WebSocket tests pass on Node 22.18.0 and 24.6.0. They use isolated `CLODEX_HOME` directories and local test servers. Proxy variables are cleared except in the explicit outbound-proxy test.
- Type checking, direct strict type checking of the new test file, and the build pass.
- Removing the upgrade handler made 13 of the 18 tests fail in a full-file run. Separate mutations caught dropped first frames, missing EOF/half-open cleanup, and incorrect pooled-socket ownership.

The full suite is **not green on Windows**. I reproduced 30 failures with unchanged `main`: 28 credential-helper `spawn EFTYPE` failures, one POSIX file-mode assertion, and one shutdown-attribution failure. I stopped the full run after an existing patcher test tried to execute its Unix-script `claude-2.1.999` fixture and opened Windows' application chooser. Those issues are outside this PR.

- [ ] `pnpm typecheck && pnpm test && pnpm build` (full-suite blockers above)
- [x] Feature-deletion mutation run against the full test file
- [ ] Native-text and translated-provider smoke tests (not run; live testing covered dictation)
- [x] Commit summary describes the user-visible fix

Upstream connection failures return HTTP 502. Rejected handshakes preserve their status and body. No persisted configuration or credentials need migration; reverting the commit restores the previous proxy behavior.

You're welcome to take over this PR or push any edits you'd like. Maintainer edits are enabled.

### DIFF
--- a/.claude/docs/claude-code-internals.md
+++ b/.claude/docs/claude-code-internals.md
@@ -362,6 +362,43 @@ open thinking block first (`case 'error'` in `src/sdk-adapter.ts`). Either alone
 and tool blocks are still closed: visible output already stops the retry, and a tool block's
 buffered arguments must be flushed.
 
+## Voice dictation transport (verified 2.1.263, darwin-arm64)
+
+How the dictation client reaches the network, read from the extracted `claude-2.1.263-*.js` bundle
+while reviewing #188. This is why proxy mode must relay a WebSocket upgrade and endpoint mode does
+not have to.
+
+- **The voice URL never follows `ANTHROPIC_BASE_URL`.** The client dials
+  `VOICE_STREAM_BASE_URL || <oauth-config>.BASE_API_URL` with `https://` rewritten to `wss://`
+  (@27520823), path `/api/ws/speech_to_text/voice_stream`. That base is the OAuth-config module's
+  hardcoded `https://api.anthropic.com`; `ANTHROPIC_BASE_URL` appears 77 times in the bundle and
+  never in that module. So endpoint mode, which points `ANTHROPIC_BASE_URL` at the local gateway and
+  strips `HTTP(S)_PROXY`, sends voice straight to Anthropic and needs no gateway support.
+- **The socket is Bun's native WebSocket behind a `ws` shim, not npm `ws`.** The module imports
+  `ws` bare and passes `{headers, proxy, tls}`; the shim forwards those to Bun's client. The
+  `proxy` value comes from the same env resolver the HTTP client uses (the `HTTPS_PROXY` family,
+  honouring `NO_PROXY`), with nothing platform-specific, so macOS and Windows both CONNECT through
+  clodex in proxy mode. `tls.ca` is built explicitly from `NODE_EXTRA_CA_CERTS` plus bundled and
+  system roots (@1604656, @1605611, @1609521); there is no `rejectUnauthorized:false` on the path,
+  so a bad CA fails closed rather than silently.
+- **Nothing is sent before the 101.** `send()` drops frames unless `readyState === OPEN`, audio
+  captured while connecting is queued outside the socket, and the first frame (`KeepAlive`) is
+  written from the `open` handler. The client's parser `head` on an upgrade is therefore always
+  empty for this client; the relay's `head`/`upstreamHead` forwarding is hardening, not a
+  production path.
+- **Failure mapping.** An HTTP response instead of a 101 goes through the `unexpected-response`
+  handler (@27524262) and renders as `Voice stream error: WebSocket upgrade rejected with HTTP
+  <status>`; only 4xx is marked fatal, anything else is retried once after 250 ms. The friendly
+  "Voice connection failed. Check your network and try again." comes from a recording-level selector
+  (@27531209) that fires when a >2 s recording ends with audio but the socket never connected —
+  which is exactly what a hung handshake produces. The handshake itself sits on Bun's native
+  `BUN_CONFIG_WS_HANDSHAKE_TIMEOUT`, default 120 s.
+- **Why main hung (Node, not Claude Code).** An `http(s).Server` with no `upgrade` listener does
+  *not* close the connection on Node >= 22.11; it falls the request through to the ordinary request
+  handler. clodex's passthrough then made an `https.request` with no `upgrade` listener of its own,
+  so when the origin answered 101 Node destroyed that socket and `response` never fired, leaving
+  the client waiting on the 120 s timeout above. Verified on 22.11.0, 22.14.0 and 24.14.1.
+
 ## Things that looked like clodex bugs and were not (not version-specific)
 
 - **"Concurrent subagents died at turn 2" was not unknown-model classification.** The agents' first
--- a/.claude/docs/launch-and-wrapper.md
+++ b/.claude/docs/launch-and-wrapper.md
@@ -138,5 +138,16 @@ Claude Code's own `NO_PROXY` matcher has two behaviors worth knowing before chan
 `no_proxy || NO_PROXY` means **lowercase wins outright — do not union the casings**, and `*` is
 bypass-all **only as the entire value** (a list-member `*` matches nothing).
 
+## WebSocket upgrade forwarding in proxy mode
+
+WebSocket upgrade requests inside intercepted `api.anthropic.com:443` connections use
+`forwardAnthropicUpgrade` (`src/http-proxy/server.ts`). It connects to the fixed Anthropic
+origin with the ordinary passthrough agent and TLS settings, preserves the upgrade status and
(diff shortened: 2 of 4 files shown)


bman654:
nice.  thanks for the submission.  I'll take a look.  I never noticed b/c I use VoiceInk.

bman654 (approved):
Thanks — this is a clean, well-scoped fix and the PR body is exactly the shape we want. It holds up on macOS.

I reviewed head `4455417` with a three-lens panel (socket lifecycle, test discrimination, integration/reachability) plus a cross-family refuter pass, all on macOS 15 / Node 24.14.1 with an isolated `CLODEX_HOME` and proxy env cleared.

## The Windows-only gap is closed

- **macOS gate:** `pnpm typecheck && pnpm test && pnpm build` — 108 files / 2406 tests green, build clean. The full-suite blocker you hit is Windows-specific; nothing in this PR is behind it.
- **Live vendor, both directions, no credentials:** an unauthenticated upgrade to the real `wss://api.anthropic.com/api/ws/speech_to_text/voice_stream` through the PR proxy returns 729 bytes byte-identical to a direct connection (101 + error text frame + close frame; only `Date`/`request-id`/`CF-RAY`/`sec-websocket-accept` differ). Same request against a main-equivalent build (only the `mitmServer.on('upgrade')` registration removed, rebuilt): **zero bytes for 40 s, no FIN, no RST.** Also confirmed with a real `ws@8` client (close 1008 "Invalid authorization" preserved) and through a chained corporate-style CONNECT proxy, where the CONNECT carries `https-proxy-agent`'s signature — so the outbound-proxy test is pinning the real agent, not a fallback.
- **Root cause confirmed, not assumed:** with no `upgrade` listener Node ≥22.11 falls the request through to the ordinary handler → `forwardRawAnthropicRequest` → upstream answers 101 → the `ClientRequest` has no `upgrade` listener, so Node destroys the socket and `response` never fires → client waits on Bun's 120 s native handshake timeout.
- **The rejected-upgrade path against the real origin** (`/v1/models` with upgrade headers → 503 + `Content-Length: 95`) relays 441 bytes identical to direct, so the `ServerResponse` re-framing preserves fixed-length bodies verbatim and re-chunks EOF-delimited ones correctly.
- **Client behaviour from the 2.1.263 bundle:** the proxy URL comes from the `HTTPS_PROXY` family and `NO_PROXY` only, nothing platform-specific, so macOS takes the same path into clodex that you observed on Windows. `NODE_EXTRA_CA_CERTS` is what makes the MITM leaf trusted (2.1.263 reads it explicitly into `tls.ca`), and there is no `rejectUnauthorized:false` on the voice path, so a broken CA fails closed.

## Tests

Feature deletion reproduces your 13/18 red in a full-file run; the five survivors each pin a separate behaviour (rejection semantics through the ordinary path, the non-Anthropic CONNECT negative), none is theatre. Ten consecutive macOS runs and a run concurrent with all four other proxy suites were green; no handle, listener, spy, or temp-dir residue; the `_getActiveHandles` filter resolves to exactly one raw TCP socket. Twelve guard mutations: `head`/`upstreamHead`/`end`-handler/`ServerResponse`/`rejectUnauthorized`/header-filter each killed by the right test.

## Demoted after checking (stating these so nobody re-derives them)

- **Endpoint mode is unaffected.** The voice base URL is the hardcoded OAuth-config `BASE_API_URL`; `ANTHROPIC_BASE_URL` (77 occurrences in the bundle) never reaches that module, and endpoint mode strips `HTTP(S)_PROXY`, so voice dials Anthropic directly there. No doc note needed.
- **502 is the right degraded UX** but does not produce the "Voice connection failed" text the title quotes — an HTTP response instead of 101 renders as `WebSocket upgrade rejected with HTTP 502` after one 250 ms client retry. Strictly better than main's 120 s hang; no change requested.
- **Sharing the keep-alive agent** can hand the upgrade a just-closed pooled socket (`reusedSocket === true` reproduced) with no `ECONNRESET` retry like the HTTP path has; the 5 s pool idle timeout plus the client's own retry make this self-heal. Nothing to do.
- **`sockets.add(socket)` on the 101 path is necessary, not redundant:** post-upgrade the socket is in neither `agent.freeSockets` nor `agent.sockets`, so `close()` could not otherwise reach it. And the client side is already covered by the pre-existing `proxyServer.on('connection')` tracking of the raw TCP socket — mutating that away makes `close()` hang, so shutdown ownership is real, not a cascade artefact.
- **Long dictation sessions are safe:** neither the agent's 5 s `timeout` nor Node's server `requestTimeout`/`headersTimeout`/`keepAliveTimeout` touch an upgraded socket (tested with those shortened to 1–2 s; socket survived 8–13 s idle). 8 MiB pushed each way through an established relay with no stall.
- **The `head`/`upstreamHead` branches are correct but not exercised by production:** Bun's native client never sends before `open`, and Cloudflare sends the 101 in its own segment ~15–35 ms ahead of the first frame (mutating out the `upstreamHead` write changed nothing in 3/3 live runs). Keep them; just don't read the mutation result as production-path coverage.

## Notes — all hardening, none blocking

1. **Rejected-upgrade bodies over ~64 KiB stall forever.** Node removes its `drain` forwarder from a socket it hands to `'upgrade'`, and `assignSocket` does not re-add one, so `upstreamRes.pipe(response)` pauses on the first backpressured write and never resumes (16 KiB completes; 64/128/512 KiB stop at ~114 KB with no FIN, no error, no timeout; same on Node 22.11/22.14/24). Real Anthropic error bodies are tiny, so a voice session cannot hit it; a captive-portal or corporate error page could. One line after `assignSocket` fixes it: `clientSocket.on('drain', () => response.emit('drain'))`, attached only in the non-101 callback.
2. **A pipelined request followed by an upgrade on the same MITM connection** throws `ERR_HTTP_SOCKET_ASSIGNED` out of the `'response'` handler (Node emits `upgrade` immediately, before the queued response finishes). With no routes configured that is an uncaught exception and the proxy process exits; with routes, `src/proxy.ts`'s log-only handler swallows it and the socket hangs. The 101 variant writes the handshake ahead of the still-unfinished earlier response. Not reachable from Claude Code, curl, or gh — needs a hand-rolled HTTP/1.1 pipelining client — so it is a NOTE. An `_httpMessage` guard at function entry (destroy and return) covers both variants; a guard only before `assignSocket` leaves the 101 interleave in place.
3. **The `PassThrough` buffer is unpinned.** Replacing `clientData.pipe(socket)` with `clientSocket.pipe(socket)` stays 18/18 green because the "each side of the handshake" test sends request+frame in one write, which lands in the parser's `head`, not the buffer. A delayed-101 test (frame in a separate write after the origin has seen the handshake) discriminates it. Also, the "bounded buffer" comment holds for memory but not for EOF above ~128 KiB of pre-handshake data — synthetic for this client.
4. Cosmetic: `shouldKeepAlive = false` only adds `Connection: close` when upstream sent no `Connection` header; a keep-alive rejection is relayed verbatim and then the socket is closed. No real client reuses a rejected-upgrade socket.

I pushed one commit on top of yours (maintainer edits, as you offered) that closes 1–3 exactly as described: 10 source lines in `forwardAnthropicUpgrade` (a `drain` forwarder on the non-101 path, an `_httpMessage` entry guard) plus four tests — the large-body rejection, both pipelined shapes, and the delayed-101 frame — each proven red on your head and green with the fix. Full suite 2410 green on macOS. Approving; I'll do a live dictation run on macOS before merging.

PR #195, opened 2026-09-09
fix(openai): keep the Artifact tool working on OpenAI models instead of failing every request

### LINKED ISSUE #194: OpenAI models fail with 400: Artifact schema pattern is not a valid regex
# OpenAI models fail with 400: `Artifact` schema pattern is not a valid regex

## Summary

After successfully patching Claude Code 2.1.266 with Clodex, all OpenAI-routed models fail when Claude Code sends a request containing the `Artifact` tool.

The OpenAI API returns:

```text
API Error: 400 Invalid schema for function 'Artifact':
'^(?!__.*__$)[^\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}"\\\\./[\\]]{1,200}$'
is not a 'regex'. (HTTP 400)
```

This happens when using Clodex-routed OpenAI models from Claude Code.

I tested switching between different OpenAI models (`terra` and `luna`), and the problem appears to be the `Artifact` tool schema rather than a specific model.

## Environment

* Windows
* Node: `v22.19.0`
* Claude Code: `2.1.266`
* Clodex: latest installed version
* OpenAI provider: ChatGPT/Codex OAuth
* Claude Code installed through npm/NVM
* Claude Code patched successfully with `CLODEX_CLAUDE_PATH`

Claude Code:

```text
2.1.266 (Claude Code)
```

The patch itself completed successfully:

```text
clodex patch: 11 applied, 0 skipped, 0 failed
```

## Reproduction

I configured GPT-5.6 Terra:

```powershell
clodex models --alias terra=clodex:openai-oauth:gpt-5.6-terra
```

Then launched:

```powershell
clodex claude
```

Inside Claude Code:

```text
/model
```

I selected:

```text
terra
```

Then simply sent:

```text
hello
```

The request failed immediately with:

```text
API Error: 400 Invalid schema for function 'Artifact':
'^(?!__.*__$)[^\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}"\\\\./[\\]]{1,200}$'
is not a 'regex'. (HTTP 400)
```

Trying again produced the same error.

I then switched to another OpenAI model:

```text
/model
```

and selected:

```text
luna
```

This also indicates that the issue is related to the shared Claude Code tool schema being translated for OpenAI, rather than being specific to GPT-5.6 Terra.

## Relevant schema

The rejected pattern is:

```text
^(?!__.*__$)[^\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}"\\\\./[\\]]{1,200}$
```

The problematic parts appear to be the Unicode property escapes:

```text
\\p{Cc}
\\p{Cf}
\\p{Zl}
\\p{Zp}
```

combined with the negative lookahead:

```text
(?!__.*__$)
```

Claude Code itself can use this schema against the Anthropic API, but the schema is rejected when translated into the OpenAI function/tool schema format.

## Additional evidence

I found another recent report describing the same `Artifact` regex being rejected when Claude Code tool schemas are translated to OpenAI-compatible function schemas.

That report specifically identifies the Unicode property classes and negative lookahead as the constructs being rejected by strict OpenAI-side schema validation.

This makes me think Clodex may need to normalize/sanitize this particular schema when converting Claude Code tools to OpenAI tools.

## Possible fix

Could Clodex handle `Artifact`'s schema before sending it to OpenAI?

Possible approaches might be:

1. Convert the unsupported regex pattern into an OpenAI-compatible equivalent.
2. Remove the `pattern` constraint when translating this tool schema to OpenAI.
3. Sanitize unsupported JavaScript/Unicode regex constructs during Claude → OpenAI tool-schema conversion.
4. Add a compatibility transform specifically for Claude Code's `Artifact` tool.

Ideally this should be handled automatically so users don't have to modify Claude Code's tool definitions.

## Important distinction

The Claude Code installation itself is working.

The Clodex patch also succeeds:

```text
Patched claude 2.1.266: 2 models, 2 aliases, 2 context windows.

clodex patch: 11 applied, 0 skipped, 0 failed
```

Claude Code starts normally and `/model` successfully switches between the configured OpenAI models.

The failure occurs only when an actual inference request is sent to the OpenAI model.

## Expected behavior

Claude Code should be able to send its normal tool definitions through Clodex to OpenAI without OpenAI rejecting the `Artifact` schema.

If a Claude Code tool schema contains regex constructs that OpenAI does not support, Clodex should either translate them into a compatible schema or safely remove the unsupported constraint.

## Suggested regression test

It would be useful to add a test containing the current Claude Code `Artifact` schema and verify that:

```text
Claude Code tool schema
        ↓
Clodex translation
        ↓
OpenAI function schema
```

produces a schema accepted by the OpenAI API.

This would also protect against future Claude Code releases introducing additional JSON Schema features that are valid for Anthropic but not accepted by OpenAI.

## Confirmed workaround

Claude Code already provides an environment variable to disable the Artifact tool:
So run this in windows powershell

```powershell
$env:CLAUDE_CODE_DISABLE_ARTIFACT="1"
```

Then start Clodex normally:

```powershell
clodex claude
```

With Artifact disabled, the OpenAI model works correctly.

For example, after applying the workaround:

```text
Claude Code v2.1.266
luna · Claude Pro

❯ hi

Hi! How can I help you today?
```

So this confirms that the `Artifact` schema is the component triggering the OpenAI 400 error.

### Permanent workaround on Windows

The environment variable can be persisted for the current Windows user:

```powershell
[Environment]::SetEnvironmentVariable(
    "CLAUDE_CODE_DISABLE_ARTIFACT",
    "1",
    "User"
)
```

Then **close and reopen PowerShell/Terminal**.

Verify it with:

```powershell
$env:CLAUDE_CODE_DISABLE_ARTIFACT
```

It should return:

```text
1
```

After that:

```powershell
clodex claude
```

will start with Artifact disabled automatically.

### Suggested Clodex fix

The permanent Clodex-side fix should ideally be equivalent to the upstream CLIProxyAPI solution:

* Detect schema `pattern` values containing unsupported Unicode property escapes such as `\p{...}` / `\P{...}`.
* Remove only those incompatible `pattern` constraints when translating Claude tool schemas to the OpenAI schema.
* Keep other valid regex patterns intact.
* Avoid replacing the regex with a wildcard pattern because that can change the schema's intended semantics.

Alternatively, Clodex could explicitly handle the `Artifact` tool's schema so that OpenAI-compatible models do not receive an unsupported regex.

### Important

The `CLAUDE_CODE_DISABLE_ARTIFACT=1` workaround is confirmed to work, but it disables the Artifact tool.

A proper Clodex fix would allow Artifact to remain enabled while making its schema compatible with OpenAI models.

Thanks!

marcelstenkewitz:
Claude Code 2.1.266 ships a tool called Artifact, and OpenAI rejected the description of one of its
inputs. Because that tool is sent with *every* request, every OpenAI-routed session died on the
first message — `hello` came back as `API Error: 400 ... is not a 'regex'`. The only workaround was
`CLAUDE_CODE_DISABLE_ARTIFACT=1`, which turns the tool off. clodex now removes just the piece OpenAI
cannot read, and leaves the tool working.

Fixes #194.

## User-visible failure

Any `clodex:openai*` model, any prompt, from the first turn:

```
API Error: 400 Invalid schema for function 'Artifact':
'^(?!__.*__$)[^\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}"\\\\./[\\]]{1,200}$' is not a 'regex'. (HTTP 400)
```

Reported on Windows / Node 22.19.0 / Claude Code 2.1.266 against `gpt-5.6-terra` and
`gpt-5.6-luna`; reproduced here on macOS against `gpt-6-astra` (see Evidence).

## Root cause

OpenAI validates every `pattern` in a function schema by compiling it **with Python's `re`** — the
message `'<pattern>' is not a 'regex'` is jsonschema's format-checker wording — while Claude Code
writes its schemas in the ECMAScript dialect. The two dialects diverge on Unicode property escapes:

```
>>> re.compile(r'^(?!__.*__$)[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}"\\./[\]]{1,200}$')
re.error: bad escape \p at position 14
```

**The lookahead is not the problem**, contrary to the issue text and to the other reports it cites.
Python compiles `(?!...)` and `(?<=...)` fine; only `\p{}` / `\P{}` fail. Artifact's `collection`,
`doc_id` and `contract` patterns are all lookahead-bearing and all compile — of its 8 patterns,
exactly 1 is rejected.

Reachability: `translateTools` in `src/sdk-adapter.ts` is the single translation path for every
non-Anthropic provider, Claude Code sends its full tool list on every request, and the 400 is
returned before any model output. Nothing about the route is optional or vestigial — it is the
default proxy-mode path for a `clodex:` model id, and the OAuth (`openai-oauth`) transport reads its
tools from the payload this function produces.

## The change

`src/tool-schema-sanitize.ts` (new, leaf module — imports nothing from `src/`, like its sibling
`tool-input-sanitize.ts`):

- `isCompatiblePattern` scans the pattern character by character so an escaped backslash (`\\p`, a
  literal backslash followed by `p`, which Python compiles) is not mistaken for the escape `\p`,
  which it does not. It rejects `\p{}`, `\P{}`, JS named groups `(?<name>` and `\k<>`
  backreferences, and deliberately keeps lookbehind, spelled the same in both dialects.
- `sanitizeToolSchema` walks the schema to any depth and removes only the offending `pattern`
  keyword. The property it constrained stays; no wildcard is substituted, so nothing in the schema
  starts claiming something untrue. Unchanged subtrees are returned by identity.

`src/sdk-adapter.ts` applies it in `translateTools` for every route **except** Anthropic-format ones
(`@ai-sdk/anthropic`, `VERTEX_ANTHROPIC_NPM`), which take Claude Code's patterns as written.

Losing the keyword loses a hint, not a guard: Claude Code validates tool input against its own
schema before executing, so a value the pattern would have rejected still fails there — as a
tool_result the model can retry, not a dead session.

### Deliberately left out

- **`translateOpenAiRequest` in `src/openai-adapter.ts`** is the only sibling arrival point for the
  same signal (`jsonSchema(` appears in exactly two places in `src/`; this is the other). It is not
  sanitized here: those schemas come from a third-party client that chose OpenAI format, not from
  Claude Code, the function has no `npm` in scope to gate on, and nothing has reported it. Two lines
  if you want it.
- **No rewrite of `\p{Cc}` into explicit ranges.** `\p{Cf}` alone is a scattered set across five
  planes; a hand-expanded approximation would be a new source of truth that silently drifts from
  Unicode.

## Tests

`tests/tool-schema-sanitize.test.ts` (new, 9) — the real Artifact `field` and `collection` patterns,
escaped-backslash handling, named groups vs. lookbehind, nesting through `anyOf`/`items`, a tool
parameter *named* `pattern` (Grep has one) surviving untouched, and identity for schemas with
nothing to strip.

`tests/sdk-adapter.test.ts` (2) — the regression test the issue asked for: Artifact through
`translateRequest` into a real `@ai-sdk/openai` provider with a fake `fetch`, asserting the
**serialized wire body**, plus a negative proving an Anthropic-format route keeps the pattern by
identity.

**Feature-deletion mutation**, full-file runs (not `-t`): making `sanitizeToolSchema` a pass-through
turns 3 tests red across both files — the two content assertions and the wire-body regression. The
identity/negative tests stay green under the mutation, as they should.

## Evidence

Environment: macOS 24.6.0, **Node 22.19.0** (the repo's `.nvmrc` pins 24.14.1, which is not
installed on this machine; 22 is what `engines` allows and what the reporter runs), pnpm 10.34.5,
no ambient proxy vars.

- **Wire A/B against the live API, identical payload.** Two `clodex server --endpoint --saved
  --no-discovery` instances, one from this branch (port 17699) and one from the published 2.11.4
  (17700), same POST carrying Artifact's real `field` and `collection` patterns to
  `gpt-6-astra`:
  - stock 2.11.4 → `400 Invalid schema for function 'Artifact': ... is not a 'regex'` — the
    reported error, verbatim
  - this branch → `200`, `{"type":"text","text":"ok"}`
- **Python-`re` check on the whole Artifact schema:** 8 patterns before sanitizing, 1 rejected;
  7 after, 0 rejected. The one dropped is the reported `field` pattern.
- **Product smoke test, proxy mode (default), real Claude Code 2.1.266:**
  `clodex claude -- --model terra -p 'Reply with exactly: MODEL-OK'` → `MODEL-OK`.
  Anthropic passthrough leg: `--model claude-sonnet-5` → `PASSTHROUGH-OK`.
- Standard gate: `pnpm typecheck && pnpm test && pnpm build` green — 113 files, 2445 tests — and
  green again under `CLODEX_HOME="$(mktemp -d)"`. `git merge-tree` against current `main` is clean;
  no open PR I can see touches `sdk-adapter.ts`'s tool path.

### What I could not verify

- The claim that OpenAI validates with Python's `re` is an inference from the error wording plus the
  observed accept/reject split (lookahead accepted, `\p{}` rejected). It is not from OpenAI
  documentation. The fix does not depend on the inference being the whole story — it depends on the
  observed A/B.
- Only ChatGPT/Codex OAuth credentials were available. `@ai-sdk/openai` API-key, Azure,
  openai-compatible, Google, Mistral and xAI routes were not exercised; all take the same code path
  and can only lose a `pattern` they never accepted.

### Failure and rollback

If `sanitizeToolSchema` were wrong in either direction the blast radius is one JSON Schema keyword:
too eager drops a constraint the model was only ever *advised* by, too shy restores today's 400. It
holds no state, touches nothing in `~/.clodex`, and needs no migration. Reverting the two-line
`translateTools` change restores the previous behavior exactly.

https://claude.ai/code/session_01LPewWu4VNToxMtFRtPibjn

### DIFF
--- a/.claude/docs/translation.md
+++ b/.claude/docs/translation.md
@@ -36,6 +36,14 @@ hand-rolled per-provider translation. Preserved hard-won behavior:
   change non-streaming responses' existing omission of reasoning, or the transport's prohibition
   on replaying already-emitted model output. The guarantee covers SDK-visible summary text,
   grouping and encrypted content, not output-only fields the SDK omits (such as `status`).
+- **A tool schema's `pattern` is dropped when Python's `re` cannot compile it**, on every route but
+  Anthropic-format ones (`src/tool-schema-sanitize.ts`). OpenAI validates each `pattern` with Python
+  — its 400 reads `'<pattern>' is not a 'regex'`, jsonschema's format-checker wording — while Claude
+  Code writes ECMAScript. Artifact's `field` constraint uses `\p{Cc}`, Python answers `bad escape
+  \p`, and since the tool ships on every request the session was dead from `hello` onward (#194).
+  The dialects agree on lookaround, so Artifact's `(?!...)` patterns are left alone; only `\p{}`,
+  `\P{}`, JS named groups and `\k<>` backreferences go. Losing the keyword loses a hint, not a
+  guard — Claude Code still validates tool input against its own schema before executing.
 - **Images in `tool_result` are lifted out of the text-only function-output channel** and delivered
   as real image parts on the following user message. Inline, a JSON.stringify'd base64 screenshot
   tokenizes at ~1.5 chars/token — 200k+ tokens per screenshot, killing agents with "Prompt is too
--- a/src/sdk-adapter.ts
+++ b/src/sdk-adapter.ts
@@ -19,6 +19,8 @@ import {
 } from './provider-factory.js';
 import { resolveUpstreamTools } from './tool-search.js';
 import { sanitizeToolInput } from './tool-input-sanitize.js';
+import { sanitizeToolSchema } from './tool-schema-sanitize.js';
+import { VERTEX_ANTHROPIC_NPM } from './constants.js';
 import type { AnthropicRequestMessage, AnthropicToolDefinition } from './proxy-types.js';
 import { anthropicErrorType, sdkUpstreamErrorDetails, upstreamHttpStatus } from './upstream-error.js';
 import { upstreamRequestBudget } from './upstream-retry.js';
@@ -456,12 +458,15 @@ function toolRequiredProps(tools?: SdkCallParams['tools']): Map<string, Readonly
 
 export function translateTools(anthropicTools?: AnthropicTool[], npm?: string): Record<string, ReturnType<typeof tool>> | undefined {
   if (!anthropicTools?.length) return undefined;
+  // Anthropic-format routes take Claude Code's ECMAScript patterns as written;
+  // every other provider validates them in a dialect that may not compile them.
+  const anthropicFormat = npm === '@ai-sdk/anthropic' || npm === VERTEX_ANTHROPIC_NPM;
   const tools: Record<string, ReturnType<typeof tool>> = {};
   for (const t of anthropicTools) {
     if (!t.name || !t.input_schema) continue;
     tools[t.name] = tool({
       description: t.description ?? '',
-      inputSchema: jsonSchema(t.input_schema),
+      inputSchema: jsonSchema(anthropicFormat ? t.input_schema : sanitizeToolSchema(t.input_schema)),
       strict: npm === '@ai-sdk/openai' ? false : undefined,
     });
   }
--- a/src/tool-schema-sanitize.ts
+++ b/src/tool-schema-sanitize.ts
@@ -1,0 +1,92 @@
+// tool-schema-sanitize.ts — drop `pattern` constraints a non-Anthropic provider cannot compile.
+//
+// Leaf module by design: it imports nothing from src/, like its sibling
+// tool-input-sanitize.ts, so the translation path can use it from anywhere in
+// the import graph.
+//
+// Claude Code writes its tool schemas in the ECMAScript regex dialect. OpenAI
+// validates every `pattern` by compiling it with Python's `re` — the 400 reads
+// `Invalid schema for function 'X': '<pattern>' is not a 'regex'`, which is
+// jsonschema's format-checker wording — and the two dialects do not agree.
+// Claude Code 2.1.266's Artifact tool spells its `field` constraint with
+// Unicode property escapes (`[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}...]`), Python's `re`
+// answers `bad escape \p`, and every OpenAI-routed request 400s before the
+// model sees the turn — the tool is sent on every request, so the session is
+// dead from `hello` onward (#194).
+//
+// The dialects agree on more than the issue reports assume: lookahead and
+// lookbehind compile in both, so Artifact's `collection` pattern
+// (`^(?!\.\.?(?:\/|$))...`) is left alone. Only the constructs Python actually
+// rejects are removed, and only the `pattern` keyword goes — never the property
+// it constrains, and never a wildcard in its place, so nothing in the schema
+// starts claiming something untrue.
+//
+// Dropping the keyword loses a hint, not a guard: Claude Code validates tool
+// input against its own schema before executing, so a value the pattern would
+// have rejected still fails there, as a tool_result the model can retry.
+
+/**
+ * True when Python's `re` can compile `pattern`, which is what OpenAI's schema
+ * validator does with it. Scans rather than pattern-matches so an escaped
+ * backslash (`\\p`, a literal backslash followed by `p` — compilable) is not
+ * mistaken for the Unicode property escape `\p` that is not.
+ */
+export function isCompatiblePattern(pattern: string): boolean {
+  for (let i = 0; i < pattern.length; i++) {
+    if (pattern[i] === '\\') {
+      const next = pattern[i + 1];
+      // \p{...} / \P{...}: Python has no Unicode property escapes ("bad escape \p").
+      if ((next === 'p' || next === 'P') && pattern[i + 2] === '{') return false;
+      // \k<name>: a JS named backreference; Python spells it (?P=name).
+      if (next === 'k' && pattern[i + 2] === '<') return false;
+      i++; // an escaped character never starts a construct
+      continue;
+    }
+    // (?<name>...): a JS named group; Python spells it (?P<name>...). Lookbehind
+    // is (?<= / (?<! in both dialects and stays.
+    if (
+      pattern[i] === '(' && pattern[i + 1] === '?' && pattern[i + 2] === '<'
+      && pattern[i + 3] !== '=' && pattern[i + 3] !== '!'
+    ) return false;
+  }
+  return true;
+}
+
+/**
+ * Return `schema` with every `pattern` keyword Python's `re` cannot compile
+ * removed, at any depth (`properties`, `items`, `anyOf`, `$defs`, …).
+ *
+ * A `pattern` whose value is not a string is a property of the tool named
+ * `pattern` (Grep has one) rather than the keyword, and is walked like any
+ * other subschema. Unchanged subtrees are returned by identity, so a schema
+ * with nothing to strip — every schema but Artifact's, today — is not rebuilt.
+ */
+export function sanitizeToolSchema(schema: unknown): unknown {
+  if (Array.isArray(schema)) {
+    let changed = false;
+    const out = schema.map(entry => {
+      const next = sanitizeToolSchema(entry);
+      changed ||= next !== entry;
+      return next;
+    });
+    return changed ? out : schema;
+  }
+  if (!schema || typeof schema !== 'object') return schema;
+
+  let changed = false;
+  const out: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(schema as Record<string, unknown>)) {
+    if (key === 'pattern' && typeof value === 'string') {
+      if (!isCompatiblePattern(value)) {
+        changed = true;
+        continue;
+      }
+      out[key] = value;
+      continue;
+    }
+    const next = sanitizeToolSchema(value);
+    changed ||= next !== value;
+    out[key] = next;
+  }
+  return changed ? out : schema;
+}
--- a/tests/sdk-adapter.test.ts
+++ b/tests/sdk-adapter.test.ts
@@ -110,6 +110,77 @@ describe('translateTools', () => {
     ]);
   });
 
+  it('sends Artifact without the pattern OpenAI cannot compile (#194)', async () => {
+    // Claude Code 2.1.266 sends this schema on every request; OpenAI compiles
+    // each `pattern` with Python's `re`, answers `bad escape \p`, and 400s the
+    // turn before the model sees it.
+    const fieldPattern = String.raw`^(?!__.*__$)[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}"\\./[\]]{1,200}$`;
+    const collectionPattern = String.raw`^(?!\.\.?(?:\/|$))[A-Za-z0-9_\-.~:@+]{1,200}$`;
+    const requestBodies: unknown[] = [];
+    const provider = createOpenAI({
+      apiKey: 'synthetic-test-key',
+      fetch: async (_input, init) => {
+        requestBodies.push(JSON.parse(String(init?.body)));
+        return new Response(JSON.stringify({
+          id: 'resp_synthetic',
+          model: 'gpt-5.6-terra',
+          output: [],
+          usage: {
+            input_tokens: 1,
+            input_tokens_details: { cached_tokens: 0 },
+            output_tokens: 0,
+            output_tokens_details: { reasoning_tokens: 0 },
+          },
+        }), { status: 200, headers: { 'content-type': 'application/json' } });
+      },
+    });
+    const params = translateRequest({
+      model: 'gpt-5.6-terra',
+      messages: [{ role: 'user', content: 'hello' }],
+      tools: [{
+        name: 'Artifact',
+        description: 'Render an HTML file to an Artifact',
+        input_schema: {
+          type: 'object',
+          properties: {
+            field: { type: 'string', pattern: fieldPattern },
+            collection: { type: 'string', pattern: collectionPattern },
+          },
+        },
+      }],
+    }, '@ai-sdk/openai', { openAiOAuth: true });
+
+    await generateAnthropicResponse(
+      provider.responses('gpt-5.6-terra'), params, 'gpt-5.6-terra');
+
+    expect(requestBodies).toEqual([
+      expect.objectContaining({
+        tools: [{
+          type: 'function',
+          name: 'Artifact',
+          description: 'Render an HTML file to an Artifact',
+          strict: false,
+          parameters: {
+            type: 'object',
+            properties: {
+              field: { type: 'string' },
+              collection: { type: 'string', pattern: collectionPattern },
+            },
+          },
+        }],
+      }),
+    ]);
+  });
+
+  it('leaves the pattern intact on an Anthropic-format route', () => {
+    const pattern = String.raw`^\p{L}+$`;
+    const input_schema = { type: 'object', properties: { field: { type: 'string', pattern } } };
+    for (const npm of ['@ai-sdk/anthropic', '@ai-sdk/google-vertex/anthropic']) {
+      const tools = translateTools([{ name: 'Artifact', input_schema }], npm);
+      expect((tools!.Artifact.inputSchema as { jsonSchema: unknown }).jsonSchema).toBe(input_schema);
+    }
+  });
+
   it('returns undefined for empty/missing tools', () => {
     expect(translateTools(undefined)).toBeUndefined();
     expect(translateTools([])).toBeUndefined();
--- a/tests/tool-schema-sanitize.test.ts
+++ b/tests/tool-schema-sanitize.test.ts
@@ -1,0 +1,85 @@
+import { describe, it, expect } from 'vitest';
+import { isCompatiblePattern, sanitizeToolSchema } from '../src/tool-schema-sanitize.js';
+
+// Claude Code 2.1.266's Artifact schema, verbatim: `field` is the pattern OpenAI
+// rejected in #194, `collection` the lookahead that compiles in both dialects.
+const ARTIFACT_FIELD_PATTERN = String.raw`^(?!__.*__$)[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}"\\./[\]]{1,200}$`;
+const ARTIFACT_COLLECTION_PATTERN =
+  String.raw`^(?!\.\.?(?:\/|$))[A-Za-z0-9_\-.~:@+]{1,200}(?:\/(?!\.\.?(?:\/|$))[A-Za-z0-9_\-.~:@+]{1,200}){0,14}$`;
+
+describe('isCompatiblePattern', () => {
+  it('rejects the Unicode property escapes Python cannot compile', () => {
+    expect(isCompatiblePattern(ARTIFACT_FIELD_PATTERN)).toBe(false);
+    expect(isCompatiblePattern(String.raw`^\p{L}+$`)).toBe(false);
+    expect(isCompatiblePattern(String.raw`^\P{Nd}+$`)).toBe(false);
+  });
+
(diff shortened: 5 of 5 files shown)


FallDownTheSystem:
I was looking into the same issue. My own fix was basically the same with a few differences, so I had CC analyze the differences, just commenting them here in case they're useful:

I found four defects in `tool-schema-sanitize.ts` and reproduced each one against the branch. None of them is triggered by a Claude Code tool today, so they do not block #194. They matter because this code now runs on every tool schema Claude Code sends, on every request, and a mistake in either direction is silent: too eager and the model loses guidance it should have had; too lenient and the user is back to the 400 that started this issue.

**1. A pattern can be dropped when it is actually fine.**
The scanner looks for the JavaScript named-group syntax `(?<name>`, which OpenAI rejects. But it does not know when it is inside a character class (the square-bracket part of a regex, where those same characters are plain literals). So a pattern that only lists `(`, `?` and `<` as allowed characters is wrongly treated as a named group and removed.

```
isCompatiblePattern(String.raw`^[(?<]+$`)  // false, but OpenAI accepts this pattern
```

Fix: remember whether the scanner is inside `[...]` and only run the named-group check outside it.

**2. One rejected construct gets through, and would reproduce the original 400.**
JavaScript allows a code point to be written as `\u{41}` (braces around the hex). I sent that form through a clodex gateway to the OpenAI Responses API and it fails with the same `is not a 'regex'` error as the Artifact pattern. The scanner does not look for it, so a tool using that spelling would fail every request exactly as #194 did.

```
isCompatiblePattern(String.raw`^\u{41}+$`)  // true, but OpenAI returns 400 for it
```

Fix: one more line in the escape branch, treating `\u{` and `\x{` the same way as `\p{`. No Claude Code tool uses this today; it is the same family of syntax and the same failure.

**3. The walker edits data that is not part of the schema.**
JSON Schema keywords like `enum`, `const`, `default` and `examples` hold values, not rules. The walker treats every nested object as a rule set, so a value that happens to contain a key called `pattern` is edited. That changes what the schema allows:

```
sanitizeToolSchema({ type:'string', enum:[{pattern:'\\p{L}'}], default:{pattern:'\\p{L}'}, 'x-meta':{pattern:'\\p{L}'} })
// → {"type":"string","enum":[{}],"default":{},"x-meta":{}}
```

Here `enum` originally allowed one specific object; after sanitizing it allows an empty object instead. Fix: only walk through the keywords that are known to contain sub-schemas (`properties`, `items`, `anyOf`, `$defs`, and so on) and leave everything else exactly as it was. That also removes the need for the special case that guesses whether a `pattern` key is a keyword or a tool parameter name: a parameter named `pattern` always lives under `properties`, whose keys the walker would never inspect.

**4. A property named `__proto__` disappears.**
JavaScript treats the key `__proto__` specially when you write `out[key] = value` on a plain object, so a tool parameter with that name is silently dropped from the copy:

```
// properties: { __proto__: {type:'string'}, f: {type:'string', pattern:'\\p{L}'} }
Object.keys(sanitizeToolSchema(schema).properties)  // ['f']  — __proto__ is gone
```

Tool schemas arrive as parsed JSON, so this key is possible. Fix: build the copy with `Object.fromEntries(entries)`, which does not have this behavior.

bman654 (approved):
Thanks for this — it fixes a live, total outage, and the diagnosis at the centre of it is right in a
way the issue and the other public reports are not.

Reviewed at head `4fd092f1` with a four-lens panel plus two adversarial refuters, and independently
against a cold investigation of #194 that was started before this PR existed. **No blockers —
approving.**

Because this fixes a live outage, we're merging now and taking the remaining items ourselves rather
than making you do another round trip: we'll open follow-up issues for the asks below and clear them
today. If you'd rather land any of them yourself, say so and they're yours — particularly ask 1,
which is a design improvement you'd have good judgement on.

## What we confirmed, so you don't have to re-derive it

- **Your root cause is right, and we upgraded it from inference to observation.** An 18-pattern
  ablation against the live Codex Responses backend reproduces your accept/reject split exactly:
  `(?!…)`, `(?=…)`, `(?<=…)`, backreferences, `(?P<n>…)`, `\Z` and conditionals are all **accepted**;
  `\p{…}`/`\P{…}` are rejected. That split is an exact fingerprint of Python's `re`, and it rules out
  RE2 (which would have rejected the backreference). Your "I could not verify this" caveat can be
  relaxed.
- **The fix reaches the reported route.** We captured the actual `socket.send` frame on the OAuth
  WebSocket: `tools[0].parameters.properties.field === {type:'string'}`, `collection.pattern`
  retained, no `\p{Cc}` bytes.
- **Your safety argument holds, verified in the 2.1.266 bundle rather than assumed.** Every dispatch
  branch was enumerated: the generic executor `safeParse`s before `validateInput` and before `call`
  (10296), PreToolUse-updated input is re-checked (8438), permission-handler `updatedInput` again
  (10298), Artifact re-applies the regex before the DB write (26062), and the route-to-another-host
  branch is **not** a bypass — it `safeParse`s before transport (36068).
- **Prompt-cache bytes are stable.** Two full `translateRequest` + real `@ai-sdk/openai`
  serializations of the same request produced byte-identical bodies. Hashing the unsanitized schema
  at `sdk-adapter.ts:158` is the conservative direction.
- **Your mutation claim reproduces exactly** — pass-through `sanitizeToolSchema` turns exactly 3
  tests red. Gate inversion and gate-widening both fire too.
- **Demotions, stated out loud.** We chased and cleared: Bedrock Claude being mis-gated (it uses
  Bedrock's transport, not the Anthropic Messages wire, and every production caller is already behind
  `isSdkMigratedNpm`, so `@ai-sdk/anthropic` and `undefined` can never reach `translateTools`);
  cyclic-schema recursion (`input_schema` arrives via `JSON.parse`, can't be cyclic); and a panel
  suggestion to replace your deny-list with an allow-list, which we refuted — the prototype had
  false accepts of its own, and CLIProxyAPI's merged fix for this identical bug is *narrower* than
  yours.

## Asks

**1. Walk only schema-bearing positions.** This is the one substantive change, and it subsumes three
separate findings. The walker has no notion of schema position, so it edits instance data:

```
{"const":{"pattern":"\\p{L}"}}      => {"const":{}}
{"enum":[{"pattern":"\\p{L}"}]}     => {"enum":[{}]}
{"default":{"pattern":"\\p{L}"}}    => {"default":{}}
```

`const` and `enum` change validation semantics, not a hint — which contradicts the PR body's "nothing
in the schema starts claiming something untrue" and "the blast radius is one JSON Schema keyword".
Recursing through a declared keyword list (`properties` values, `items`/`prefixItems`, `anyOf`/
`oneOf`/`allOf`/`not`, `$defs`/`definitions`, `if`/`then`/`else`, `contains`, `propertyNames`,
`additionalProperties`, `patternProperties`) fixes it, and — as @FallDownTheSystem notes above —
**also removes the need for the `typeof value === 'string'` special case**, because a parameter named
`pattern` always lives under `properties`, whose keys you'd never inspect. It's what CLIProxyAPI
ships. Reachable today via an MCP/plugin tool schema, which Claude Code forwards verbatim.

**2. Check `patternProperties` keys — this one is a proven live 400.** They're regexes too, and the
walker only looks at values. An isolating A/B against the real Codex backend with `strict:false`:

```
A  ordinary pattern ^[a-z]+$                     -> 200
B  pattern keyword  ^\p{L}+$                     -> 400 "is not a 'regex'"
C  bad regex ONLY in a patternProperties KEY     -> 400  same message
D  patternProperties with valid key ^[a-z]+$     -> 200
H  bad patternProperties key nested under $defs  -> 400
```

C vs D proves the far side compiles the key. python-jsonschema's metaschema agrees
(`patternProperties.propertyNames` carries `format: "regex"`). CLIProxyAPI gained this handling only
after a review bot flagged the same gap in their first pass. Falls out of ask 1 naturally.

**3. Add `\u{` (and `\x{`) to the escape branch — one line.** Three of us found this independently,
and two confirmed it live: `^\u{41}+$` passes `isCompatiblePattern` but Python answers
`incomplete escape \u`, and @FallDownTheSystem got the same 400 through a clodex gateway. It's the
same `u`-flag family as the `\p{}` that caused #194, and there are 14 `\u{…}` regexes in the 2.1.266
bundle. Latent today — none is in a schema — but it's the cheapest possible close.

**4. `__proto__` — use `Object.fromEntries`.** `out[key] = next` invokes the prototype setter, so a
tool parameter named `__proto__` silently vanishes from the copy. No global pollution; just a lost
property.

**5. Optional: make the `(?<` check character-class-aware.** `^[(?<]+$` is rejected today but Python
compiles it fine. Costs a dropped hint, not a 400 — take it or leave it.

## Prose corrections

The code is more careful than the prose, and a wrong claim in this repo has propagated into later
work before, so these matter more than they look.

- **`isCompatiblePattern`'s stated contract is false.** "True when Python's `re` can compile
  `pattern`" — it isn't; it's a targeted scanner for the constructs 2.1.266 actually emits. Suggested:
  *"True when `pattern` contains none of the Python-incompatible constructs this targeted sanitizer
  recognizes."*
- **"The dialects agree on lookaround" / "lookbehind, spelled the same in both dialects"** is wrong
  on variable-width lookbehind: JS permits it (ES2018), Python's `re` requires fixed width. Same
  claim is in `.claude/docs/translation.md`. Worth adding that OpenAI *does* reject lookaround on
  strict paths (`vercel/ai#16021`, `SmartBear/smartbear-mcp#491`) — we confirmed that's unreachable
  here **because clodex sends `strict: false`**, which is a documented opt-out, not the accident it
  looks like. Please say so; it's load-bearing for keeping `collection`'s lookahead.
- **`contract` is not lookahead-bearing.** It's `^(0|[1-9]\d{0,3})\.(0|[1-9]\d{0,4})\.(0|[1-9]\d{0,5})$`.
- **The same pattern is also on `ArtifactData`,** not only `Artifact` — worth not naming a single tool
  in the changelog line.
- **"8 patterns, exactly 1 rejected"** counts *sites*, not distinct values. Artifact has 10 sites / 7
  values; across all 9 pattern-bearing tools there are 15 distinct values. The reject count is right.
- **"`jsonSchema(` appears in exactly two places"** — there are three (`openai-adapter.ts:117`, `:120`,
  `sdk-adapter.ts:469`). Two *files*, which is probably what you meant.
- **`\p{Cf}` spans three planes, not five** (0, 1, 14) — checked on Unicode 16 and 17.
- **`CLAUDE_CODE_DISABLE_ARTIFACT=1` isn't the only workaround;** 2.1.266 also honours the
  `enableArtifact: false` setting.
- **"contrary to the issue text" is a bit unfair** — #194's *suggested fix* specifically said to
  detect and remove `\p{...}`/`\P{...}`. It hedged on the cause; it didn't get the fix wrong.
- **"any `clodex:openai*`" and "can only lose a `pattern` they never accepted"** outrun the evidence —
  only ChatGPT/Codex OAuth was exercised. Shared-code-path is source evidence, not proof that Azure,
  Google, Mistral or xAI share OpenAI's acceptance behaviour; some may accept a pattern this drops.
- **`.claude/docs/translation.md`'s "every route but Anthropic-format ones"** is wider than the truth.
  Two other arrival points forward schemas untouched: `translateOpenAiRequest` (which you disclose)
  and a raw `completionsUrl` relay at `src/server/router.ts:681-719` (which you don't). Neither is a
  defect for #194 — both start from an OpenAI-format client that owns its own `parameters` — but the
  doc should say "every Anthropic-input SDK-translated route but Vertex Anthropic". Note your stated
  reason for skipping `translateOpenAiRequest` ("no `npm` in scope") isn't quite it — `router.ts:743`
  has it. The scope call is still fine; just give it the real reason.
- **Test fixture naming:** `sdk-adapter.test.ts:118`'s `collectionPattern` is actually Artifact's
  `doc_id` pattern. The bundle builds both from one fragment — `O$ = ^${N}$` (doc_id) and
  `ZG = ^${N}(?:/${N}){0,14}$` (collection). `tool-schema-sanitize.test.ts` has the real collection
  pattern; only this one is mislabelled.

## Tests

Two gaps worth closing alongside the asks: nothing pins that the input schema is **not mutated in
place** (an in-place sanitizer plus a widened gate passes 102/102 while destroying the Anthropic
pattern), and the `\k<>` branch is unpinned — its fixture `(?<y>a)\k<y>` trips the named-group check
first, so deleting the `\k` guard entirely stays green.

Environment for everything above: macOS, Node 24.14.1, pnpm 10.34.5, isolated `CLODEX_HOME`,
`CLAUDE_CODE_ENTRYPOINT=cli`, and re-run under `HTTPS_PROXY=http://127.0.0.1:9` where an ambient
proxy could have mattered.

bman654:
@marcelstenkewitz I reopened this — I think you closed it having seen only the critique above, about
twelve minutes before our review landed. Sorry for the timing; that's on our side for taking a while.

**This PR is approved and we're merging it.** The panel found no blockers. Your central diagnosis —
that the negative lookahead is innocent and only `\p{}` is at fault — is right, and it contradicts
issue #194, the upstream reports it cites, and two other bridges' shipped fixes. We independently
confirmed it with an ablation against the live backend: lookahead, lookbehind, backreferences,
`(?P<n>…)` and `\Z` are all accepted; only `\p{…}`/`\P{…}` are rejected. That split is a Python `re`
fingerprint and rules out RE2. Your "I could not verify this" caveat was more modest than it needed
to be.

We also verified your safety argument against the 2.1.266 bundle rather than assuming it: Claude Code
really does `safeParse` tool input before `validateInput` and before `call`, PreToolUse-updated input
is re-checked, and the route-to-another-host branch is not a bypass. And your mutation claim
reproduced exactly — pass-through `sanitizeToolSchema` turns exactly 3 tests red.

The items in @FallDownTheSystem's comment are real — we found most of them independently — but every
one is latent: no shipped Claude Code tool triggers any of them, and your fix handles all 15 `pattern`
values 2.1.266 actually sends, with zero false accepts and zero false rejects. They're follow-ups, not
objections. Since this fixes a live outage we've filed them ourselves rather than making you do
another round trip:

- #197 — walk only schema positions, and check `patternProperties` keys (that one's a proven live 400)
- #198 — `\u{…}` escapes, the JSDoc contract wording, `__proto__`

If you'd like to take either, say the word and it's yours — #197 in particular is a design call you'd
have good judgement on. Otherwise we'll clear them today.

Thanks for a genuinely well-evidenced PR. The wire A/B against two server instances was the right
experiment, and it's what let us skip re-litigating the root cause.

Issue #204, opened 2026-09-09
clodex patch --restore still picks a backup by version tag when no manifest exists


bman654:
Remaining leg of #199. #199's manifest leg is fixed — restore refuses when the patch manifest
records a different install, or records this one but names pristine bytes that are gone. **When no
manifest exists at all, selection still rests on the version tag**, so `clodex patch --restore` can
still publish one install's pristine bytes over a different install.

## Reachable without wiping anything

A successful `--restore` **deletes the manifest**. So:

1. Two same-version installs, native A and npm B (different files — the population #199 names).
2. Patch A. Manifest records A, backup X is stored.
3. `clodex patch --restore` on A — succeeds, and **removes the manifest**. X stays on disk.
4. A is uninstalled or `TWEAKCC_CC_INSTALLATION_PATH`/PATH now reaches B.
5. `clodex patch --restore` → exit 0, **B's bytes replaced with A's pristine bytes**, one warning.

`~/.clodex` wiped or a different `CLODEX_HOME` reach the same state. The plan carries a note saying
the version tag is all that ties the backup to the install, but it does not refuse.

## It can also launder itself into a manifest

Patch B, patch A, `--restore` A (manifest gone), then re-patch B. Both backups are on disk and
disagree, so clodex refuses with `Found conflicting pristine backups`. If the user removes B's own
backup, `clodex patch` re-seeds B from A's bytes and writes a manifest recording
`binaryPath: B, pristineSha256: <A's hash>`. Every later restore then "correctly" publishes A's
bytes over B. (The refusal message no longer says "remove the wrong one" for this reason, but the
sequence is still available to anyone who picks a file.)

## Two candidate fixes

1. **Inspect before restoring.** `clodex patch`'s implicit restore extracts the source and checks
   for a `/*ccpatch:` marker; `--restore` never looks at the live binary. Requiring either a manifest
   for this install or a patch marker in the live bytes closes the sequence above outright — B is
   pristine, so there is nothing to restore. `readClaudeBundle` is read-only, so a native binary
   needs no copy; an npm install is plain JavaScript. Needs the per-format probe
   (`scripts/probe-patch-mechanism.mjs`) because this read path is Mach-O/ELF/PE sensitive, and a
   fallback for when extraction fails, so that a binary broken by a bad patch can still be rescued.
2. **Per-backup provenance.** Record which install each backup was made for (a sidecar beside the
   `.orig`, or a manifest keyed by binary path instead of a single slot), so elimination is not
[...]

Issue #199, opened 2026-09-09
clodex patch --restore can overwrite an install with a different install's backup


bman654:
`clodex patch --restore` can overwrite a Claude Code install with **another install's** pristine
bytes, and then delete the manifest, leaving no backup of what it clobbered.

Found during the review of the #193 fix. **It is pre-existing and independent of that change** — it
reproduces with no launchers involved at all.

## Mechanism

`selectRestoreSource` (`src/patch-backup.ts:240-285`): when the manifest's recorded `binaryPath`
differs from the currently-resolved target, the manifest is discarded, and restore falls back to
selecting the sole backup carrying the same **version tag**. It never establishes that those bytes
belong to the current install. "Same version" is treated as install provenance; it isn't.

## Why that matters — same version, different bytes, both supported

The npm-platform binary and the native-installer binary for the *same* Claude Code version are
genuinely different artifacts:

```
$ npm pack @anthropic-ai/claude-code-darwin-arm64@2.1.263 --ignore-scripts
native: bytes=205905136 sha256=eaa531ef...4cde1  Mach-O 64-bit arm64
npm:    bytes=199257984 sha256=ef5d2909...f4f9  Mach-O 64-bit arm64
cmp_exit=1
```

Both are labelled 2.1.263.

## Reproduction (no launchers)

Patch direct target A; change `TWEAKCC_CC_INSTALLATION_PATH` to a pristine same-version target B;
run `--restore`:

```
DIRECT_RETARGET_RESTORE={"patchCode":0,"firstManifestBinary":".../direct-retarget/wrong",
 "restoreCode":0,"rightStillOriginal":false,"rightNowEqualsWrongPristine":true,"manifest":null}
```

B now holds A's bytes, no backup of B exists, and the manifest is gone.

## Who triggers it

A user with two supported same-version installs whose bytes differ (npm + native installer is the
common shape — issue #193's reporter ended up with exactly that), who patches one, then has the
selected target change — via `CLODEX_CLAUDE_PATH`/`TWEAKCC_CC_INSTALLATION_PATH`, a PATH change, the
native-vs-PATH ranking, or a corrected launcher — and runs `--restore` before the second install has
acquired its own backup.

## Fix

When a manifest exists but records a different `binaryPath`, refuse rather than falling through to
version-only backup selection. Restore should require provenance (recorded path and/or content
hash), not a matching version tag. `--restore` is the rescue path; it must not be able to damage an
install it has no backup for.

bman654:
[...]

Issue #193, opened 2026-09-09
Windows: clodex patch detects claude.cmd instead of the native claude.exe


bloodykheeng:
# Windows: `clodex patch` detects `claude.cmd` instead of the native `claude.exe`

## Summary

On Windows, when Claude Code is installed through npm/NVM, `clodex patch` detects the `claude.cmd` launcher instead of the actual native Claude Code binary.

In my case, Claude Code was installed with:

```powershell
npm install -g @anthropic-ai/claude-code
```

and the actual native binary was:

```text
C:\nvm4w\nodejs\node_modules\@anthropic-ai\claude-code\bin\claude.exe
```

However, Clodex detected:

```text
C:\Users\bk\AppData\Local\nvm\v22.19.0\claude.cmd
```

This caused patching to fail.

The issue can be worked around with `CLODEX_CLAUDE_PATH`, but I think Clodex should automatically resolve the `.cmd`/`.ps1` launcher to the actual Claude Code native binary on Windows.

## Environment

* Windows
* Node: `v22.19.0`
* Claude Code: `2.1.266`
* `@bman654/clodex`: installed globally
* Claude Code installed with npm
* NVM for Windows

Claude Code installation:

```powershell
npm list -g @anthropic-ai/claude-code
```

Output:

```text
C:\nvm4w\nodejs -> .\
`-- @anthropic-ai/claude-code@2.1.266
```

Claude resolves to these launchers:

```powershell
where.exe claude
```

Output:

```text
C:\nvm4w\nodejs\claude
C:\nvm4w\nodejs\claude.cmd
```

PowerShell shows that `claude.ps1` ultimately launches the real native binary:

```text
C:\nvm4w\nodejs\node_modules\@anthropic-ai\claude-code\bin\claude.exe
```

Claude itself initially worked normally:

```powershell
claude --version
```

```text
2.1.266 (Claude Code)
```

### 1. Installed Clodex

```powershell
npm install -g @bman654/clodex
```

Installation completed successfully.

### 2. Signed into OpenAI

First attempt:

```powershell
clodex providers auth openai
```

returned:

```text
fetch failed
```

Running it again succeeded:

```text
Signed in to OpenAI (ChatGPT) — credential saved to the credential store.
```

### 3. Added GPT-5.6 Terra

I ran:

```powershell
clodex models
```

and selected:

```text
GPT-5.6 Terra (gpt-5.6-terra) (OpenAI OAuth (ChatGPT))
```

Then:

```powershell
[...]

PR #205, opened 2026-09-09
fix(patch): refuse to restore Claude Code from a backup that belongs to a different install

bman654:
If you have two Claude Code installs of the same version — an npm install beside a native one —
`clodex patch --restore` could copy the first install's pristine bytes over the second, tell you it
succeeded, and delete the record that was the first install's only way back. You ended up with two
damaged installs and a backup of neither. It now refuses instead, and tells you which install the
backup it found actually belongs to.

## The failure

`clodex patch --restore` (and the implicit restore inside `clodex patch`) selected a pristine backup
by matching **version tag**, and treated that as proof of which install the bytes came from. It
isn't. The npm platform package and the native installer ship genuinely different files under the
same version — for 2.1.263, 199,257,984 bytes versus 205,905,136 — and both are supported.

Reachable sequence, no contrived state: patch install A, have the resolved target become install B
(a stale `CLODEX_CLAUDE_PATH`/`TWEAKCC_CC_INSTALLATION_PATH`, a PATH change, a corrected launcher, or
A simply being uninstalled), run `--restore`. The manifest recorded A, so it was discarded for
naming a different path, and selection fell through to "the backup tagged with this version" — which
was A's. Issue #199, reported against real 2.1.266 binaries; the npm-beside-native shape is what
issue #193's reporter already runs.

## The change

The patch manifest is the only record of which install a backup was made for, and it holds **one**
install — so it can confirm a backup but can never rule one in by elimination. `selectRestoreSource`
now refuses, naming what to do instead, in the two states where that record contradicts the target:

- it records a **different** `binaryPath` — it vouches for nothing in the backup directory;
- it records **this** target but the pristine bytes it named are gone or corrupt — its own testimony
  says the same-version backups still on disk were made for some other install.

A manifest speaks only for the `claudeVersion` it was written for. After an upgrade it names a
backup of the old version, which is not among the new version's candidates at all, so neither
refusal fires across a version change.

`src/patcher.ts` only loses a stale comment and gains one explaining why deleting the manifest after
a restore is now sound. I first added a guard there to keep another install's manifest; the redesign
makes that unreachable, so it was removed rather than left in as dead code.

### What I deliberately left out

**Disqualifying only the backup the manifest names, and restoring the survivor.** That was my first
implementation and it is unsound: the manifest holds one install, so every earlier install's backup
is an unrecorded orphan with the same version tag, and "restore the one it did not name" hands a
third install's bytes to the target. It converted a safe `conflicting pristine backups` refusal on
`main` into a destructive copy. Both review passes found it independently. A regression test pins the
three-install shape specifically.

**The no-manifest case — `--restore` can still pick a backup by version tag when no manifest exists
at all, and that is reachable: a successful `--restore` deletes the manifest.** Tracked in #204 with
two candidate fixes and the laundering variant. Do not read this PR as closing #199 entirely; it
closes the manifest leg, which is what #199's Fix section asks for, and the note the plan now
carries says out loud when selection is resting on a version tag.

**`realpathSync.native` for path identity.** One install spelled two ways (case differences reach
this on APFS, not just Windows) reads as two installs and now refuses — safe, but a false refusal,
and it blocks `clodex patch` as well. Fixing it invalidates the `binaryPath` in every existing
manifest, so it needs a migration. Recorded in #204; `evaluatePatchState` compares the same way.

Two adjacent corrections came out of review: the `conflicting pristine backups` message no longer
says "remove the wrong one" (on a two-install machine both files are genuine, and deleting one
launders the wrong bytes into a manifest the new rules then trust), and the new version-tag warning
no longer suggests checking that the restored binary runs — both artifacts are working Claude Code.

## Tests

Nine new cases across the planner and the end-to-end command, including the three-install shape, the
deleted-recorded-backup shape, and both directions of version scoping.

Feature-deletion mutations, full-file runs, `tests/patch-backup.test.ts` + `tests/patcher-command.test.ts`:

| Mutation | Red |
| --- | --- |
| drop the different-install refusal | 8 |
| drop the recorded-backup-gone refusal | 5 (incl. a pre-existing corrupt-backup test) |
| drop version scoping | 3 |
| restore the rejected "disqualify and keep the survivor" design | 8, incl. the three-install test |

The three-install test is green on `main` (it refuses there for a different reason); it exists to
stop the unsound design coming back, and the table says which mutation isolates it.

## Runtime evidence

Real Claude Code 2.1.263 Mach-O binaries, isolated `CLODEX_HOME` and `TWEAKCC_CONFIG_DIR`, install B
a byte-differing copy of the same pristine bytes. Node 24.14.1, macOS 15, `CLAUDE_CODE_ENTRYPOINT=cli`.

| | `main` (d6a901b) | this branch |
| --- | --- | --- |
| `--restore` targeting install B | exit 0, "Restored pristine claude 2.1.263" | exit 1, names install A |
| B's bytes | **replaced with A's pristine bytes** | unchanged |
| manifest | **deleted** | kept |
| `--restore` targeting A afterwards | impossible, no manifest | exit 0, A pristine and runs |

`pnpm typecheck && pnpm test && pnpm build` green: 2513 tests, 114 files.

## Boundaries I could not verify

- **macOS only.** The planner is pure and platform-independent, and nothing here touches the
  read/repack/restore path, so `scripts/probe-patch-mechanism.mjs` was not run; if a reviewer wants
  ELF/PE coverage for the end-to-end command that is a reasonable ask.
- The second install in the real-binary run is a byte-modified copy of a genuine 2.1.263 binary, not
  a separately published npm artifact — the two published artifacts are independently known to
  differ in size and hash, but I did not install both.
- No provider legs were smoke-tested: this change is confined to backup selection and cannot reach
  the launch, gateway, or translation paths.

## Failure behavior

Every new path is a refusal before any write — exit 1, the binary untouched, the manifest left in
place. The rescue path is unaffected in the states that matter: a manifest recording this install
still restores it, including when the binary is too broken to report its version, and the
stale-launcher rescue from #193 still works.

Refs #199

### DIFF
--- a/.claude/docs/patcher.md
+++ b/.claude/docs/patcher.md
@@ -614,7 +614,47 @@ tweakcc's own repack reads back as an ordinary module name.
   the version probed from the binary being patched, its hash must match its own name, and a legacy
   `claude-<ver>.orig` (no hash in the name, possibly mislabeled by an older clodex) must
   additionally report that version when executed (`verifyPristineSource`). Conflicting or
-  unverifiable backups produce a loud error, never a copy. This gate covers both consumers —
+  unverifiable backups produce a loud error, never a copy. **A matching version tag is not install
+  provenance**: the npm platform package and the native installer ship different files under the same
+  Claude Code version, and both are supported, so one machine can hold two same-version installs whose
+  bytes differ. The manifest is the only record of which install a backup was made for, and it holds
+  **one** install — so it can confirm a backup but can never rule one in by elimination. Two states
+  therefore refuse outright rather than fall through to version-tag selection (issue #199, reproduced
+  on real 2.1.266 binaries in issue #199, and again on real 2.1.263 bytes with a byte-differing
+  second copy standing in for the second install; the published 2.1.263 npm and native artifacts are
+  independently known to differ in both size and hash):
+  - the manifest records a **different** `binaryPath` than the resolved target — it vouches for
+    nothing in the backup directory, and the error names the install it does belong to;
+  - the manifest records **this** target but the pristine bytes it named are gone or corrupt — its own
+    testimony says the same-version backups still on disk were made for some other install.
+
+  A manifest speaks only for the `claudeVersion` it was written for. After an upgrade it records a
+  backup of the OLD version, which is not among the new version's candidates at all — so neither
+  refusal fires across a version change, and a restore that was never in danger is not rejected.
+
+  Disqualifying only the backup the manifest names is NOT sufficient and was rejected during review:
+  the manifest holds one install, so every earlier install's backup is an unrecorded orphan carrying
+  the same version tag, and "restore the one it did not name" hands a third install's bytes to the
+  target — turning a `conflicting pristine backups` refusal into a destructive copy. **What remains
+  unprotected is the no-manifest case**, and it is reachable by an ordinary sequence: a successful
+  `--restore` DELETES the manifest, so patch A → restore A → have the target resolve to install B →
+  restore again publishes A's pristine bytes over B, with only a warning. (`~/.clodex` wiped or a
+  different `CLODEX_HOME` get there too.) Selection there still rests on the version tag alone.
+  Worse, a user who answers the `conflicting pristine backups` refusal by deleting one file can
+  launder the wrong bytes into a manifest that the rules above then trust on sight — which is why
+  that message no longer says "remove the wrong one". Closing this needs provenance recorded per
+  backup rather than a single-slot manifest, or requiring `--restore` to see a `/*ccpatch:` marker in
+  the live binary before it copies anything (`clodex patch`'s implicit restore already inspects the
+  source; `--restore` never does). Tracked separately — do not read the rules above as a full
+  guarantee.
+
+  Path identity is plain string equality, so a manifest written under a different spelling of the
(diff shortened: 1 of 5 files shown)


bman654:
Merged with admin after all five checks passed; the branch rule needs one review and GitHub blocks self-approval on an own-authored PR, so recording the review trail here instead.

Two adversarial panels ran before the first push and again after the rewrite:

- The **first implementation was unsound** — it disqualified only the backup the manifest names and restored the survivor. Two reviewers independently showed that hands a *third* install's orphaned backup to the target, converting a safe `conflicting pristine backups` refusal on main into a destructive copy. Rewritten so the manifest can only ever confirm a backup, never rule one in by elimination.
- A **cold pass on the rewrite** found three false refusals across Claude Code version upgrades plus a diagnostic that stated two falsehoods; both fixed by scoping the manifest's testimony to the `claudeVersion` it was written for.
- Residual, deliberately not closed here and reachable because a successful `--restore` deletes the manifest: with no manifest at all, selection still rests on the version tag. Tracked in #204 with two candidate fixes.

Post-rebase onto d8767b5: typecheck, 2550 tests, build green; the three feature-deletion mutations still turn 8/5/3 tests red respectively; real 2.1.263 Mach-O binaries re-verified (refuse-then-rescue).

Issue #196, opened 2026-09-09
Using Clodex with the Claude Code VS Code Extension on Windows


bloodykheeng:
# Using Clodex with the Claude Code VS Code Extension on Windows

## Overview

This document explains how to run **Clodex with the Claude Code VS Code extension on Windows** and route Claude Code requests to OpenAI models such as **GPT-5.6 Luna** through Clodex.

This was tested on:

* Windows
* Node.js installed through NVM for Windows
* Claude Code `2.1.266`
* Clodex
* VS Code
* Claude Code VS Code extension
* OpenAI OAuth through Clodex

The final working setup does **not** require modifying `claude.cmd` or replacing the normal Claude Code executable.

The important part is to run Clodex in proxy mode and configure the Claude Code VS Code extension to use the proxy environment variables.

---

# 1. The problem

The normal terminal command:

```powershell
clodex claude
```

was working correctly and allowed Claude Code to use OpenAI models such as:

```text
luna
GPT-5.6 Luna
```

However, the Claude Code VS Code extension was launching Claude Code directly.

The result was:

```text
VS Code Claude Code
        ↓
Claude Code directly
        ↓
No Clodex proxy
        ↓
OpenAI model routing does not work
```

The goal was therefore:

```text
VS Code Claude Code
        ↓
Clodex proxy
        ↓
OpenAI OAuth
        ↓
GPT-5.6 Luna
```

---

# 2. Important Windows detail

One of the first approaches was to configure:

```json
"claudeCode.claudeProcessWrapper": "C:\\nvm4w\\nodejs\\clodex-claude.cmd"
```

This did **not** work.

Claude Code reported:

```text
Error spawning Claude: spawn EINVAL
```

The important lesson is that the Claude Code VS Code extension uses a native process-spawning path, and directly supplying the Clodex `.cmd` launcher caused Windows process spawning to fail.

The extensionless:

```text
C:\nvm4w\nodejs\clodex-claude
```

also did not work. Claude Code treated it as a native executable and produced:

```text
Claude Code native binary ... failed to launch
```

Therefore, on Windows, **do not assume that `clodex-claude.cmd` can simply be placed into `claudeProcessWrapper`.**

---

# 3. The solution: Clodex proxy mode

The solution was to use Clodex's proxy mode:

```powershell
clodex server --proxy
```

Run this from a PowerShell terminal.

The command starts the local HTTP/HTTPS proxy and prints something similar to:

```text
clodex proxy-mode server running

[...]

PR #208, opened 2026-09-10
docs: explain running clodex with the Claude Code VS Code extension o…

bloodykheeng:

Windows users running Claude Code's VS Code extension had no documented way to use clodex models
from the chat panel — the extension launches Claude itself, so there was nothing to hook into, and
the obvious attempt (pointing the extension's process wrapper at `clodex-claude`) fails on Windows.
This adds a setup page covering both what makes clodex models work there and what makes them show up
in the extension's model picker.

## Problem and root cause

The extension does not launch the npm-installed Claude Code binary that `clodex patch` patches. It
ships and launches its own copy at
`…\.vscode\extensions\anthropic.claude-code-<version>-win32-x64\resources\native-binary\claude.exe`.

On my machine that bundled copy was byte-identical by SHA-256 to the pristine backup clodex recorded
before patching (`pristineSha256` in `~/.clodex/patch-state.json`), confirming it was unpatched. The
model picker's entries are strings inside the binary, so the dropdown reflects whichever binary was
launched.

This produces a confusing half-working state, which is what I originally hit in #196: routing works
without the patch, because the proxy does the routing — a model name saved into
`~/.claude/settings.json` is passed through by an unpatched binary and mapped by clodex — but only
that one model is ever visible, because there is no picker list, just the saved string.

`docs/background-agents.md` already documents `CLAUDE_CODE_PROCESS_WRAPPER=clodex-claude`, and
`src/claude-wrapper.ts` uses the same contract the extension does. Two things stop it applying here:
on Windows npm installs `clodex-claude` as a shebang script plus `.cmd` and `.ps1` with no `.exe`,
and the extension spawns the wrapper without a shell, so the `.cmd` fails with `spawn EINVAL`; and
since the wrapper execs the binary path it is handed, it would exec the extension's unpatched
bundled copy anyway.

Reachability: any Windows user with the VS Code extension and an npm-installed Claude Code hits
this. Nothing unusual is required — it is the default shape of that combination.

## The change

Documentation only. Adds `docs/windows-setup.md` and links it from the README's bridge-modes section,
next to the existing `docs/background-agents.md` tip.

The page is split into two levels: proxy env vars in `claudeCode.environmentVariables` (models work),
then an optional `claudeCode.claudeProcessWrapper` shim that runs the clodex-patched binary instead
of the bundled one (models appear in the picker). It also documents the `spawn EINVAL` constraint and
the two behaviors Claude Code changes when a process wrapper is configured.

The shim resolves its target at runtime rather than hardcoding a path — `CLODEX_WRAPPER_TARGET`, then
`binaryPath` from clodex's own `patch-state.json`, then the binary the caller handed it. It mirrors
`getAppHome()` from `src/paths.ts` for `CLODEX_HOME`, and the final fallback means claude always
launches even with clodex absent, matching the fail-open behavior `clodex-claude` documents for
itself. So the same compiled exe works on any machine and follows a re-patch with no rebuild.

Deliberately left out: no code changes. Two clodex-side improvements suggested in #196 — shipping a
native Windows `clodex-claude` launcher, and letting the wrapper prefer the patched install over the
path it is handed — are not attempted here and belong in their own PR if wanted.

## Evidence

Environment: Windows 11, NVM for Windows, node v22.19.0, clodex 2.11.6, Claude Code 2.1.267,
Claude Code VS Code extension 2.1.267, ChatGPT/Codex-plan OAuth provider, live `clodex server --proxy`.

Both documented setups were executed on that machine before writing them up:

- Env-var setup: extension traffic routes through clodex, and a model saved as default via
  `clodex claude` is used by the extension (Claude Code prints
  `Using luna (from .claude\settings.json)`).
- Wrapper setup: after a window reload, Terra and Luna appear in the extension's model picker
  alongside the built-in models and are selectable.
- The bundled-binary SHA-256 comparison described above.
- The shim's resolution was exercised in four states: manifest present (ran the recorded binary,
  proven with a marker executable), manifest absent (fell back to the handed-in binary, printed
  `2.1.267`), real environment with a deliberately invalid handed-in path (still resolved via the
  real manifest, so the path could only have come from it), and nothing resolvable (exit 1 with a
  message, no hang).
- The Go listing was extracted programmatically from the committed page and compiled, confirming it
  builds as printed rather than as I had it locally.

Stated as inference, not observation: the two consequences of setting a process wrapper (skipped
update check, permission-mode resolution moving out of the CLI) were read from the extension's
bundled code, not reproduced as failures. The page labels them that way.

Not verified: any Node version manager other than NVM for Windows, and any provider other than
ChatGPT/Codex-plan OAuth.

- [ ] `pnpm typecheck && pnpm test && pnpm build` — typecheck and build pass; `pnpm test` does not
      pass in my environment: 16 files / 95 tests fail, 2428 pass. I am on node v22.19.0, while
      `.nvmrc` pins v24.14.1 and CI runs 24, so I cannot attribute the failures and did not
      investigate them. This PR changes no source, and no test reads README.md or docs/ (the only
      match is a prose mention in a comment in ws-upgrade-pacer.test.ts). `pnpm install` also needed
      a retry after an ERR_PNPM_EPERM file lock.
- [ ] Feature-deletion mutation run — N/A, no source changed and no test pins a doc
- [ ] Manual smoke — N/A, no launch path changed; the documented procedures were executed manually
      as described above
- [x] Commit summary line reads as a release note for a non-technical user

No runtime behavior changes and nothing is persisted or migrated. If the documented wrapper setup
misbehaves for a reader, removing the `claudeCode.claudeProcessWrapper` setting and reloading the
window returns them to the env-var-only setup; the page says so. Reverting this PR removes a page and
a README link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ flowchart LR
 > [!TIP]
 > Using Claude Code's agents view or background agents? Ask your Claude Code agent to read [docs/background-agents.md](docs/background-agents.md) and set it up for you — one global `clodex server --proxy` plus the `clodex-claude` wrapper bin bridges every claude process automatically.
 
+> [!TIP]
+> On Windows with the Claude Code VS Code extension? See [docs/windows-setup.md](docs/windows-setup.md) — proxy-mode env vars route the extension through clodex, and an optional process wrapper makes clodex models appear in its model picker.
+
 ## CLI reference
 
 ### `clodex claude [options] [claude-flags]`
--- a/docs/windows-setup.md
+++ b/docs/windows-setup.md
@@ -1,0 +1,290 @@
+# Using clodex with the Claude Code VS Code extension on Windows
+
+This page covers running clodex alongside Claude Code's **VS Code extension** on Windows, so
+`clodex:` models and aliases are usable from the extension's chat panel rather than only from a
+terminal.
+
+There are two levels of setup, and they solve different problems:
+
+| Setup | What you get |
+| --- | --- |
+| [Proxy env vars](#1-route-the-extension-through-clodex) | clodex models **work** in the extension |
+| [+ process wrapper](#2-make-clodex-models-appear-in-the-model-picker) | clodex models also **appear in the model picker** |
+
+The first is enough if you are happy selecting a model once and leaving it. Add the second if you
+want to switch between clodex models from the extension's UI.
+
+## 1. Route the extension through clodex
+
+The extension launches Claude Code itself, so there is no `clodex claude` step to hook into.
+Instead, run a proxy-mode server and point the extension's environment at it.
+
+**Start the server and leave it running** (a minimized terminal is fine):
+
+```powershell
+clodex server --proxy
+```
+
+It prints the values you need:
+
+```
+clodex proxy-mode server running
+  HTTPS_PROXY=http://127.0.0.1:17645
+  HTTP_PROXY=http://127.0.0.1:17645
+  NODE_EXTRA_CA_CERTS=C:\Users\<you>\.clodex\http-proxy\clodex-ca.pem
+```
+
+**Put those in your VS Code settings** (`Ctrl+Shift+P` → `Preferences: Open User Settings (JSON)`),
+using the values your server printed:
+
+```json
+"claudeCode.environmentVariables": [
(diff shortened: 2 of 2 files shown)


bman654 (approved):
Thanks for this. I'm sure it'll help some people out.

PR #210, opened 2026-09-11
chore(canary): say what to do about a release the canary could not fully check

bman654:
When the release-monitoring canary cannot check a new Claude Code release on every platform, its
Slack alert now opens by telling you what to do about it. The usual cause is simply that Docker
Desktop is not running, which stops two of the Linux builds being tested properly. Until now the
alert mentioned that only in a footnote below a long platform table, so it read as an unexplained
wall of warnings — and because an incompletely-checked release gets retried, the same message
arrived twice in one evening with nothing saying that starting Docker would settle it.

The alert also stops repeating itself: a note the new action line already made is no longer printed
again further down.

## User-visible failure

Two notifications, six hours apart, for Claude Code 2.1.268, identical in substance. Neither said
why coverage was short or what would fix it. Nothing was wrong with the release or with clodex —
Docker was quit, so `linux-arm64` and `linux-arm64-musl` fell back to a bundle-only probe, the run
was recorded `partial` rather than `pass`, and `partial` is deliberately retried every six hours
until a run achieves full coverage.

The second alert additionally listed eleven patch checks as "not attempted at all this time", which
was untrue.

Both reach every scheduled hourly run; no flags required.

1. **No action in the alert.** The message reported the hole and left the reader to infer the cause
   from the per-platform table.
2. **The eleven phantom dropped checks.** The per-platform baseline merge used jq's `*`, which
   merges recursively, so a platform's new entry inherited `sites` keys from the entry it replaced.
   That defeats the leg-mode guard in the comparison: a host/container leg records `probe:PATCH …`
   names that a probe-only leg never emits, so after one Docker outage the fresh probe entry carried
   the container run's leftover names and the next run reported them as dropped. `+` keeps the
   per-platform merge intact while making a leg's record exactly what that leg observed.

## Also fixed

Found while reviewing the above; each is reachable, and the first three are ways the canary could
fail in the direction it exists to prevent.

- `coverage_is_complete` decided the downgrade question with `[ -z "$(jq …)" ]`, so a jq that
  *failed* — no output — read as "nothing was downgraded" and returned full coverage. A green tick
  issued because the check itself broke.
- Composing the message inside `slack "$(…)" || true` let that `|| true` — which exists so a Slack
  outage cannot take the canary down — also swallow a failure in the renderer. A fault injected into
  one of its jq calls posted a "safe to update" headline over a body whose counts had rendered
  empty, and returned success. `set -e` does not catch this: a failing assignment inside a function
  running in a command substitution does not trip errexit. Composition is now fatal, transport is
  still not, and both directions are pinned by tests.
- A failing run recorded no coverage notes, so `--status` showed a leg whose patched binary was
  never started as a plain `pass`.
- `--platforms` was disclosed on the pass path but not the failure path, which announced "breaks
  `clodex patch` on 1 of 1 builds" with nothing saying the other seven were never looked at.
  Duplicate names are now rejected and completeness compares distinct platform names rather than row
  counts — naming one build eight times used to reach the published-platform count and earn a green
  "safe to update". Naming all eight no longer claims builds went untested when none did.

## Deliberately not changed

The green "safe to update" message for a fully covered run is byte-identical to `main`; the action
machinery stays entirely out of it.

## Tests

The verdict write, both notification bodies, the note collection, the subset note and the choice
between the two alerts move into named units, because the self-test previously carried its own
transcriptions of them — which is how the `*` was edited with every assertion still green. The tests
now drive what the script actually uses.

179 assertions. Verified by mutation, including reverting `*`, deleting the action line from the
posted message, disconnecting either composer from the posting, dropping the subset note from the
note list, reverting the fail-open coverage check, removing the composer guards, and tightening the
transport by mistake — each turns at least one case red.

Extracting the two message bodies was proven behaviour-preserving by rendering old and new and
diffing bytes: the pass body over three branch combinations, the failure body over six.

## Runtime evidence

This branch's canary was run against all eight live 2.1.268 builds, Docker up, isolated state
(`CANARY_STATE_DIR`) and Slack suppressed. All eight legs passed, both Linux ARM legs as real
containers, and it composed the correct green message — "ran on 3 of 8 builds", no action line, no
subset note. The real state directory was untouched.

Environment: macOS arm64, node v24.14.1, self-test green under bash 5.3.9 and bash 3.2.57.
`pnpm typecheck`, 2550 tests and `pnpm build` green.

## What I could not verify

Nothing automated runs the canary's top-level flow, so deleting a main-flow *call* to one of the
extracted units is still invisible to the suite. That is why the end-to-end run above was done.

ShellCheck goes from 51 findings to 65 at default severity, introducing no new codes: SC2016 on the
jq filter constants, which must not expand shell variables, and SC2034/SC1090/SC2329 on globals and
functions shared across the sourced files and the self-test. All classes are already present on
`main` for the same reasons.

## Docs

`canary/README.md`: the real patch runs on three of eight builds, not two, and the `container` leg
covers `linux-arm64-musl` as well as `linux-arm64`.

### DIFF
--- a/canary/README.md
+++ b/canary/README.md
@@ -22,9 +22,11 @@ each run, plus sandboxes — removed on success, kept on failure for the investi
 
 ## The three legs, and what each one actually proves
 
-Only two of the eight published builds can go through the real `clodex patch`: `clodex patch`
+Only three of the eight published builds can go through the real `clodex patch`: `clodex patch`
 resolves the version by **executing** the binary, so a foreign build needs either this Mac's own
-architecture or a native-architecture container. The other five get the probe,
+architecture or a native-architecture container. That is `darwin-arm64` on this Mac, plus
+`linux-arm64` and `linux-arm64-musl` in a container — the only two rows in `CANARY_PLATFORM_TABLE`
+carrying an image. The other five get the probe,
 `scripts/probe-patch-mechanism.mjs`, which works on any build from any host because it never
 executes anything.
 
@@ -46,7 +48,7 @@ A probe **does not** prove:
 
 - **that the patched binary runs.** Nothing here starts it, so a repack that produces a PE
   Windows refuses to load, or an ELF the kernel rejects, would still pass. Only the `host` and
-  `container` legs answer that, and only for macOS-arm64 and Linux-arm64.
+  `container` legs answer that, and only for `darwin-arm64`, `linux-arm64` and `linux-arm64-musl`.
 - **that an anchor bound to the right code.** A regex that matches a lookalike function still
   emits valid JavaScript and reports `OK`. The probe checks that a site *applied*, never that it
   applied to the thing it was aimed at.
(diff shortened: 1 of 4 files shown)


bman654:
Review record (self-approval is not possible as PR author; admin-merged past the branch-protection requirement).

Three adversarial review panels, three model families. Each found real defects:

1. A failing `--platforms` run disclosed no subset; tests did not exercise the message assembly at all.
2. Duplicate `--platforms` names could earn a green tick over seven untested builds; tests still admitted disconnected production wiring.
3. Two fail-open paths (a failed `jq` reading as full coverage; a renderer failure swallowed by the transport's `|| true`, posting a "safe to update" headline over empty counts), plus `--status` hiding lost coverage on failures and a false disclosure when all eight platforms are named explicitly.

Two of those were regressions introduced by this change. All resolved and pinned by tests; 179 self-test assertions green under bash 5.3.9 and 3.2.57, and verified end-to-end against all eight live 2.1.268 builds with Docker up.

PR #215, opened 2026-09-11
feat(oauth): keep every parallel subagent's ChatGPT chain alive

TweedBeetle:

When Claude Code runs several subagents at once on a ChatGPT/Codex model, most of them used to lose their conversation chain and re-send full context on every turn. Each subagent now keeps its own chain, so a parallel fan-out is cached about as well as its parts. Measured on five parallel Luna subagents: 64.9% cached input before, 82.0% after.

## Problem and root cause

Reachable by any ChatGPT/Codex OAuth user on Claude Code 2.1.268 who launches more than one subagent in the same message on a routed model.

2.1.268 marks every request from an in-process subagent with `x-claude-code-agent-id` and `x-claude-code-parent-agent-id` (verified in the bundle: the client factory adds both when `agentId` is set; the main agent sends neither). Every subagent still inherits its parent's session id, so a whole fan-out shares one socket partition. The possible-parent gate from #211 is deliberately conservative for a head that has committed nothing yet, and a fan-out hits exactly that case: five siblings start seconds apart, so siblings two to five each arrive while the first's initial response is still streaming. They are isolated (`isolatedByConnectionId: 1`), retain no head, and their next turns are isolated again by whichever sibling is still on its first response.

From the `ws_head_decision` log of one such run (post-#211 build): 1 `new_partition_head`, 10 `parallel_isolated`, 4 `history_mismatch_new_head`, 32 `continuation` across 47 Luna requests.

## The change

- `src/sdk-adapter.ts`: `extractClaudeAgentIds(headers)` reads the two headers (first value only, checked against `[A-Za-z0-9._-]{1,128}`).
- `src/proxy.ts`, `src/server/router.ts`: pass them into `withResponsesWebSocketDiagnosticContext` on all four call sites.
- `src/http-proxy/server.ts`: the MITM forwards both headers to the relay adapter alongside the session header (it previously forwarded an allowlist that dropped them).
- `src/oauth/responses-websocket.ts`: the context type gains `claudeAgentId` / `claudeParentAgentId`; `responsesWebSocketPartitionKey` takes the agent id as a fifth argument and appends it to the key material; the head decision's `keyTuple` records both ids.

`prompt_cache_key` is untouched, so siblings still share the server-side prefix cache; only the head lookup is per agent. The main agent and its auxiliary requests carry no agent id, so their partition and the isolation it relies on are unchanged.

Left out: a per-sibling `prompt_cache_key`. Tried as an experiment (each sibling's key suffixed with its agent id): 83.2% vs 82.0%, one run each, not distinguishable, and a staggered fan-out would lose the shared prefix on every sibling's first turn. Not included.

## Evidence

- [x] `pnpm typecheck && pnpm test && pnpm build`: 2561 of 2562 tests, 116 files, `CLODEX_HOME="$(mktemp -d)"`, node v22.23.2, no ambient proxy vars. The one failure, `credential-helper.test.ts > force-kills a helper that exceeds the runtime limit`, is a timing test untouched by this diff; it passes in isolation on this branch and on `main`.
- [x] Feature-deletion mutations, full files: dropping the agent id from the key material fails 2 tests in `tests/responses-websocket.test.ts`; dropping the context spread in `proxy.ts` fails the proxy plumbing test; dropping it in `router.ts` fails the API-server one.
- [x] Reachability stated above.
- [x] Manual: one Opus main + five parallel Luna subagents, each fixing a copy of a small Python project (all five green both times), through the proxy on Claude Code 2.1.268, isolated `CLODEX_HOME`, `--ws-diagnostics`. Before: 64.9% cached input over 47 Luna requests. After: 82.0% over 53, head decisions 5 `new_partition_head` + 48 `continuation` and nothing else. A three-sibling mixed run (Luna / a Go model / Haiku passthrough) still answers on all three. Anthropic passthrough is not touched by the diff.
- [x] Commit summary line reads as a release note.

Tests: the partition key discriminates on agent id and treats absent and empty alike; two siblings whose first responses overlap each get their own head and each continue their own chain with `previous_response_id`; the MITM test's adapter hop now asserts both forwarded headers; `tests/claude-agent-id-context.test.ts` reads the AsyncLocalStorage store from a mocked adapter and asserts what the proxy and the API server plumbed, and that both ids are absent for a main-agent request. That file uses a small test-only getter, `responsesWebSocketDiagnosticContextForTests`, added next to the existing `*ForTests` exports.

What the remaining gap to single-thread (92% on the same task) is: the five first turns are each a full miss (siblings two to five got ~1.8k cached tokens from the shared key, because all five prefills were in flight at once), and four real continuations came back with zero cached tokens from the server. The OpenAI guide notes that traffic above ~15 requests a minute to one cache key can overflow to another machine; this run sent 32 a minute on one key. That is the experiment above, and it did not help.

Not verified: behaviour on Claude Code builds that do not send the headers (the change is inert there by construction, but I only have 2.1.268 to hand).

A missing or malformed header means no agent id, which is today's behaviour. Nothing persisted. Each subagent now holds its own head, so a wide fan-out reaches the nursery cap (`maxNurseryConnections`, default 8) at eight concurrent siblings instead of sharing heads; that cap was already reachable before and is unchanged.

### DIFF
--- a/.claude/docs/oauth-continuation.md
+++ b/.claude/docs/oauth-continuation.md
@@ -48,24 +48,38 @@ arrival gate and the post-pacing re-check use the predicate, and the `ws_head_de
 
 ### Subagents partition by agent id
 
-Claude Code 2.1.268 sends `x-claude-code-agent-id` (and `x-claude-code-parent-agent-id`) on every
-request from an in-process subagent and neither on the main agent's. The relay reads them in
-`proxy.ts` / `router.ts` (the MITM forwards them to the relay adapter alongside the session header),
-carries them through `withResponsesWebSocketDiagnosticContext`, and `responsesWebSocketPartitionKey`
-appends the agent id to the key material. `prompt_cache_key` is untouched, so siblings still share
-the server-side prefix cache; only the head lookup is per agent.
+Claude Code sends `x-claude-code-agent-id` (and `x-claude-code-parent-agent-id`) on every request
+from an in-process subagent and neither on the main agent's — present in every bundle checked back
+to 2.1.238, so nothing here is gated on a client version. The relay reads them in `proxy.ts` /
+`router.ts` (the MITM forwards them to the relay adapter alongside the session header), carries them
+through `withResponsesWebSocketDiagnosticContext`, and `responsesWebSocketPartitionKey` appends the
+agent id to the key material. `prompt_cache_key` is untouched, so siblings still share the
+server-side prefix cache; only the head lookup is per agent.
 
-The possible-parent gate above is deliberately conservative for a head that has committed nothing
-yet, and a parallel fan-out hits exactly that case: five siblings start within seconds, so siblings
-two to five each arrive while the first's initial response is still streaming and were isolated
-(no head retained), and their next turns were isolated by whichever sibling was on its first response
-then. Measured on one Opus main + five Luna subagents fixing a small Python project in parallel
-(53 Luna requests): shared partition 64.9% cached input, 10 `parallel_isolated` + 4
-`history_mismatch_new_head`; per-agent partitions 82.0% cached, 5 `new_partition_head` + 48
-`continuation` and nothing else. The remaining gap to the single-thread figure (92%) is the five
-unavoidable first turns plus a few server-side cache misses on genuine continuations. Giving each
-sibling its own `prompt_cache_key` as well was tried and measured 83.2% — not distinguishable — so
-the key stays shared.
+**What this closes is the one shape the gate above cannot.** A sibling whose opening turn differs
+from the turn in flight already keeps its own head — that is the gate's job. A sibling whose opening
+turn is BYTE-IDENTICAL to the turn in flight is indistinguishable from a retry of that turn, so the
+gate must isolate it; the agent id is the only signal that tells the two apart. Measured on fresh
+single-purpose servers, one leg at a time, decision mix read from each server's own diagnostics:
+
+| siblings' opening turns | shared partition | per-agent partition |
+| --- | --- | --- |
+| distinct (5 x 4 turns) | 0 isolated, 5 sockets | 0 isolated, 5 sockets — identical |
+| identical (5 x 4, two trials) | 8 of 20 isolated, 13 sockets | 0, 5 sockets |
+| identical (8 x 3) | 8 of 24 isolated, 16 sockets | 0, 8 sockets |
+| identical, staggered 1.5s (4 x 3) | 3 of 12 isolated | 0 |
+| identical, headers stripped (control) | 8 of 20 | — (equals shared) |
+
+The control leg shows the difference is the header, not the build. The structural cost on a shared
+partition is N-1 extra full-context sends per same-prompt fan-out, more when first answers coincide
+or a late sibling's first response is still streaming. Per-agent partitions also remove a
+cross-sibling stitch the shared partition permits when visible histories coincide.
+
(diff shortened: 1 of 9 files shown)


bman654 (approved):
Merging this — thank you. It closes a real gap, and one the work we shipped in the meantime specifically cannot: the same-prompt fan-out. Since the only changes needed were a rebase and a rewrite of the rationale, not code, I've made them on your branch rather than send it back, and I want to explain exactly what changed and why.

### What happened while this was open

Two PRs landed on `main` that changed the ground under this one. #211 stopped isolating a request when the busy head's *committed* history diverged from it, and #219 extended that to a head that has committed nothing yet — the gate now compares an arriving request against `entry.current.originalPayload`, the conversation the busy head is generating right now. That is precisely the case this PR's description attributes the loss to ("siblings two to five each arrive while the first's initial response is still streaming"), so we measured whether anything was left for the agent id to do.

### What we measured

Fresh single-purpose servers, one leg at a time, decision mix read from each server's own `ws_head_decision` file rather than token percentages (which were noisy because other sessions shared the upstream account):

| siblings' opening turns | `main` | `main` + this PR |
| --- | --- | --- |
| **distinct** (5 × 4 turns) | 0 isolated, 5 sockets | 0 isolated, 5 sockets — identical |
| **identical** (5 × 4, two trials) | 8 of 20 isolated, 13 sockets | 0, 5 sockets |
| identical (8 × 3) | 8 of 24 isolated, 16 sockets | 0, 8 sockets |
| identical, staggered 1.5 s (4 × 3) | 3 of 12 | 0 |
| identical, headers stripped (control) | 8 of 20 | — (equals `main`) |

So for the case the description describes, `main` already does what this PR does — your 10-isolated baseline was measured before #219. But for siblings given a **byte-identical** opening turn, the gate on `main` *must* isolate them: a second copy of the turn in flight is indistinguishable from a retry of it, and the agent id is the only signal that tells them apart. The control leg shows the difference is the header, not the build. It also removes a cross-sibling stitch the shared partition permits when visible histories coincide.

**Calibration, which I've put in the doc so the value isn't oversold:** in a 27.6-hour local ledger, 0 of 3,995 isolations had a prefix-or-equal busy candidate — every one diverged at item 0. This shape didn't occur in that account's traffic at all. It's real for redundancy and consensus fan-outs; it isn't common.

### What I changed on the branch

- Rebased onto `main` — one conflict, both PRs inserting tests at the same point in `tests/responses-websocket.test.ts`; resolved keep-both.
- Rewrote the doc section and the `responsesWebSocketPartitionKey` comment around the residual above. The previous text described `main`'s behaviour in a way that is now false and was contradicted by the paragraph directly above it.
- Replaced the pre-#219 measurements (64.9% → 82.0%) with the table above.
- Dropped the "Claude Code 2.1.268 marks…" framing: both headers are present in every bundle we have back to 2.1.238, so nothing here is gated on a version, and it works on the installed 2.1.267.
- Fixed a line in our own doc that this PR made false.
- Summary line now says what it does: `keep prompt caching when several subagents are given the same prompt`.

Your code is untouched. Verified on the rebase: typecheck, build, 2570 tests green under an isolated `CLODEX_HOME`; mutations — dropping the agent id from the key material fails 2 tests, `extractClaudeAgentIds` returning `{}` fails 3, the MITM no longer forwarding fails 1. The credential-helper timing failure you mentioned did not reproduce.

Checked and not a problem: sockets equal or fewer, `prompt_cache_key` stays shared (verified in `keyTuple`), agent ids are minted once per subagent via `randomBytes(8)` and stable across every turn, resume reuses them, an absent header is inert.

Issue #225, opened 2026-09-11
Tool calls whose arguments are echoed with a different scalar type break the ChatGPT chain (Bash)


bman654:
## Symptom

On a ChatGPT/Codex OAuth model, a turn following a `Bash` tool call can lose its conversation chain
and re-send the whole history uncached, even though the tool call is plainly the same call.

## Evidence

The tool-argument canary already reports this. In a 27.6-hour local diagnostics ledger there are 254
requests whose head-decision diagnostic carries
`heads[].mismatch.toolArgumentNormalizationGap` — all with `equalAfterStrip: false`, meaning the
existing filler-strip rule cannot reconcile the two sides:

| tool | records |
| --- | --- |
| `Edit` | 177 |
| `Bash` | **77** |

The `Edit` ones have a known cause (the client echoes a tool call with its schema `default` filled in
— see #214). **The `Bash` ones do not:** the real `Bash` schema declares no defaults at all, across
`command`, `timeout`, `description`, `run_in_background` and `dangerouslyDisableSandbox`. So whatever
#214 does, these 77 survive it.

## Leading mechanism — scalar type coercion, not filler

A probe against a real Claude Code 2.1.267 binary with canned tool-call responses shows the client
rewriting the *type* of an argument before echoing it:

| model emitted | client echoed |
| --- | --- |
| `"timeout": "5000"` | `timeout: 5000` |
| `"run_in_background": "false"` | `run_in_background: false` |

The head snapshot holds the model's raw arguments, so a `function_call` with the same `call_id` and
`name` compares unequal on the coerced property, and the request falls to
`history_mismatch_new_head` with full context. That is exactly the `equalAfterStrip: false` signature
above.

**Not yet attributed.** The ledger stores hashes rather than content, so it cannot prove coercion
caused those specific 77 records — only that the mechanism is real and produces the same signature.
Confirming it means capturing a real `Bash` call whose model-emitted arguments carry a string where
the schema wants a number or boolean, then checking the next request's echo. GPT-family models do
emit stringified scalars for optional numeric parameters, which is what makes this reachable.

## Notes for whoever picks this up

- Do **not** fix it by coercing both sides generally. The comparison exists to notice a genuinely
  re-sent value, and `src/oauth/responses-websocket.ts` already documents that a changed value under
  the same `call_id` must stay a mismatch. A type-only reconciliation is narrower: same JSON scalar
  after coercing to the type the tool's schema declares for that property.
[...]

PR #242, opened 2026-09-16
fix(oauth): keep conversations cached after tool calls with number or true/false options

bman654:
---

On ChatGPT/Codex OAuth models, a turn after a tool call with number or true/false options — a
`Bash` call with a `timeout`, a `Read` with an `offset` — could quietly re-send the whole
conversation as a brand-new, uncached request. GPT-family models often write such an option as a
string, `"timeout": "5000"` or `"run_in_background": "false"`, and Claude Code hands the call back on
the next turn with the real types, `5000` and `false`. clodex saw a different call and started the
conversation over. Those turns now continue on the existing conversation like any other.

## The user-visible failure

Each time this fires, one turn is sent as a brand-new conversation with no cache — the full context
again — and the conversation then continues normally on the replacement chain. Nothing shows on the
terminal. A 36-hour `--ws-diagnostics` ledger from 2026-09-10/11 (clodex before #214, sol and luna)
holds 77 head-decision records with the canary's `toolArgumentNormalizationGap { tool: "Bash",
equalAfterStrip: false }` (the issue counted the same rows over the ledger's first 27.6 hours; none
were added after). They are **three incidents**: each first record is a `history_mismatch_new_head`
with zero cached tokens (82,318 / 66,609 / 62,645 input tokens), and the 74 later records are the
stale gap head being re-scanned on every following request while the conversation continued
normally on the replacement head (2-item deltas, 96–99% cache). Those incidents are *consistent
with* this mechanism, not measurements of it: the ledger stores hashes, and Bash's other ingest
rewrites (the `cd <cwd> &&` prefix strip, the `\\;` → `\;` rewrite) leave the same signature. #214 did not fix
this: `main` strips schema defaults only. The current 2.13.0 ledger (51 h) contains no such rows,
so the fix could not be re-measured live; its value scales with how often a model spells a scalar as
a string, which the hash-only ledger cannot tell.

Read out of the 2.1.273 bundle, executed with its own zod, and confirmed by running the real 2.1.267,
2.1.270 and 2.1.273 binaries against a synthetic Anthropic-format server
(`.claude/docs/claude-code-internals.md`, "Tool-call arguments are rewritten at ingest", has the line
citations and captured pairs). The probe: a local HTTP server answering `/v1/messages` with a canned
`tool_use` on the first turn and text afterwards; `ANTHROPIC_BASE_URL` pointed at it,
`ANTHROPIC_API_KEY` a dummy, isolated `CLAUDE_CONFIG_DIR`, proxy variables unset,
`claude -p probe --model claude-sonnet-4-5-20250929 --dangerously-skip-permissions` (and the same
with `--allowedTools Bash` / `Edit`, and under `acceptEdits`), then reading the `tool_use.input` the
client sent back in its second request. Canned `{"command":"ls","timeout":"5000","run_in_background":
"false"}` came back as `{"command":"ls","run_in_background":false,"timeout":5000}` on all three
versions and in every mode; `{"file_path":"/etc/hosts","offset":"5","limit":"10"}` came back as
`{…,"offset":5,"limit":"10"}`; with `CLAUDE_CODE_HUMBLE_HAMMOCK=1` the strings came back unchanged.

- Claude Code rewrites every `tool_use.input` against the tool's schema **when the assistant message
  arrives**, and the rewritten form is what it echoes. Two client rules: a generic repair that
  JSON-parses a string for any plain number/integer/boolean property (numbers kept only if they print
  back identically; booleans kept without that check), and per-tool cases that run Bash's `timeout`
  and its two booleans and Read's `offset` through their own coercers (trim + decimal literal, or
  exactly `"true"`/`"false"`), and fill TaskOutput's defaults. The generic repair skips properties
  declared through a zod preprocess, so PowerShell, Grep and CronCreate are never re-typed; Bash,
  Read (`offset` only), ToolSearch, Agent, TaskOutput and every MCP tool are.
- Bash's per-tool rewrite is all-or-nothing: one value its strict parse rejects (`"abc"`, `"False"`,
  `"0"`) or one unknown key, and the step is skipped — the transcript keeps the generic repair's
  output, scalar strings untouched and the unknown key kept (not necessarily the exact model input,
  since the generic repair may already have changed another field).
- This is **not** the permission path. Bypass mode, `--allowedTools` and `acceptEdits` echo the same
  rewritten arguments — and the same holds for the `replace_all` fill from #214, whose doc section
  claimed otherwise; that attribution is corrected here (#214's fix is unaffected).
- With `CLAUDE_CODE_HUMBLE_HAMMOCK=1` (or the matching remote flag, default off), **and** when the
  request builder's consistency gate accepts the raw input, the client sends the raw wire strings
  instead; the captured Bash pair passes that gate.
- On the wire clodex only sees `{type:"number"}` / `{type:"boolean"}`; the preprocess is invisible,
  so clodex cannot tell which client rule a property falls under.

Reachable by any ChatGPT/Codex OAuth user whose model stringifies a scalar option (sol and luna do)
on Bash, Read, ToolSearch, Agent, TaskOutput or an MCP tool with a number/integer/boolean parameter.
clodex's own translation layer was ruled out: the AI SDK's `jsonSchema(...)` wrapper carries no
validator, and `sanitizeToolInput` only drops `null`s and empty arrays.

## The change

Same design as #214, extended: the per-request map `toolSchemaDefaults(payload)` records, per
property, the declared default **and** the scalar kind, resolved the way the client resolves an MCP
schema (`type` string or array — a sibling `type` wins over `$ref`, and `{type:"number", enum}` is a
number — then a local `$ref`, then `anyOf`/`oneOf`; a union with string, an enum without a scalar
`type`, or an annotation-only property yields none). Two client quirks are mirrored rather than
corrected — refs are looked up in `$defs` whenever it exists, and resolution stops after 64 schema
visits — because either can only equate two spellings of one value. `stripSchemaDefaults` applies
both rules to BOTH sides of every comparison: re-type a string first, then drop a value equal to the
default — the client's own order, so `"false"` emitted for Edit's `default: false` comes back as
filler on both sides. A required property with a declared default is stripped like any other,
because the client fills those too (TaskOutput).

The re-typing rule lives in a new leaf module, `src/tool-input-coerce.ts`, and accepts the **union**
of what the two client rules accept, since clodex cannot tell them apart: a decimal literal after
trimming (`"5000"`, `"5000.0"`, `" 5000 "`, `"05"`, `"+5"`, `"-0"`, `"5.5"` even where the schema
says integer — Read echoes `offset:"5.5"` as `5.5`), a spelling that prints back identically
(`"1e+21"`), and `"true"`/`"false"` with surrounding JSON whitespace after at most one leading BOM.
`"1e3"`, `"0x10"`, `"False"`, `"0"`, `""` and anything non-string are left as written. **Precision
is the one place this fails closed:** a decimal literal is re-typed only when it has at most 15
significant digits (DBL_DIG), because `Number()` would fold `"9007199254740993"` onto
`"9007199254740992"` and make two different raw values compare equal. The client does re-type such
a literal, so that rare echo loses its chain — the safe direction. The exact safety claim is
therefore: distinct values with at most 15 significant digits never become equal, and longer
literals are not re-typed; the exponent path is injective because `String(n)` is unique per double.

Both sides rather than a snapshot rewrite, deliberately: under the wire-echo flag, or for a Bash
call that failed to re-type, the raw strings come back — normalizing each side keeps the chain
either way, while `"5000"` against `6000`, `"false"` against `true`, `"False"` against `false`, a
string-typed property echoed as a number, and `"9007199254740992"` against `"9007199254740993"` all
still start a new chain. The symmetry also equates a typed head with a string echo, which no client
path produces; that is harmless for the same reason #214's symmetric stripping is — upstream keeps
the model's own emission, the tool already ran, and tool outputs compare byte-exact (now pinned by a
test). The memo fingerprint covers the scalar kind, so a head scanned under a schema with the same
default but no kind is recomputed.

What this changes on the wire, precisely: the normalized argument values are compare-only — the
`arguments` string sent upstream and the input the client sees are byte-identical to before. What
the comparison *decides* does change: a turn that used to be sent as full history now goes out as
`previous_response_id` plus the delta. That is the fix.

Left out on purpose: stringified arrays/objects and annotation-only properties (the client parses
those too), Bash's `cd <cwd> &&` prefix strip and `\;` rewrite, Edit's `old_str` aliasing. Each is a
sibling shape of the same ingest normalizer and would show as `equalAfterStrip: false`.

## Tests and mutations

`tests/responses-websocket.test.ts` (all staged through `response.output_item.done`): the Bash repro
continues; the raw-string echo still continues; a changed `timeout`/`run_in_background` starts a new
chain; every Bash coercer spelling (`"5000.0"`, `" 5000 "`, `"05"`, `"+5"`, `"-0"`,
`"999999999999999"`) continues; `"9007199254740993"` echoed as `9007199254740992` starts a new
chain, as do `"9007199254740992"` vs `"9007199254740993"` under one `call_id` and the long literal
against an omitted default that rounds to the same double; the generic-repair spellings (`"1e+21"`,
`" true "`) continue on an Agent-shaped tool; an MCP `anyOf:[integer,null]` and a `$ref` scalar
continue while a `number|string` union does not; `"False"`, `"0"`, `"1e3"`, `"0x10"` start a new
chain; a string-typed `timeout` echoed as a number starts a new chain; Read `offset:"5"` +
`limit:"10"` echoes as `{offset:5, limit:"10"}` and continues, `"5.5"` → `5.5` continues, `"5"` →
`6` does not; `"false"` for `replace_all` is re-typed before it is judged filler against the client's
real `replace_all:false` echo; the memo fingerprint covers the kind; a TaskOutput call whose
required-on-the-wire defaults the client filled continues; a changed opaque function output
(`{"a":1,"b":2}` vs `{"b":2,"a":1}`) does not continue; the continued head's diagnostic carries no
gap; the canary still warns `equalAfterStrip: true` when only a `null` filler remains beside a
re-typed scalar. `tests/tool-input-coerce.test.ts` pins the accepted/rejected spelling table per
kind, the digit cap, the resolver rows (`type` arrays, `anyOf`/`oneOf`, `$ref`, both-bucket lookup,
sibling `type` over `$ref`, typed enum, cycles, the 64-visit cap), and that whitespace before the
BOM is not a boolean.

Each mutation below was applied to a snapshot of `src/`, both test files run in full, and `src/`
restored and diffed against the snapshot. Every one reds at least one test:

| mutation | tests red | first test that reds |
| --- | --- | --- |
| re-typing removed (identity) | 13 | re-types the boolean spellings either client rule accepts |
| re-type every string regardless of schema kind | 2 | continues for an MCP nullable scalar declared through anyOf or $ref |
| boolean spellings case-insensitive | 2 | leaves every boolean spelling neither client rule accepts |
| boolean regex narrowed to exact `true`/`false` | 2 | re-types the boolean spellings either client rule accepts |
| exponent-path print-back check removed | 3 | leaves every number spelling neither client rule accepts |
| integrality check re-added for `integer` | 2 | applies the same rule to an integer property, with no integrality check |
| process-global rule map, no keyed memo invalidation | 6 | does not let one client's tool schema decide another client's continuation |
| process-global map rebuilt per request, invalidation kept | 1 | does not re-type a property the tool schema declares as a string |
| default stripped before re-typing | 1 | re-types a string before judging it against the declared default |
| memo fingerprint ignores the scalar kind | 1 | recomputes a prefix memo cached under a schema with the same default but no scalar type |
| trim removed on the decimal path | 3 | re-types a decimal literal after trimming, like the Bash and Read coercer |
| scalar kind never derived from the schema | 9 | continues when the client echoed a string scalar re-typed to the schema number or boolean |
| `type` arrays ignored | 1 | resolves a `type` array with the client's preference |
| `anyOf`/`oneOf` resolution removed | 4 | resolves anyOf and oneOf branches under the same preference |
| `$ref` resolution removed | 4 | follows a local $ref into the tool's $defs or definitions |
| `parameters` not passed to the resolver (no `$defs`) | 1 | continues for an MCP nullable scalar declared through anyOf or $ref |
| decimal-literal pattern removed | 4 | leaves every number spelling neither client rule accepts |
| a `string` branch no longer vetoes the kind | 4 | resolves a `type` array with the client's preference |
| `function_call_output.output` compared as JSON | 1 | starts a new chain when an opaque JSON-looking function output changes bytes |
| required properties exempt from default stripping | 1 | continues a TaskOutput call whose required-on-the-wire defaults the client filled at ingest |
| 15-significant-digit cap removed | 2 | leaves a decimal literal with more than 15 significant digits alone |
| digit cap loosened to 17 | 2 | leaves a decimal literal with more than 15 significant digits alone |
| `definitions` preferred over `$defs` | 1 | mirrors the client's single-bucket lookup when both $defs and definitions exist |
| `$ref` resolved before a sibling `type` | 1 | follows a local $ref into the tool's $defs or definitions |
| `{type, enum}` declined | 1 | names the scalar kind a `type` string declares |
| whitespace accepted before the boolean BOM | 1 | leaves every boolean spelling neither client rule accepts |
| 64-visit cap lifted | 2 | survives a cyclic or deep $ref chain |

## Runtime evidence

- Real binaries against the synthetic server described above: 2.1.267, 2.1.270 and 2.1.273, in
  bypass mode, `--allowedTools` and `acceptEdits`, isolated `CLAUDE_CONFIG_DIR`, no proxy env.
- Gate: `pnpm typecheck && pnpm test && pnpm build` green — 119 files, 2717 tests — under Node
  v24.14.1, `CLODEX_HOME=$(mktemp -d)`, `CLAUDE_CODE_ENTRYPOINT=cli`, no `HTTP(S)_PROXY`.
- `origin/main` has not moved since the branch cut (`db43de6`); no open PR touches
  `responses-websocket.ts`, the coerce/sanitize modules or the two docs.

## Not verified

- No live ChatGPT-OAuth A/B: the current 2.13.0 ledger holds no function_call-vs-function_call
  divergence of any kind (5,392 of its 5,398 new-head decisions are `user` vs
  `function_call_output` at the first divergence at scan time — an unrelated pattern).
- The three ledger incidents record hashes only; their argument spellings, and whether scalar
  re-typing or another Bash rewrite caused them, are unknowable from the ledger.
- No pristine client path was observed emitting a genuinely changed value under an existing
  `call_id`; the changed-value negatives drive the comparison with that wire shape directly.
- Deferred-tool wire JSON (PowerShell, CronCreate, Monitor) is inferred from the same zod→JSON path
  captured for Bash, Read, Grep and ToolSearch.
- Annotation-only MCP properties are re-typed by the client and deliberately not here; a chain lost
  there is the pre-fix behaviour.

## Failure and rollback

The `arguments` sent upstream and the input the client sees are unchanged. A re-typing can only make
two spellings of one value (at most 15 significant digits) compare equal; a longer literal, an
unrecognised spelling or an unclassifiable schema is left as the model wrote it, so that call falls
back to a full-context send exactly as before — the pre-fix cost, never a wrong continuation.
Reverting restores the pre-#225 comparison with no persisted state to migrate.


bman654:

Three-lens panel plus two refuters and a cold reviewer, each in its own worktree, before the first push. What changed the design: the first draft re-typed only the intersection of the client's two coercion rules and missed real echoes (Bash `"5000.0"`, Read `offset:"5.5"`, MCP `anyOf:[integer,null]`); the union rule replaced it after the client's resolver was re-derived from the 2.1.273 bundle and executed with its own zod. Three "false match" BLOCKERs were refuted with proof: each needed a history the client cannot produce or a schema kind change between consecutive turns, and `main` already has the same property for #214's default stripping — one proposed fix ("never strip a required default") was shown to regress `main` and is now pinned by a guard test. The cold reviewer found the one real defect the panel missed: `Number()` collapsed literals beyond 2^53, which the ≤15-significant-digit gate now fails closed. 27 mutations, every one red; gate 119 files / 2,717 tests.

Issue #234, opened 2026-09-12
clodex-claude as a VS Code claudeProcessWrapper runs the extension's unpatched binary, so clodex models never appear in the picker


bman654:
## What happens

The Claude Code VS Code extension ships its own Claude Code binary (reported on Windows at `~/.vscode/extensions/anthropic.claude-code-<version>-win32-x64/resources/native-binary/claude.exe`). `clodex patch` patches the user's installed `claude`, not that copy, so the extension's model picker never lists clodex models.

Setting `claudeCode.claudeProcessWrapper` to `clodex-claude` does not fix it:

1. **It runs the wrong binary.** The extension invokes the wrapper as `<wrapper> <path-to-bundled-claude> <args...>`, and `src/claude-wrapper.ts` execs the binary it is handed when that path is an executable file. So routing works (the proxy env is injected), but the picker stays empty because the bundled binary is unpatched.
2. **On Windows it can't be spawned at all.** npm installs `clodex-claude` as `.cmd`/`.ps1` shims with no `.exe`, and the extension spawns the wrapper without a shell, so the `.cmd` fails with `spawn EINVAL`.

The workaround documented in `docs/windows-setup.md` (#208) is a hand-built Go shim that ignores the handed-in path and execs the patched npm binary.

Reported and diagnosed in #196 (Windows 11, Claude Code 2.1.267, extension 2.1.267: the bundled binary's sha256 matched clodex's pristine backup).

## Scope to decide

- Whether `clodex-claude` should prefer the clodex-patched install when the handed-in binary is a known-unpatched build of the same version, or whether `clodex patch` should also patch extension-bundled binaries.
- Whether to ship a Windows-spawnable launcher.
- Not yet verified: whether the macOS/Linux extensions bundle a binary the same way (likely, but no one here has checked). Also note the extension skips its own update check and moves permission-mode resolution out of the CLI when a wrapper is set.

bman654:
Fixed by #244 (macOS/Linux: the wrapper runs the clodex-patched install, so the picker shows clodex models) and #243 (Windows: `clodex install-vscode-launcher` builds a spawnable launcher). Enabling the patched-install substitution on Windows is tracked separately in the follow-up issue.

PR #243, opened 2026-09-16
feat(launch): add a Windows launcher so the VS Code extension can run Claude through clodex

bman654:
---

On Windows, the Claude Code VS Code extension could not be pointed at clodex: its
`claudeCode.claudeProcessWrapper` setting needs a real `.exe`, and npm installs `clodex-claude` only
as script shims, which the extension rejects with `spawn EINVAL`. The only way through was
hand-building a Go program from the docs. There is now a command for it: `clodex
install-vscode-launcher` builds a small `clodex-claude.exe` on your machine — with the C# compiler
that is already part of Windows, so nothing is downloaded — puts it under `~/.clodex\bin`, and prints
the one setting line to paste. After a window reload the extension launches Claude Code through
`clodex-claude`, the same wrapper background agents use, so sessions bridge to your running
`clodex server --proxy` without copying proxy values into VS Code settings. Re-run the command after
switching Node versions or moving clodex. Windows half of #234 (reported in #196).

**What this does not do:** it does not put clodex models in the extension's model picker on
Windows. `clodex-claude` (#244) substitutes your verified clodex-patched install for the bundled
binary on macOS and Linux only — `wrapperSubstitutionEligible` returns false on `win32` — so on
Windows the launcher's wrapper runs the bundled, unpatched `claude.exe` the extension hands it and
the picker stays as it was. The docs, the command's output and the help text say exactly that.

## The failure

The extension spawns the wrapper as `<wrapper> <bundled claude.exe> <args...>` with no shell and
`windowsHide: true`, from a setting that is a single executable string with no argument field
(verified in the 2.1.267 and 2.1.273 extension bundles). npm's three Windows shims for
`clodex-claude` — extensionless POSIX shell, `.cmd`, `.ps1` — are scripts; none can be spawned that
way. `docs/windows-setup.md` (#208) worked around it with a Go program the user compiles.

## The change

- **`launcher/clodex-claude-launcher.cs`** (shipped via `files`; deliberately not a `bin` — a
  bin-link would put a `.cmd` shim in front again). C# 5, `System` + kernel32 P/Invoke:
  - takes the raw command line from `GetCommandLineW`, drops only its first token (its own path,
    quoted or not, by the C runtime's argv[0] rule), and appends the rest verbatim to
    `"<node.exe>" "<dist\claude-wrapper.js>"`. Node parses it with the same C-runtime rules the
    extension's spawn wrote it for, so every argument value — JSON, `""`, trailing backslashes,
    Unicode — arrives as sent. Both paths are compiled in; if node.exe moved it falls back to
    `node.exe` on `PATH`; if the wrapper script is gone it exits 127 naming the command to re-run;
  - creates a job object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, starts Node `CREATE_SUSPENDED`,
    assigns it, resumes it, never closes the job handle — so when the extension kills the launcher
    (cancel, reload, exit) Node and the Claude Code grandchild die with it. Every Win32 return is
    checked: job create/configure/assign failure prints one `warning: … cancellation cleanup is not
    guaranteed` line and continues; a failed resume or wait terminates the child and exits 1 with
    the reason; the exit code comes from `GetExitCodeProcess` or the launcher exits 1 saying why;
  - returns true from its console control handler for Ctrl+C/Break so the child handles them.
- **`src/vscode-launcher.ts`**: finds `csc.exe` (`%WINDIR%\Microsoft.NET\Framework64\v4.*`, then
  `Framework\v4.*`, numeric-newest first; `SystemRoot`/`C:\Windows` fallbacks), renders the two
  paths as C# literals with every non-printable-ASCII unit escaped, writes the source with a UTF-8
  BOM to a scratch directory, runs csc there with relative paths and no shell, publishes by
  copy-beside-and-rename (the patcher's `publishFileByRename`, now exported). Every failure is a
  one-line result, never a stack: not Windows, no compiler, compile error (csc output shown),
  compiler exit 0 without output or non-zero leaving a partial file, compiler runner throwing,
  missing wrapper script, unusable temp directory, unwritable destination (with the "close VS Code"
  hint). Scratch removal retries and, if it still fails, is a warning that does not change the
  outcome. No settings file is read or written.
- **`src/claude-wrapper.ts` / `src/wrapper-env.ts`**: the wrapper's Windows spawn fallback used
  `shell: true` for every target, so arguments crossed `cmd.exe` (`%VAR%` expansion, quote handling)
  even for a native `.exe`. `wrapperSpawnShell` now returns true only for `.cmd`/`.bat` targets,
  which Node cannot spawn without a shell; the extension's bundled `claude.exe` and npm's
  `bin\claude.exe` are spawned directly. #244's selection block above the spawn site is untouched
  (rebased onto `97073ce`).
- **`src/cli.ts`**: `install-vscode-launcher` parsing, help text, root-help entry, dispatch.
- **CI**: new `windows-launcher` job on `windows-latest` (Node 24, `pnpm build`, then only
  `tests/vscode-launcher.windows.test.ts`).
- **Docs**: `docs/windows-setup.md` rewritten around the command and reconciled with #244 (Go
  listing removed, kept as a "previous workaround" note that names what it lacks; the Windows
  picker caveat above stated in the intro, step 2 and troubleshooting; #244's "Consequences"
  wording kept; the npm shim inventory corrected). README tip, `docs/background-agents.md`
  troubleshooting bullet, `.claude/docs/launch-and-wrapper.md` and `CLAUDE.md` updated; every
  "companion change / remains open" phrase from either side is gone.

## Tests

`tests/vscode-launcher.windows.test.ts` (runs on the Windows CI job; skipped elsewhere) drives the
built `dist/cli.js` so the runner's real Framework `csc.exe` compiles the shipped source, then spawns
the `.exe` the way the extension does (no shell, argv array) with `node.exe` + a probe script as the
"bundled claude": (1) install succeeds with empty stderr and prints the setting; (2) twelve hostile
argument values — spaces, embedded `"`, JSON, `""`, `%PATH%`, `&|<>^`, trailing `\`, Cyrillic +
emoji — arrive at the probe identical through node and the wrapper, stdin is relayed, cwd is
inherited, exit code 7 propagates; (3) exit 0 propagates; (4) after `child.kill()` on the launcher,
the probe and its grandchild are gone from `tasklist` within 10 s and no job-object warning was
printed; (5) re-installing over the running launcher ends in a one-line verdict (either outcome
Windows allows), and the launcher works afterwards; (6) a launcher baked against a copied package
whose `dist/claude-wrapper.js` is then deleted exits 127 with the exact re-run message. The
JSON/`%PATH%` assertions in (2) only hold on the wrapper's direct `.exe` spawn, so they pin
`wrapperSpawnShell` at its call site.

`tests/vscode-launcher.test.ts` (40 tests, any OS) drives the command against a fake `csc.exe` in a
fake `%WINDIR%` tree: discovery order and numeric version sort, rendering of hostile paths decoded
by an independent C#-literal parser, BOM and relative csc argv, publish and replacement as a new
file (hard-link oracle), scratch cleanup and the cleanup warning, every failure message, and a
tsup-built package layout run through the real `dist/cli.js` (platform forced to win32) so the
default `dist/claude-wrapper.js`/`launcher/` resolution is exercised with no injected paths.
`tests/wrapper-env.test.ts` has the `wrapperSpawnShell` table; `tests/cli.test.ts` covers the help.

## Mutations (src snapshot/restore, full-file runs)

| Mutation | Red |
| --- | --- |
| Baked path written unescaped | 9 |
| Publish by writing into the existing file instead of rename | 1 (hard-link oracle) |
| csc looked up only under 32-bit `Framework\` | 20 |
| Newest-v4 pick inverted | 2 |
| Non-Windows guard removed | 3 |
| Scratch directory not removed / cleanup warning swallowed | 2 / 2 |
| UTF-8 BOM dropped | 1 |
| Compiler exit status ignored (only output existence checked) | 1 |
| Build-phase throw left unwrapped | 1 |
| `defaultLauncherInputs` pointed at `missing-dist/` | 1 (packaged-layout test) |
| `wrapperSpawnShell` always true on Windows / platform check dropped | 4 / 1 |
| C# first-token stripping drops two tokens | 5 of 9 in the dotnet harness (macOS, manual) |
| C# `Main` body replaced with `return 0;` | can only red on the Windows CI job (tests 2–6 there; every probe assertion fails) |

## Runtime evidence (macOS, Node 24.14.1, `CLODEX_HOME=$(mktemp -d)`, no proxy env)

- `pnpm typecheck && pnpm test && pnpm build` on the rebased tree: 121 files passed + 1 skipped (the Windows file), 2813 tests.
- `node dist/cli.js install-vscode-launcher` on macOS: exit 1, one stderr line; `--help` and an
  unknown option behave like the other commands.
- The C# source compiles with zero warnings under the .NET 9 SDK at `LangVersion=5`, and its
  command-line splitting was executed there against the same hostile inputs.

The `windows-launcher` CI job has run on this PR: real csc compile, argument fidelity, kill
cascade, and re-install over a running launcher passed; the one failure (an em dash in a launcher
message mangled by the console code page) was fixed by making every launcher message pure ASCII.

## Needs a person with Windows and the extension

CI spawns the launcher from a test harness, not from VS Code. Three things remain manual; the exact
PowerShell and expected output are in `docs/windows-setup.md` → "What was verified, and where":

1. The real extension spawning the launcher: chat works; `Get-Process node, claude` before and after
   pressing stop shows the chat's processes (by `StartTime`) gone; the `Claude Code` output channel
   has no `From claude: clodex-claude.exe: warning:` line.
2. Windows Defender / SmartScreen: any prompt or quarantine on `~/.clodex\bin\clodex-claude.exe`.
3. Windows on ARM: whether `csc.exe` is found and the `anycpu` build runs.

## Failure and rollback

Removing the `claudeCode.claudeProcessWrapper` line and reloading returns the extension to its
default launch (or the env-var setup). Deleting `~/.clodex\bin\clodex-claude.exe` removes everything
the command created. A failed build never touches an existing launcher: the new file is renamed over
the old one only after csc succeeded.


bman654:
@bloodykheeng — you offered to test on Windows in #196, and this is the change that removes the hand-built Go shim. Everything CI can check on a Windows runner already passes here (the launcher compiles with the C# compiler that ships in Windows, arguments round-trip through it unchanged, and killing it takes down Node and Claude with it), so what's left is what only a real editor can show. If you have 10 minutes:

1. Install this branch: clone it, `pnpm install && pnpm build && npm link` (or wait for the next release if you'd rather), then run `clodex install-vscode-launcher` and paste its output.
2. Put the printed path into `claudeCode.claudeProcessWrapper`, reload the window, start `clodex server --proxy`, and open a chat. Paste anything the extension's output channel says, and whether the chat works.
3. Cancel a chat mid-response, then in PowerShell run `Get-Process node, claude -ErrorAction SilentlyContinue | Select-Object Id, ProcessName, StartTime` — we expect no leftovers from the cancelled chat.
4. Whether Defender or SmartScreen said anything about the built `clodex-claude.exe`.

One thing to set expectations: on Windows the extension's picker still shows the bundled binary's model list for now (substitution of the clodex-patched install is macOS/Linux-only in #244). Routing through clodex works; the picker follows once the substitution is verified on Windows.

The exact steps are also in `docs/windows-setup.md` on this branch. Thanks.

bman654:

Two-lens panel in separate worktrees before the first push, plus a real Windows run: the new `windows-launcher` CI job compiles the committed C# source with the runner's own .NET Framework `csc.exe`, spawns the built `.exe` the way the extension does (no shell), and checks that twelve hostile argument values (spaces, embedded quotes, JSON, empty string, `%PATH%`, `&|<>^`, trailing backslash, non-ASCII) reach the probe unchanged, that stdin/stdout/exit codes pass through, and that killing the launcher removes the Node child and its grandchild from `tasklist`. Replacing the launcher body with `return 0;` fails that job and nothing else, which is why the job exists. What changed from the review: every Win32 return is now checked (job-object setup failure warns and continues; resume/wait failure terminates the child), install-time filesystem failures produce a one-line result instead of a stack, the wrapper spawns a native `.exe` directly instead of through `cmd.exe`, and the docs say plainly that on Windows the picker still shows the bundled binary's list until substitution is verified there. Merged ahead of the requested manual test; that test still covers the real extension spawn, Defender/SmartScreen and ARM64, which no CI runner can.

bloodykheeng:
Hello @bman654, I've tested on Windows 11 / NVM for Windows (node v22.19.0) / clodex 2.14.0 from npm / Claude Code 2.1.276 / VS Code extension 2.1.276 / ChatGPT-plan OAuth.

**1. `clodex install-vscode-launcher`** — built cleanly, no prompts:

```
Built C:\Users\bk\.clodex\bin\clodex-claude.exe
  compiler: C:\WINDOWS\Microsoft.NET\Framework64\v4.0.30319\csc.exe
  runs:     C:\nvm4w\nodejs\node.exe C:\Users\bk\AppData\Local\nvm\v22.19.0\node_modules\@bman654\clodex\dist\claude-wrapper.js
```

**2. Chat works, and the picker now lists clodex models on Windows.** Set `claudeProcessWrapper` to the printed path, removed the three `HTTPS_PROXY`/`HTTP_PROXY`/`NODE_EXTRA_CA_CERTS` entries as the installer says, reloaded. Picker shows Terra and Luna alongside Sonnet/Fable/Opus/Haiku; picked Luna, got a reply (`[engine] turn 1 end (turns=1 usage in=21886 out=13 api=7114ms stop=end_turn)`). The "Claude VSCode" output channel showed only normal Claude debug output — no launcher or mismatch line.

Substitution is visible directly in the process tree. The extension hands the launcher its bundled binary, and what actually runs is the patched npm install:

```
50816 Code.exe
└─ 42336 clodex-claude.exe  …\.vscode\extensions\anthropic.claude-code-2.1.276-win32-x64\resources\native-binary\claude.exe --output-format stream-json …
   └─ 29276 node.exe        …\@bman654\clodex\dist\claude-wrapper.js  <same bundled path> …
      └─ 26476 claude.exe   …\node_modules\@anthropic-ai\claude-code\bin\claude.exe --output-format stream-json …
```

**3. Cancel mid-response** — sent a long prompt, hit Esc mid-stream. The session tab stays open (as expected) with exactly one launcher→node→claude chain and nothing extra. Full `Get-Process node, claude` afterwards: one `node` for `clodex server --proxy`, plus one chain per open chat tab. The only other `claude.exe` on the box traced back to a session in a *different* VS Code window I hadn't reloaded, still on my old hand-built shim — not a leftover from this launcher.

**4. Defender / SmartScreen** — nothing, at build or at first launch.

**One thing that will bite real users — and a concrete suggestion.** When I started, the extension had auto-updated itself to 2.1.276 while my npm Claude Code was still 2.1.267. In that state the launcher (correctly) declined to substitute and ran the bundled binary: chats worked, but the picker silently showed only the bundled model list. I had to `npm install -g @anthropic-ai/claude-code@2.1.276` and `clodex patch` (11 applied) before Terra/Luna appeared.

Why this matters more than a doc line: the extension updates itself from the marketplace whenever it likes, the npm install never does, and the only signal is one line in the "Claude VSCode" output channel that nobody has open. So every Windows user will eventually land in "clodex stopped working" with no error anywhere — and the fix is two commands they have no way to guess.

Two suggestions, either or both:

1. **Docs** — a "check the versions match" step *before* `install-vscode-launcher` in `windows-setup.md`, with the actual check:
   ```powershell
   (Get-ChildItem "$env:USERPROFILE\.vscode\extensions\anthropic.claude-code-*" -Directory | Sort-Object Name | Select-Object -Last 1).Name
   claude --version
   ```
   and the fix: `npm install -g @anthropic-ai/claude-code@<extension version>` then `clodex patch`.

2. **Code** — have `clodex patch` (and/or `clodex install-vscode-launcher`) read the extension's bundled version from the `~/.vscode/extensions/anthropic.claude-code-<ver>-win32-x64` folder name and, when it differs from the npm install, say so loudly with the exact command to run:
   ```
   VS Code extension is 2.1.276, npm Claude Code is 2.1.267 — the launcher will fall back to the bundled binary.
   Run: npm install -g @anthropic-ai/claude-code@2.1.276 && clodex patch
   ```
   A `--sync` flag that does it would be a bonus, but the warning alone turns a silent failure into a loud one, which is the real fix.

I'm deliberately *not* suggesting the launcher update npm itself — it runs on every spawn, has to stay fast and fail-open, and concurrent agent launches would race on the install. `clodex patch` is the one-time, user-triggered place where this check belongs.

Happy to retest on the next release.

bman654:
Thank you — this is exactly the test we couldn't run ourselves, and the process tree is the most
convincing evidence we could have asked for. It shows the substitution happening end to end: the
extension hands the launcher its bundled `claude.exe`, and what actually runs is the patched npm
install.

To close the loop on the four checks:

1. **Launcher build** — builds cleanly with the system `csc.exe`, no prompts. ✅
2. **Chat + picker** — works, and clodex models appear in the picker on Windows. ✅
   (My note above predates #246, which shipped Windows substitution in 2.14.0 — that's why you
   saw Terra and Luna rather than the bundled list I warned about. Sorry for the stale warning.)
3. **Cancel mid-response** — one chain per open tab and no orphans; the extra `claude.exe` traced to
   another window still on the old shim. ✅
4. **Defender / SmartScreen** — silent at build and first launch. ✅

ARM64 is the one thing still untested anywhere, if you ever have access to one.

**On the version skew — you're right, and it's worse than a doc line.** I checked: nothing in clodex
reads the extension's directory today, so there is no way for it to notice. I also think it isn't
Windows-specific. Substitution is gated on the bundled binary's hash matching the pristine copy
clodex recorded, so when the extension auto-updates and the npm install doesn't, macOS and Linux
fall back to the bundled list just as silently.

I agree with where you'd put it and with your reasons for keeping it out of the launcher — it runs
on every spawn, has to stay fast and fail-open, and concurrent launches would race on an install.
`clodex patch` is the right place: user-triggered, once, and already the thing you run after an
update. I've filed that as #257, with your framing and credit.

One small caution on the PowerShell in suggestion 1, so it doesn't end up in the docs as-is:
`Sort-Object Name` sorts version strings lexically, so `2.1.99` would sort *after* `2.1.276`. It
happens to work today because every current version is `2.1.2xx`, but it will pick the wrong
directory the first time the patch number crosses a digit boundary. The check needs a real version
sort — which is one more argument for doing it in `clodex patch`, where we can compare versions
properly.

Thanks again, and yes please on the retest once it lands.

bloodykheeng:
Thanks, good catch on the lexical sort, and agreed #257 is the right home for it. I'll leave it to you and retest once it lands. No ARM64 here unfortunately.

PR #244, opened 2026-09-17
fix(launch): show clodex models in the VS Code extension picker on macOS and Linux

bman654:
The Claude Code VS Code extension can now launch a verified clodex-patched install on macOS and Linux instead of always using its pristine bundled copy. That gives the extension access to the patched install's model list; the picker outcome was not exercised in a running editor. The companion Windows launcher is draft PR #243, while substitution remains macOS/Linux in this change.

## Setup

1. Set `claudeCode.claudeProcessWrapper` to the absolute path printed by `command -v clodex-claude` (or the documented stable launcher when a Node version manager owns that path).
2. Keep `clodex server --proxy` running so the wrapper can select the live proxy route.
3. Re-run `clodex patch` after either the extension or the local Claude Code CLI updates, then align their builds if the fingerprints differ.

## User-visible failure and mechanism

The extension ships and launches its own Claude Code executable. Proxy routing already made a saved clodex model work, but the bundled executable was unpatched, so its initialization model list omitted clodex models and aliases from the picker.

The mechanism chain verified from source is: the extension launches the selected Claude binary, that binary supplies its initialization model list, and the extension uses that list for the picker. This change makes the first link select the verified patched binary; the complete chain was not exercised in a running editor.

## Root cause and change

`clodex-claude` previously treated an existing first argument as authoritative and executed the extension's bundled file unchanged. It did not consult the patch manifest.

A pure eligibility helper now permits substitution only for POSIX, `CLAUDE_CODE_ENTRYPOINT=claude-vscode`, no child markers, and a live proxy-mode server. The wrapper then requires the full identity chain:

- the strict patch manifest contains the pristine and patched SHA-256 fingerprints;
- the handed-in executable's full SHA-256 equals the recorded pristine fingerprint; and
- the manifest target is executable and its full SHA-256 still equals the recorded patched fingerprint at the launch handoff.

Every refusal keeps the exact handed-in path and arguments. A manifest-backed mismatch emits one bounded synchronous stderr notice only for the persistent SDK `stream-json` chat spawn; auth, MCP, plugin, edit-hook, and nonpersistent suggestions helpers stay silent so their own errors remain first. Missing state without an identifiable manifest target stays silent, and stdout remains reserved for the Agent SDK protocol. A missing explicit VS Code path fails as that path instead of discovering another Claude install.

The manifest reader was separated from the patcher while preserving legacy-tolerant patch and restore behavior. Full-file hashing now uses bounded one-megabyte chunks. No launch-time writes, version subprocesses, directory scans, caches, watchers, automatic patching, or extension modifications were added.

## Scope and fallback behavior

Substitution does not run on Windows, in endpoint mode, without a server, for direct `clodex-claude` use, generic/background process wrappers, or `--check`. Same-file and already-patched inputs remain unchanged. Ordinary top-level extension helper commands are eligible for the same verified binary but do not print selection notices when refused.

If VS Code inherits `CLAUDECODE=1` from being launched inside Claude Code, its sanitized chat spawn remains eligible while helper commands re-merge the marker and retain the bundled executable; neither path runs an unverified binary.

Draft PR #243 is the companion Windows launcher. This change neither ships nor selects it, and the documented Go workaround bypasses the new full-hash guards. Remaining overlap will be reconciled when #243 rebases onto this change.

## Verification

- Selector tests cover exact provenance, same-size byte changes, strict/legacy/missing manifests (including a missing `patchedSha256`), unavailable targets, chunk-boundary hashing, and identity changes before and after handoff hashing.
- Eligibility table tests cover Windows, wrong entrypoint, each child marker, endpoint mode, and no server.
- Built-wrapper tests use a real patch-command fixture to prove executable identity, exact arguments, proxy environment, protocol-clean stdout, chat-only stderr delivery, silent refused helpers, closed-stderr survival, exit status, and POSIX process identity.
- Eight source mutations reddened the intended tests: force the handed-in target; skip the pristine proof; trust patched size without its full hash; drop the fallback notice; drop one argument; skip identity rechecks; remove the Windows exclusion; and emit a notice for a helper spawn.
- On macOS 26.6.2 with Node 24.14.1, a 200,000,000-byte file hashed in 107.1 ms on its first F_NOCACHE-assisted read; warm runs were 99.2, 97.2 and 97.4 ms (97.4 ms median). A successful eligible spawn reads roughly 400 MB; a mismatch still reads the roughly 200 MB handed-in file each time.
- Selection code grew the wrapper entry from 3.62 KiB to 11.79 KiB as `wrapper-target.ts` was inlined; importing `sha256File` through `patch-backup.ts` helped grow the shared chunk from 60.5 kB to 82.7 kB. No patcher, tweakcc, prompt, or new registry dependency entered the bundle.
- With an isolated `CLODEX_HOME`, `CLAUDE_CODE_ENTRYPOINT=cli`, and all four HTTP proxy variables unset, `pnpm typecheck`, the full test suite, and `pnpm build` pass.

## Evidence boundaries

The extension was not installed or run in an editor on this machine. Its wrapper contract and initialization-list-to-picker path were verified from downloaded macOS extension packages for Claude Code 2.1.267 and 2.1.273; the 2.1.273 bundled executable was byte-identical to the npm pristine artifact used by the installed patch. Linux follows the shared extension selector and POSIX wrapper path but was not run on a Linux host. No Windows launch or editor behavior was tested here; draft PR #243 is the companion launcher change.


bman654:

Two-lens panel in separate worktrees before the first push; no finding was contested, so no refutation round. Verified independently of the author: the substitution gate is the same three-condition test Claude Code 2.1.273 itself uses to recognise a top-level VS Code session, and in the extension's source only its environment builder sets that value while hooks, MCP servers, background wrappers, the generic Agent SDK and JetBrains are excluded; when substitution is refused the handed-in binary runs with argv and env identical to the pre-change build, one stderr line, nothing on stdout or disk, and it survives a closed stderr; the manifest reader extracted from the patcher is byte-identical to `main` (15-input differential) and the chunked `sha256File` matches the old digests across size boundaries; the test fixture drives the real patch command so the oracle is which fake process ran. What changed from the review: the eligibility check became a pure, table-tested function (removing the Windows exclusion was green before), and the fallback notice is limited to the chat process so it cannot prefix a failed helper command's error text. Hash cost measured at ~107 ms per 219 MB spawn, warm. Not exercised: a running editor's picker; Linux host.

PR #246, opened 2026-09-17
fix(launch): show clodex models in the VS Code extension picker on Windows

### LINKED ISSUE #245: Enable patched-install substitution for the VS Code extension on Windows
## Where things stand after #244 and #243

- #244 makes `clodex-claude` run the clodex-patched install instead of the VS Code extension's bundled binary when the bundled file's full SHA-256 matches the pristine hash in the patch manifest and the patched install still matches its recorded hash. That is what puts clodex models in the extension's picker.
- #243 gives Windows a spawnable `clodex-claude.exe` (`clodex install-vscode-launcher`) that runs the same wrapper, and a `windows-launcher` CI job that exercises it on a real Windows runner.
- The substitution gate, `wrapperSubstitutionEligible` in `src/wrapper-env.ts`, deliberately returns `false` on `win32`, so Windows users get clodex routing through the launcher but the picker still shows the bundled binary's model list.

## What this issue is for

Turn the gate on for Windows once the Windows CI job can prove the same guarantees #244 proved on macOS:

1. The selector's identity checks hold on NTFS: `dev`/`ino` from `statSync` (Node reports `ino` from the file index on NTFS — confirm it is stable across handles and not 0), `ctimeMs` semantics (Windows `ctime` is creation time, not change time, so the "rewritten with mtime restored" race in `tests/wrapper-target.test.ts` needs a different tell — probably size + `mtimeMs` + a re-hash, or `FileId` via `fs.statSync(..., {bigint:true})`), and `requireExecutable` (currently skipped on win32).
2. The fallback notice survives the extension's stderr handling on Windows the way it does on macOS (one line, before the child starts, chat spawn only).
3. Hash cost on Windows (Defender scans the file on open; measure cold and warm on the runner) is acceptable per spawn.
4. Extend `tests/vscode-launcher.windows.test.ts` (or a sibling) to build a fake patched install + manifest through the real patch command pattern in `tests/patcher-command.test.ts`, hand the launcher a pristine copy, and assert the patched fake process is the one that runs — the same oracle `tests/claude-wrapper.test.ts` uses on macOS.
5. Flip `wrapperSubstitutionEligible` for `win32`, update `docs/windows-setup.md`, the `install-vscode-launcher` success output and help text (all three currently say the Windows picker stays the bundled list), and the wrapper docs.

Manual confirmation that the picker actually shows clodex models in a Windows editor is still owed by a human (#196's reporter offered); CI cannot open VS Code.

## Not in scope

Patching the extension-bundled binary in place (option B in the #234 discussion) — it fights marketplace updates and depends on #230/#231/#232.

bman654:
On Windows, launching the Claude Code VS Code extension through `clodex-claude.exe` routed clodex models to your server but the extension's model picker still showed only the bundled binary's list, because `clodex-claude` deliberately never ran your clodex-patched install there. Windows users now get the same behaviour as macOS and Linux: when the extension hands the wrapper its bundled `claude.exe` and that file is byte-for-byte the pristine build `clodex patch` recorded, the wrapper runs your patched install instead, so clodex favorites and aliases appear in the picker. A mismatch (extension and CLI on different builds) still runs the bundled binary, and for the chat logs one line to the extension's Claude VSCode output channel.

Closes #245.

## What was missing

`wrapperSubstitutionEligible` (`src/wrapper-env.ts`) returned `false` on `win32`. It was held back because the selector's guard against the inspect-to-exec window relies on what `statSync` reports — `dev`/`ino`/`size`/`mode`/`mtime`/`ctime` — and each of those has a Windows-specific meaning that had never been measured, and because nothing could exercise the substitution on a Windows host.

## What changed

- **Gate**: `wrapperSubstitutionEligible(env, state)` no longer takes a platform. Every supported platform is eligible.
- **File identity** (`src/wrapper-target.ts`): read with `statSync(path, { bigint: true })` and compared as `dev`, `ino`, `size`, `mode`, `mtimeNs`, `ctimeNs`. NTFS file ids are 64-bit (16-bit MFT sequence over a 48-bit record number); past sequence 32 a double rounds them and two adjacent files compare equal, so the comparison no longer narrows. Times compare at nanosecond resolution rather than the rounded `*Ms` doubles.
- **Executable rule** (`requireWrapperExecutable`): POSIX still asks the kernel (`access(X_OK)`). Windows has no execute bit and libuv's `access` ignores `X_OK`, so the rule there is explicit: a `.exe` — the only shape the wrapper's Windows spawn starts directly. A `.cmd`/`.bat` (spawned through `cmd.exe`, naming any program) is refused on either side of the substitution, and the refusal reads as "could not be inspected".
- **Docs/help**: `docs/windows-setup.md` (intro, step 2, troubleshooting, verification), `docs/background-agents.md`, README, `clodex install-vscode-launcher` help and success output, `.claude/docs/launch-and-wrapper.md`, CLAUDE.md. Every sentence saying the Windows picker stays the bundled list is gone.
- **CI**: the `windows-launcher` job now also runs `tests/wrapper-substitution.windows.test.ts`.

Left out on purpose: patching the extension-bundled binary in place, and any caching of the selection decision.

## Windows proof (`tests/wrapper-substitution.windows.test.ts`, `windows-latest`)

Suite 1 records what Node reports about files on the runner's NTFS volume, as assertions, so a libuv change that breaks one is caught:

- `ino` is a non-zero bigint, identical across two `statSync` calls, an `fstatSync` on an open handle, and a hard link; different for another file. `dev` equals the volume serial `Win32_LogicalDisk` reports.
- An in-place rewrite of a file with the same size and `mtime` restored (integer seconds, so the FILETIME round-trips exactly) leaves `mtimeNs`, `size`, `ino` and `birthtimeNs` unchanged and moves `ctimeNs` — `ctime` on Windows is the NTFS change time, not creation time. The selector itself is then shown declining exactly that rewrite between `prepareWrapperTarget` and `finalizeWrapperTarget` (`input-changed-before-handoff`).
- `mode` changes when the read-only attribute is set; `accessSync(X_OK)` passes for a `.txt` file (why the `.exe` rule exists).
- SHA-256 of a 200 MB file (checked against an independently computed digest), timed twice and printed (warm only — the runner offers no cache eviction).

Measured on the first green run (`windows-latest`, C: is NTFS, Node v24.20.0, TEMP under an 8.3 path):

| Fact | Value |
| --- | --- |
| `ino` of a fresh file | `844424931515648` = MFT record 1383680, sequence 3; next file `…649`; below 2^53 on this volume, exact as a double — the bigint read matters only for volumes whose sequence numbers have climbed past 32, which the synthetic test in `wrapper-target.test.ts` covers |
| `dev` | `3606225537` = `0xd6f2a281`, the C: volume serial |
| in-place rewrite, `mtime` restored | `mtimeNs` unchanged at `1600000000000000000`; `ctimeNs` moved +80 ms (`1789610044350881800` → `1789610044431487200`); `birthtimeNs` unchanged |
| `mode` | `0o100666` writable → `0o100444` read-only |
| `accessSync(X_OK)` on a `.txt` | passes |
| SHA-256 of 209,715,200 bytes | run 1: 1180 ms on the first read after writing, 348 ms on the second; run 2 (same code): 177 ms and 183 ms |

The hash cost is the one per-spawn number worth a sentence: in steady state a launch pays a few hundred milliseconds to hash the bundled binary; one run showed a first read over the ~1 s mark this PR was asked to call out (first touch after the write, or a Defender scan on open), the next run did not. The selection is deliberately not cached, so the cost recurs on every eligible spawn; nothing here makes it cheaper than on macOS.

Suite 2 is the end-to-end oracle, the Windows counterpart of the tests in `tests/claude-wrapper.test.ts`:

- `clodex install-vscode-launcher` (the built `dist/cli.js`) compiles the launcher with the runner's own .NET Framework `csc.exe`.
- `tests/fixtures/fake-claude.cs` is compiled with the same compiler into a stand-in `claude.exe` that answers `--version` and reports which copy of itself ran by looking for clodex's `/*ccpatch:` marker in its own bytes; the fake JavaScript bundle is appended after a sentinel as PE overlay bytes.
- The real `runPatchCommand` patches that install with tweakcc's three calls mocked byte-for-byte (latin1, because the head is a PE image), producing a real manifest whose `pristineSha256` equals the pristine copy and whose `patchedSha256` equals the patched one.
- The launcher is spawned the way the extension spawns it (no shell, argv array, `CLAUDE_CODE_ENTRYPOINT=claude-vscode`, no child-session markers, a live proxy-mode record in the isolated `CLODEX_HOME`) with the pristine copy as the handed-in path, and:
  - the chat spawn (`--output-format stream-json`) runs the **patched** copy with its arguments (spaces, quotes, non-ASCII, empty) and the proxy env intact, stderr empty, exit code passed through;
  - a helper spawn (`auth status --json`) is substituted too, silently;
  - a same-size copy with different bytes runs as handed in with exactly one stderr line naming the handed-in path, the recorded install, the manifest and the reason; with both fds into one file, the line arrives through the launcher's inherited handles ahead of the binary's own output (the wrapper writes it synchronously before spawning — that ordering is code order, and this test observes the delivery rather than proving the order);
  - refused `--no-session-persistence` and `auth` helper spawns stay silent;
  - each spawn decides afresh: after a refused mismatch, restoring the pristine bytes substitutes again; closing the server disables substitution.
- Every test starts from the pristine copy (`beforeEach`); tests that need a mismatch make their own.

## Mutations (local, macOS, full-file runs)

| Mutation | Red |
| --- | --- |
| `sameIdentity` reduced to size only | 5 tests in `wrapper-target.test.ts` (during-hash and before-handoff races) |
| `sameIdentity` reduced to `mtimeNs` only | 2 tests: the during-hash replacements, which now carry the original's mtime (found by review; previously all 35 stayed green because the replacement got a fresh mtime) |
| Windows executable rule dropped (accept anything on win32) | 5 rows of `requireWrapperExecutable` |
| `ino` narrowed through `Number()` in `identityOf` | `compares file identity without rounding a 64-bit inode` |
| Notice emitted for helper spawns too | `claude-wrapper.test.ts` chat/helper notice test |
| Gate re-disabled on `win32` | only the Windows suite can see it — see below |

## Mutations run on the Windows job itself

Two throwaway commits were pushed to this branch and then dropped, so the runner could show the new suite discriminates:

| Mutation | Windows job result |
| --- | --- |
| gate re-disabled on `win32` | 5 failed / 15 passed: the chat-substitution, helper-substitution, same-size-mismatch notice, merged-output delivery, and decides-afresh tests ([run](https://github.com/bman654/clodex/actions/runs/35172399125/job/105046656374)) |
| `sameIdentity` reduced to size only **and** notice emitted for helper spawns | 3 failed / 17 passed: the selector's decline of the mtime-restored rewrite, and both refused-helper silence tests ([run](https://github.com/bman654/clodex/actions/runs/35172554561/job/105047130857)) |

## Review before push

An adversarial review (OpenAI model, different family from the author) of the staged diff reported before the first push. Its should-fix findings — an `afterAll` that could hang if setup failed early, tests sharing an undocumented mutated handed-in file, an ordering claim the merged-output test cannot actually discriminate, the output channel named differently from the verified **Claude VSCode**, "a mismatch logs a line" overstating what is chat-only, and a "measured" paragraph that outran the assertions — are all addressed in this revision; its hardening notes (locale-independent volume serial, digest check on the 200 MB hash, captured PowerShell status) are in too.

## Local gate

`CLODEX_HOME=$(mktemp -d) CLAUDE_CODE_ENTRYPOINT=cli`, no proxy variables, Node v24.14.1, macOS: `pnpm typecheck && pnpm test && pnpm build` — 121 files passed, 2 skipped (the two Windows suites), 2823 tests. The C# fixture compile-checks at `LangVersion=5` with the local .NET 9 SDK, and its report format was run under the `dotnet` host.

## Not verified

- A real VS Code extension spawning the launcher and showing clodex models in its picker on Windows — CI cannot open an editor; the reporter of #196 offered to try it.
- Cold-cache hash time: the runner has no way to evict the file cache; the 1180 ms first read is the closest available proxy.
- Defender/SmartScreen behaviour on the freshly compiled executables, and Windows on ARM.


bman654:

Two-lens panel in separate worktrees at the PR head, plus the Windows CI run as the Windows evidence. Verified independently of the author: the Windows end-to-end test is a real oracle — it builds `clodex-claude.exe` through the real install command, hands it a pristine copy of a `csc`-compiled fake `claude.exe` with the extension's environment and chat arguments, and the process that ran reports from its own bytes, where only the copy transformed by the real `runPatchCommand` carries the marker; the pushed "gate re-disabled" mutation failed exactly the five substitution tests with the happy path reporting the bundled copy. NTFS facts (`ino` stable, `dev` = volume serial, `ctime` moves on a same-size rewrite with `mtime` restored) are asserted, not printed. On macOS the bigint identity switch was checked for mixed-type comparisons and proven still to substitute end-to-end with the built wrapper. One change from the review: the during-hash replacement tests now preserve the replacement's `mtime`, so an identity check weakened to `mtime` only goes red instead of silently accepting an unverified file. Still unproven: a real Windows editor showing the picker, Defender/SmartScreen, ARM64, cold-cache hash cost.

PR #247, opened 2026-09-17
fix(patch): scope PATCH 6's presence test to the alias resolver

TweedBeetle:
## The bug

PATCH 6 skips an alias the bundle already resolves — deliberately, so a native
`case"sol":return "native";` wins rather than being shadowed by an injected duplicate. That
behaviour is pinned by `publishes built-ins when a local proof cannot be captured` and is
unchanged here.

The presence test for it read the **whole bundle**, and `case"<word>":return` is not a rare
string. zod's schema walker ships `case"union":return d.options...`, so an alias named `union`
read as natively resolved. Its case was never injected, `captureBuiltInPatchProofs` could then
not capture `case"union":return "union";`, and the entire **LOCAL PATCH SET** was abandoned:

```
  OK   PATCH 6: alias resolver switch
  ...
  FAIL LOCAL PATCH SET — could not capture built-in postconditions
clodex patch: 11 applied, 0 skipped, 1 failed
```

The patch still reports `Patched claude 2.1.274: 7 models, 7 aliases, 7 context windows`, so
the alias looks configured while `/model union` falls through to `default:return null` and every
local patch silently stops being applied.

## Reachability

Any user who saves an alias whose name collides with a `case"<name>":return` anywhere in the
bundle — no special configuration, no privileged access; it is decided entirely by the alias
NAME. Found while adding `union` for OpenCode Go's Union Alpha on Claude Code 2.1.274. `union`
is the instance I hit; the mechanism is not specific to it.

## The fix

Read the resolver switch alone, through the same anchor the injection uses, so the presence
test and the injection cannot disagree about which switch they mean. `RESOLVER_SWITCH` spans
`case"best":{…}` to the switch's own `default:return`. Cases *before* `case"best"` are the
reserved tier names an alias cannot take, so the region does not need to reach back past it.

`PATCH_TRANSFORMS_VERSION` 12 → 13: a site's regex changed materially.

## Evidence

New test `injects an alias whose name collides with an unrelated switch elsewhere in the
bundle` stages an unrelated `case"union":return d.options.map(walk);` in the fixture bundle,
configures a `union` alias, and asserts the alias reaches the resolver, the unrelated switch
is untouched, and the local patch set still runs.

**Delete the fix and it turns red** (revert the two changed lines to `.test(js)`):

```
 × runPatchCommand local patches > injects an alias whose name collides with an unrelated switch elsewhere in the bundle
      Tests  1 failed | 81 passed (82)
```

Restored: `Tests  82 passed (82)`. The digest pin is not the failing test — it is updated
separately and deliberately.

Full suite on this branch: `Tests 2824 passed | 20 skipped (2844)`, 0 failed. macOS 15.6
(darwin arm64), Node 22, `pnpm test`. Two tests in `http-proxy-index` and one in
`credential-helper` are timing-sensitive and have failed intermittently on this machine under
load in unrelated runs on `main` as well; they pass in isolation and passed in the run quoted
above.

Verified beyond the suite: with the fix, `clodex patch` on a real Claude Code 2.1.274 install
reports `12 applied, 0 skipped, 0 failed` with `LOCAL model-self-identity` applied, and a
`union` alias resolves live (`provider opencode-go`, `route translated`, HTTP 200).


bman654 (changes requested):
Thanks for this — the diagnosis is right, the bug is real, and the write-up is one of the better ones this repo has seen. I reproduced your failure on a real binary before reviewing the fix. I'm asking for one change to the new regex and one stronger fixture before it lands.

## What I confirmed

Ran the real `applyClodexPatches` + `captureBuiltInPatchProofs` over an extracted **2.1.274** bundle (your version) with `{'clodex:opencode-go:union-alpha': {alias: 'union'}}`:

| | `main` | this PR |
|---|---|---|
| PATCH 6 | `SKIP "already patched"` | `OK` |
| case injected | no | yes |
| proof capture | **throws** → local patch set abandoned | OK, 10 proofs |

The collision is exactly as you describe: `case"union":return` occurs **twice** in 2.1.274 and 2.1.276, both in zod-like schema walkers, neither in the model resolver. The anchor binds on all **57** bundles I have (2.1.208 → 2.1.276), matching an 85–88 char region against the 2000 budget, with exactly one `case"best":{` candidate per bundle.

Your comment's claim that the cases before `case"best"` are reserved names an alias cannot take is **exactly true** — I checked it rather than assuming: the five natives are `opus/sonnet/haiku/fable/opusplan`, all in `RESERVED_MODEL_ALIASES`, each rejected by the `fail()` at `patch-transforms.ts:215`.

## MAJOR — the fix reintroduces its own failure, re-keyed from alias *spelling* to alias *length*

PATCH 6 injects its cases immediately after `case"best":{...}` — **inside** the region `RESOLVER_SWITCH` measures. So the patch's own output spends the `{0,2000}` budget, at `2·len + 17` bytes per alias. Past ~2000 bytes the region stops matching, `resolver` falls back to `''`, and every alias re-reads as missing.

That is fatal rather than cosmetic because `builtInPatchVerificationFailed` (`patcher.ts:792`) re-runs `applyClodexPatches` over the already-patched content and demands a strict no-op:

```js
if (verification.content !== source || ...) return true;
...
} else if (result.status !== 'SKIP') return true;
```

On overflow PATCH 6 returns `OK` (not `SKIP`) and appends duplicates, tripping **both** conditions → `discardLocalPatchOutcome` → the user's entire local patch set is rolled back. Same symptom your PR exists to fix.

To be fair to the failure mode: it is not *silent*. `clodex patch` prints explicit `FAIL` lines, still publishes the built-in patches, and exits 0. What the user loses without a clear causal explanation is their local patch set — and nothing points at alias length as the reason.

Executed on the real bundle (pass 1, then pass 2 over pass 1's output):

| aliases × len | bytes | `main` | this PR |
|---|---|---|---|
| 20 × 20 | 1140 | idempotent | idempotent |
| 13 × 64 | 1885 | idempotent | idempotent |
| **20 × 42** | **2020** | idempotent | **duplicated → rollback** |
| **20 × 64** | **2900** | idempotent | **duplicated → rollback** |
| 40 × 42 | 4040 | idempotent | — |

`main` is idempotent at every size, so this is a regression the PR introduces.

**Reachability**, stated precisely: it needs local patches configured *and* >2000 bytes of alias text. Typical 3–12 char aliases spend ~820 bytes for twenty of them — nowhere near the cliff. But the worst case sits entirely inside documented limits: `MAX_MODEL_CATALOG` = 20 favorites × `MODEL_ALIAS_PATTERN`'s 64-char maximum = 2900 bytes, no hand-editing and no unsupported configuration required. `MAX_MODEL_CATALOG` is enforced in `favorites.ts`/`catalog.ts`/`routes.ts` and never referenced in `patcher.ts`, so there is no patch-time cap either.

**The ask.** My first instinct was `?? js` — fall back to main's whole-bundle search. A refuter killed it: in the degraded state that reintroduces exactly the foreign-switch collision this PR fixes (region misses, whole-bundle search finds zod's `case"union":return`, alias skipped, proof capture fails). Fail-safe in one direction, fail-open in the other.

What survived both probes is sizing the budget to account for what we inject:

```diff
-    const RESOLVER_SWITCH = /case"best":\{[^{}]*\}[\s\S]{0,2000}?default:return/;
+    const RESOLVER_BUDGET = 2000 + ALIASES.reduce((n, a) => n + 2 * a.length + 17, 0);
+    const RESOLVER_SWITCH = new RegExp(
+      'case"best":\\{[^{}]*\\}[\\s\\S]{0,' + RESOLVER_BUDGET + '}?default:return',
+    );
```

The 2000 stays as headroom for upstream drift; the added term is precisely the space our own cases occupy, so the patch can never crowd itself out. I verified it: 20×42, 20×64, 40×64 and 100×60 all return to `SKIP`/idempotent, and `tests/patcher-command.test.ts` + `tests/patcher.test.ts` give **232 passed** with the only red being the sha256 digest pin. Your `union` test stays green.

It also carries no over-reach risk, which is the thing worth checking before enlarging any bound. The quantifier is lazy, so the region always ends at the *first* `default:return`; on 2.1.276 the matched region is 85 chars at budgets of 2000, 4900, 13700 and 100000 alike. A bigger bound can only turn a no-match into a match — it can never move the endpoint.

## MINOR — the region's backward edge is behaviourally unpinned

Widening the region backward to swallow every native case, leaving the injection anchor untouched:

```diff
-    const RESOLVER_SWITCH = /case"best":\{[^{}]*\}[\s\S]{0,2000}?default:return/;
+    const RESOLVER_SWITCH = /switch\([\w$]+\)\{[\s\S]{0,2000}?default:return/;
```

→ `Tests 1 failed | 232 passed`, and the single failure is the source-digest pin. Every behavioural test passes, including your new `union` test. Our own rule says if the digest pin is the only red, the behaviour is unpinned — and the backward edge is what this fix turns on.

The fixture can't catch it by construction: its resolver is `function rz(x){switch(x){case"best":{return "opus"}default:return null}}`, with no case before `case"best"` at all, and "a fixture with no earlier candidate can only match correctly". Your decoy is appended at the *end*, which pins the forward/global edge — the `union` collision — and not the backward one.

Please add the decoy we ask for by default: one competing `case"<alias>":return ...` immediately **before** `case"best"` in the fixture resolver. It passes at your head and turns red on the widened mutant. This is test strength, not shipped behaviour — on today's bundles backward widening is harmless precisely because all five natives are reserved. It matters the day upstream adds a non-reserved case before `case"best"`.

## Checked and dismissed

- **Lazy match binding a foreign switch.** Real bundles put the next `default:return` 3128–5758 chars away, all outside the bound. Not reachable.
- **Anchor decoys.** `RESOLVER_ANCHOR` is byte-identical to `main`; the injection point didn't move.
- **Upstream adding `case"sol"` before `case"best"`.** The injected duplicate is dead code — the first label wins — and `main` is the side that loses the built-in postcondition there. This PR is an improvement.
- **`PATCH_TRANSFORMS_VERSION` 12 → 13.** Warranted; a site's regex changed materially, and the upgrade path is safe.
- **Your suite numbers.** `2824 passed | 20 skipped` reproduced exactly, along with your deletion-mutation result.

## One prose note

The retained comment justifies the presence test as stopping a native `case"sol":return "native";` from being shadowed by an injected duplicate. Under the new region that skip can no longer fire on a pristine bundle — every native case sits before `case"best"`, so the region contains zero `case"X":return`. The test's real remaining job is **re-patch idempotency**, which is exactly what the MAJOR above is about. Worth saying so in the comment, so the next person to touch this knows which property they must not break.

TweedBeetle:
Thanks for the review. Both changes are in b67763d, rebased onto main (which now includes #250).

**MAJOR (the budget).** I took your diff as written: the bound is `2000 + Σ(2·len + 17)` over the configured aliases, built with `new RegExp`. The new test `keeps a re-patch idempotent at the documented alias maximum` configures 20 favorites with 64-char aliases, runs `runPatchCommand` with a local patch set, and asserts each case appears exactly once, the local marker is present and no `FAIL LOCAL PATCH SET` is logged. With the budget reverted to a fixed 2000 it fails (the rollback you described), and at head it passes.

**MINOR (the backward edge).** The new test `reads the resolver region from case"best" forward, not the cases before it` puts `case"orbit":return "decoy";` immediately before `case"best"` and configures the alias `orbit`. Your widened mutant (`switch\([\w$]+\)\{[\s\S]{0,…}?default:return`) turns it red, and head passes.

**Comment.** The presence-test comment now says its remaining job is re-patch idempotency, and notes that the native-case protection can only fire for a non-reserved case after `case"best"`, which no current build ships.

`PATCH_TRANSFORMS_VERSION` stays at 13, since this PR's bump hasn't been released yet. The digest is re-pinned.

Environment: macOS 26.6.2 (arm64), Node 22.23.2, pnpm 10.34.5. Full suite: `2826 passed | 20 skipped`.

I also ran the #250 harness over the nine bundles I have locally (2.1.268 to 2.1.276). All nine pass `patches once per alias, inside the resolver, and is idempotent`. Two tests fail, both as the harness expects: `has a corpus` (9 bundles, not 50+), and LENS 2, which asserts the re-run output differs and now gets identical output. That is the flip the harness header predicts once the budget is dynamic, so its two pinned constants will want re-pointing when this lands. LENS 1's drift cases (`default:{return`, and upstream inserting a case with a body over the 2000 headroom) still duplicate on re-patch. They are unchanged by this revision, and I didn't try to cover them, given the `?? js` result.

bman654 (approved):
**APPROVE.** Head `b67763d` resolves both prior review asks, and I confirmed the result independently rather than relying on the updated fixture alone.

I reran the real `applyClodexPatches` and proof capture over all 57 cached Claude Code bundles spanning 2.1.208 through 2.1.276, with both a small alias set and the documented maximum of 20 64-character aliases. At both sizes, the anchor and region bound once at the same resolver in every bundle; every alias was injected exactly once; and proof capture succeeded. The small configuration's second pass was byte-identical and PATCH 6 reported `SKIP`; at the maximum configuration, both the second and third passes were byte-identical and the second reported `SKIP`. That corpus run passed 173/173 tests. I also confirmed that reverting to the fixed 2000-byte bound breaks the maximum-alias behavior, while widening the backward edge breaks the new decoy test. CI is green.

## BLOCKING

None.

## MINOR / NIT — maintainer-owned follow-up, not a contributor ask

1. **Two intentional safety properties still deserve behavioral pins.** In the two relevant test files, changing the fail-closed `?? ''` to the rejected `?? js` fallback leaves every behavioral test green; only the source-digest pin turns red. A resolver-drift fixture with an unrelated `union` switch then reports PATCH 6 as already patched and drops the alias. Separately, halving the per-alias coefficient passes the current maximum-alias fixture but no longer preserves the stated 2000 bytes of upstream-drift headroom: on the real 2.1.276 bundle with a syntactically valid 1900-byte comment before the resolver default plus 20 × 64-character aliases, head remains idempotent while the halved mutant reports `OK` on the second pass and duplicates the cases.

   Your update comment already calls out the unresolved resolver-drift boundary and says you did not try to cover it given the `?? js` result. Neither synthetic drift exists in the 57-bundle corpus, and the shipped formula is exact, so I am not asking you for another revision. These are maintainer-side test-strength follow-ups, not current defects.

## Checked and demoted

- The larger dynamic budget does not change the matched span on any cached bundle, even at the maximum legal alias configuration.
- **The comment's stated reason for the bound being safe is inaccurate, though the bound *is* safe.**
  `patch-transforms.ts` says a larger bound "can only turn a miss into a match; it never moves where
  the region ends." That holds for a single-anchor bundle, but not in general. With two `case"best":{...}`
  anchors — an early one whose `default:return` is far away, and the real resolver later — enlarging the
  budget moves the *start*, and the endpoint moves with it:

  ```
  budget=2000  start=2720  end=2778   region contains case"sol":return -> true
  budget=4900  start=120   end=2655   region contains case"sol":return -> false
  ```

  Both are successful matches, the endpoint moved, and the alias dropped out of the region — the exact
  failure this PR fixes. What actually makes it safe is not laziness: the region's prefix *is*
  `RESOLVER_ANCHOR`, so two viable starts mean two anchor matches, and `applyOnce` `fail()`s on
  `count > 1`. Any bundle that could move the endpoint aborts the patch first, and all 57 cached bundles
  carry exactly one candidate.

  **This one is ours, not yours** — the sentence came from the previous review comment and you copied it
  in good faith. We will reword it to name the load-bearing property, so a future maintainer does not
  relax the ambiguity check believing laziness alone is sufficient. No action needed from you.
- Falling back to the whole bundle on a region miss restores the unrelated-switch collision this PR fixes; keeping the fail-closed fallback is the right choice.
- `PATCH_TRANSFORMS_VERSION` 13 is correct: version 12 shipped in 2.14.0, while 13 is the single unreleased bump for this PR.

## Merge instructions

This is standalone, current with `main`, and no other open PR touches these files. Squash it with an empty body and this user-facing subject:

`fix(models): stop a custom model alias from disappearing when its name matches other code (#247)`

After the squash, I will re-point `.claude/harnesses/pr247-patch6-resolver-region-over-real-bundles.harness.ts`: derive its copied region matcher, `prMissing` helper, and drift filler from the dynamic budget; invert the deliberately pre-fix LENS 2 expectations to require an idempotent re-run; qualify the foreign-margin check for resolver drift instead of preserving the obsolete fixed-2000 invariant; and re-stamp the header. That harness fallout is ours and should not hold the contributor's fix.

PR #248, opened 2026-09-17
feat(opencode-go): add union-alpha on the Anthropic Messages transport

TweedBeetle:
## What this adds

`union-alpha` — Union Alpha on OpenCode Go. It shows up in the updater's own
"present on models.dev but not transport-mapped" report; this maps it after verifying the wire
protocol, as that report asks.

- Free: `cost` 0 input and 0 output, no cache pricing published.
- 262,144 context / 131,072 output. Text + image in, text out. Tool calling.
- `reasoning: true` with `reasoning_options: []` — the feed publishes no effort ladder, and the
  Messages route could not send one anyway, so it takes the standard anthropic-format
  `supportsReasoningEffort: false` / `supportsCountTokens: false` pair.

It is a **cloaked model**: models.dev publishes no vendor for it, and asked directly it answers
"made by a currently anonymous company". Flagged here because "free" and "undisclosed operator"
travel together and are worth a user's attention, not because it changes anything mechanically.

## How the transport was determined

One request per endpoint against the live Go gateway, 2026-09-17, key from an OpenCode Go
subscription, `x-api-key` on `/v1/messages` and Bearer on the other two:

| endpoint | result |
|---|---|
| `/v1/messages` | **200**, `stop_reason: end_turn`, usage reported |
| `/v1/chat/completions` | 500 `Internal server error` |
| `/v1/responses` | 500 `Internal server error` |

That agrees with models.dev, which publishes `provider.npm: "@ai-sdk/anthropic"` for this entry.
No other `TRANSPORTS` entry was touched; re-running the updater produced an additive diff only,
with no metadata drift on existing models.

## Evidence

`Tests 2823 passed | 20 skipped (2843)`, 0 failed. macOS 15.6 (darwin arm64), Node 22.

Delete the catalog entry and the tripwire turns red:

```
 × OpenCode Go catalog > records its OpenCode catalog source and excludes Responses-only models
      Tests  1 failed | 30 passed (31)
```

Restored: `Tests 31 passed (31)`. The two literal counts in that test were updated on purpose,
per its own comment, not loosened.

Verified beyond the suite: with the entry present, a real Claude Code 2.1.274 session aliased to
it answered through the proxy — `provider opencode-go`, `route translated`, HTTP 200 — ran the
Bash tool and returned a correct answer in 23 s.

## Note

Adding this alias is also what surfaced #247: `union` collides with zod's own
`case"union":return ...` in the Claude Code bundle, which made PATCH 6 drop the alias and
abandon the local patch set. This PR is independent of that one, but a user who aliases this
model `union` will want both.


bman654 (commented):
Thanks — the method here is right, and I want to be clear up front that what follows is **not a criticism of your work**. You measured the transport against the live endpoint, recorded what you observed, and said which parts you couldn't verify. That is exactly how a provider mapping should be established.

The problem is that the model stopped existing while the PR was open.

## Blocker: models.dev has delisted `union-alpha`

```
$ curl -s https://models.dev/api.json | grep -c union-alpha
0
```

models.dev added the entry on 2026-09-16T14:44Z and removed it on 2026-09-17T23:55:46Z — commit subject `chore: remove expired Union Alpha model`. You fetched at **2026-09-17T20:50:29Z**, inside that ~33-hour window, so every number in your PR body was accurate when you took it. The vendor pulled the model roughly three hours after you opened this.

The consequence is not cosmetic, because the updater fails closed on a transport-mapped id the feed no longer publishes:

```
# this branch
$ node scripts/update-opencode-go-models.mjs
Transport-mapped models missing from models.dev: union-alpha
exit 1          (wrote nothing)

# main, same moment
$ node scripts/update-opencode-go-models.mjs
Updated 18 OpenCode Go models from https://models.dev/api.json (opencode-go).
exit 0
```

So merging this would block every future catalog refresh for **all 18 other OpenCode Go models** until someone deletes the entry again.

Runtime degradation, separately, is benign: `opencode-go` is `staticModelPolicy: 'allowlist'`, so the catalog is intersected with the live model list and a dead entry goes inert rather than corrupting anything. The blocker is the updater, not the runtime.

## What we could not verify

We have no OpenCode Go subscription on the maintainer side, so we **cannot** test whether `/v1/messages` still serves `union-alpha` now that it is delisted. Your 200 predates the expiry. The provider's `/v1/models` does still advertise the id, but that list demonstrably retains retired ids (`deepseek-flash` was removed from models.dev on 09-10 as deprecated and is still listed), so its presence there isn't evidence the model answers. Saying this plainly rather than guessing: we don't know, and you're better placed to find out than we are.

One calibration note, because it will matter next time rather than now. The PR says the transport choice agrees with models.dev, "which publishes `provider.npm: \"@ai-sdk/anthropic\"` for this entry." I don't think that holds: models.dev publishes `npm` **per provider, not per model**, and for `opencode-go` it is `@ai-sdk/openai-compatible` — individual models carry no `npm` at all. The `"@ai-sdk/anthropic"` you saw is almost certainly clodex's own generated field: `scripts/update-opencode-go-models.mjs:340` writes `npm: anthropic ? '@ai-sdk/anthropic' : '@ai-sdk/openai-compatible'` from the `TRANSPORTS` map, and line 13 of that file marks `npm` as local-only.

So that particular corroboration is circular — it's our output, not the feed's input. Your actual evidence, the 200/500 table, is unaffected and is the right way to establish this. I'm flagging it only so the reasoning doesn't get reused as independent confirmation later. Our own file is what made it look like feed data, so this one is on us.

## Closing this one

I'm closing it — the model it adds no longer exists, so there is nothing for it to enable, and leaving it open would only invite a rebase onto a dead id.

To be explicit that this is not a judgement on your work: the delisting is entirely vendor-side, your measurements were correct when taken, and the entry itself came through the review clean.

The more useful outcome is #249, which this PR is what prompted. The argument there is that **the PR pipeline is slower than this class of model's lifetime** — `union-alpha` lived about 33 hours, and no review process fixes that. Instead of mapping cloaked models one PR at a time, users should be able to add an OpenCode Go model themselves, with clodex probing the transport live against their own credentials at the moment they add it. That is the same thing you did by hand for this PR, just done by the tool, at the time it actually matters, without waiting on a release.

If you'd like to take that on, it's yours — you've already done the hardest part of the research.

## A gap this surfaced (now #249)

The panel mutated the entry to see what the suite would catch. It caught less than I expected:

- Changing the transport in `scripts/update-opencode-go-models.mjs` from `anthropic-messages` to `openai-completions` leaves **all 2,823 tests green**.
- So does a wrong `contextWindow`, a wrong `upstreamModelId`, or an arbitrary replacement id.
- `modelFormat` and `compatibility` corruptions are caught only by generic invariants and the literal counts.

So the one fact that only you could establish — the transport, measured by hand against a live endpoint — is also the one fact this repo has **no automated defence for**. That isn't a flaw in your PR; the count bumps you made were the correct and honest thing to do with the test shape that exists. But it means every OpenCode Go mapping rests entirely on the contributor's measurement, and a silent transport error would ship green. Both of these are now captured in #249, along with the design constraint that makes them awkward: discovery returns only the model id, so the curated list is carrying nearly all the metadata, not just the transport.

Small thing while it's open: the comment in `tests/opencode-go.test.ts` still says `4/13` after the counts moved.

## Worth recording: everything else checked out

None of this was wasted. The panel traced the whole entry and found no defects:

- **Entry shape** — all 19 catalog entries carry the identical 12 keys; yours omits nothing. The only difference is `cost.cache_read`, correctly absent (the updater drops falsy cache pricing, and `isZeroOrMissing` treats 0 and undefined alike).
- **Zero cost** — `src/free-models.ts` uses strict `=== 0` / `> 0` / `??` throughout. No `||` fallback, no truthiness test, no division, no price formatting anywhere in `src/`, and `preserveModelPricing: true` stops the pricing cache overwriting `{0,0}`. Resolves to `verified_free`. This was my main worry and it's clean.
- **`reasoning: true` + `supportsReasoningEffort: false`** — yields `mode: 'internal-only'`, so the patcher emits no effort ladder and nothing reaches the wire. The pair is load-bearing: without it the `@ai-sdk/anthropic` branch would advertise the Anthropic thinking ladder on a passthrough route. You got this right.
- **`family: "union"`** — reaches only fuzzy search, display, and `deriveBrand` (→ `'Other'`, same as `hy3`). Never enters the patcher, patch-transforms, or the proxy, and feeds no regex or switch. Note the asymmetry with #247: that break was the *alias* string on the patcher path, a different value entirely.
- **Additive-diff claim** — verified against today's feed by de-mapping only `union-alpha` and re-running: exactly the 23-line removal, zero drift on the other 18.

Two nits for whenever a mapping does land: `union-alpha` sits out of alphabetical order in `TRANSPORTS` (index 14, belongs at 18); and an absent feed `cost` would mint `{0,0}` and read as "verified free", which is a pre-existing updater hazard rather than anything you introduced.
On the cloaked-operator point: you flagged it in the PR body and nowhere else, and I think that was the right call for a *code* change — but it does mean a free model from an undisclosed operator would appear in the picker looking like any other. That's a docs note, not a gate, and not something to hold a PR for.

PR #252, opened 2026-09-18
fix(translate): give translated messages ids Claude Code won't anchor a thread on

TweedBeetle:
On current Claude Code builds, a translated model that keeps no conversation state on its side could fail every other request after a tool call. Claude Code would then resend the whole conversation, so the session carried on, but each failure cost a round trip and showed up as a 400 in the logs. clodex now labels the replies it translates so that Claude Code always sends the full conversation.

## The failure

Observed on Muse Spark 1.3 Contributor through OpenCode Go (Responses transport, a local fork route) with Claude Code 2.1.274: after each tool call, the next request came back 400:

```
Error from provider (Console Go): Upstream request failed: [invalid_request_error]
No function call found for function call output with call_id 'call_01a0b13d…'
```

and Claude Code retried it about 700 ms later with the full history, which succeeded. In the proxy's request log on my machine, 42 of 245 Muse requests were these 400s.

## Root cause

Claude Code anchors server-side thread continuation on the previous assistant message. This is the gate in the 2.1.276 bundle:

```js
var xQo="msg_";function qst(e){if(e.isApiErrorMessage)return!1;if(e.requestId)return!0;if(e.isVirtual)return!1;let n=e.message?.id;return typeof n==="string"&&n.startsWith(xQo)&&e.message?.model!==Rl}
```

The same function, under other minified names, is in every build I have from 2.1.268 to 2.1.276. Threads are only attempted when `mN()` allows them: first-party auth, no `ANTHROPIC_BASE_URL` (so proxy mode, not `--endpoint`), and `CLAUDE_CODE_TETHER_LIVE` or the `tengu_curious_tower` rollout flag. Whether a given user hits this therefore depends on a server-side rollout. I observed it on 2.1.274 and 2.1.276. Translated replies carry no `request-id` header, so the id prefix decides. `sdk-adapter.ts` gave every translated reply `'msg_' + Date.now()` (the stream path in `writeAnthropicStream` and the non-stream path in `generateAnthropicResponse`), so each one qualified as an anchor. The following request then carried `thread:{type:"continue",previous_message_id:"msg_…"}` and only the messages after the anchor. clodex does not read `thread`, so the provider received a bare tool result with nothing before it. This is a captured request after a tool call (abridged):

```
{"model":"muse","messages":[{"role":"user","content":[{"type":"tool_result",…}]},…],
 "thread":{"type":"continue","previous_message_id":"msg_1789680540945"},"stream":true}
```

This is what went upstream for it (395 bytes):

```
{"model":"muse-spark-1.3-contributor","input":[{"type":"function_call_output","call_id":"call_01a0b13d…","output":"step-one"},…],"store":false,…}
```

## The change

Translated replies get `'clodex_' + randomUUID()` (dashes removed) instead, from one helper used at both sites. Claude Code does not anchor on them, so for new replies it sends the full history, which is the request these routes are built to translate. Nothing else changes, and nothing in `src/` or `tests/` read the old id shape.

Left out deliberately: clodex still ignores `thread`. Reconstructing a thread server-side would mean keeping conversation state that these routes do not otherwise need.

## Tests

A new `translated message ids` block in `tests/sdk-adapter.test.ts` checks three things:
- the streamed `message_start` id does not start with `msg_`
- the non-streamed message id does not start with `msg_` (driven through a real `@ai-sdk/openai` provider with a mocked `fetch`)
- two translated messages get different ids, each matching `/^clodex_[0-9a-f]{32}$/`

Deletion mutation (the helper returning `'msg_' + Date.now()` again): all three fail. A timestamp id with the new prefix (`'clodex_' + Date.now()`) also fails the third, on the format check.

## Runtime evidence

macOS 26.6.2 (arm64), Node 22.23.2, Claude Code 2.1.276, clodex proxy mode with this commit on my local branch (which also carries the Muse Responses route):

- Muse, before: the capture above (each post-tool-call request a delta, a 400, then a full retry).
- Muse, after: a `claude -p --model muse` session making three separate Bash tool calls made 4 requests, all 200, and the session transcript shows `clodex_…` message ids.
- Flash, as a non-regression check only: a two-tool-call session made 3 requests, all 200. Flash is on the Anthropic Messages transport, so its replies are forwarded with the provider's own ids (UUIDs) and never reach the code this PR changes.

Full suite: `2825 passed | 20 skipped`, with `tests/http-proxy-index.test.ts` timing out once under full-suite load and passing 11/11 when run alone.

## What I could not verify

- OpenAI OAuth routes also go through `writeAnthropicStream`, and their replies carried `msg_` ids too, so they qualified as anchors before this change. I couldn't run one live, because my ChatGPT quota is spent until the weekend. Their websocket diagnostics from 2026-09-12/13 show full-history requests, and I have no OAuth traffic from after I first saw the Muse deltas. If OAuth sessions do get deltas now, this change sends full history instead, which is what the websocket chain matching expects.
- Upstream's completions-routed Go models (Kimi, GLM, Qwen and so on) go through the same writer, so I expect they had the same 400s. That is inference, and I didn't run them.

## Out of scope: passthrough routes have the same problem, and fail silently

Passthrough forwards the provider's message id unchanged, and forwards the request body unchanged, `thread` included. Some OpenCode Go models on the Anthropic Messages transport return `msg_` ids. I checked with one direct request each to `/v1/messages` today:

| model | id returned |
|---|---|
| `qwen3.7-max` | `msg_8ed0ab5b-…` |
| `qwen3.7-plus` | `msg_09e950db-…` |
| `qwen3.8-max` | `msg_4c571f9f-…` |
| `minimax-m3` | `06fc3fa1…` (no prefix) |
| `deepseek-v4.1-flash` | `e3bcf999-…` (no prefix) |

So in proxy mode Claude Code can anchor on a Qwen reply and send the next request as a delta. Unlike the translated case, the gateway does not reject it. I have one transcript of this on a model that has since been withdrawn, a Go passthrough model whose ids were `msg_chatcmpl-…`. The next request went out with about 1,000 input tokens where the previous one had about 85,000, got a 200, and the model answered as if the conversation had just started. I have not reproduced it on the Qwen models. Fixing it means either rewriting `msg_` ids on non-Anthropic passthrough routes, or refusing a `thread` continuation on routes that cannot hold one, and either needs its own review, so I've left it for a separate change.

## Upgrade and rollback

This change stops new anchors, but it does not retire old ones. Claude Code picks its anchor by searching backward for the last qualifying assistant message, so a newer `clodex_` reply does not hide an older `msg_` one. A session that started before the upgrade can therefore keep sending continuations anchored on an old id, and so can a session that switches from a genuine Claude model to a routed one, since the genuine reply's id still qualifies. A continuation that opens with user text can then be answered without its context and still return 200. #253 closes that by refusing thread continuations on routes that cannot hold them. This PR alone is a prevention layer, not the whole fix. Reverting the commit restores the old ids and the old behaviour.


TweedBeetle:
A correction to "What I could not verify" above. I went back through the proxy's request diagnostics on my machine, and OAuth routes did get continuations: 71 of them on openai-oauth on 2026-09-12, all anchored on the fabricated ids (for example `msg_1789233423894`). Seventy opened with a tool result. One opened with user text, the kind a provider can answer without error, so that turn may have lost its context silently. So this change does affect OAuth in practice, and in the direction the PR describes.

#253 covers the rest of this class. It refuses a continuation on any route that can't hold a thread, translated routes included, and rewrites `msg_` ids on passthrough from hosts other than Anthropic.

bman654 (approved):
**Verdict: approve. No blocking findings.** Reviewed at head `e712110`, diffed against base
`2608576`. This is a correct diagnosis, a minimal fix in the right place, and tests that pin the
three properties that matter independently of each other. Thank you for the write-up — the bundle
quote, the captured request/upstream pair and the "Out of scope" section are what made this fast to
confirm — and for coming back with the OAuth correction unprompted rather than leaving "could not
verify" standing. Our own data (below) agrees with that correction.

## What we confirmed independently

We did not take the diagnosis on trust. Four separate confirmations, all reproducible:

1. **The gate is real and you quoted it correctly.** Pulled verbatim out of our own cached 2.1.276
   bundle:

   ```console
   $ B=~/.cache/clodex-review-bundles/new/claude-2.1.276-9de364db11a410d5.js
   $ grep -o -a 'var xQo="msg_";function qst(e){[^}]*}' "$B"
   var xQo="msg_";function qst(e){if(e.isApiErrorMessage)return!1;if(e.requestId)return!0;if(e.isVirtual)return!1;let n=e.message?.id;return typeof n==="string"&&n.startsWith(xQo)&&e.message?.model!==Rl}
   ```

   Widening the pattern to `…}[^}]*}` also pulls in `Kst`, the backwards walk that consumes it.

2. **Executed, not just read.** We lifted `xQo`/`qst`/`Kst` verbatim out of the bundle and ran them
   over hand-built transcripts. `msg_<ts>` (today's `main`) anchors; `clodex_<32 hex>` (this PR)
   returns `undefined`, which closes both the engine's `no_continue_pointer` short-circuit and the
   `anchor_missing` branch in the request builder. That is your whole claim, executed.

3. **It is live on our machine too, at roughly 20x the volume you measured, and still failing
   today.** We streamed all 51 `~/.clodex/logs/sessions/*websocket-diagnostics*.jsonl`
   (1,201,106 lines, 594,676 `request_diagnostic` records) and bucketed translated-route requests by
   thread class. Snapshot as of 2026-09-18 ~11:00 local — the log is live, so today's row keeps
   climbing while you read this:

   ```
    date       | thread class    | outcome                          count   (create, same day)
    2026-09-11 | thread:continue | ERROR invalid_request_error 400     286   (312)
    2026-09-12 | thread:continue | ERROR invalid_request_error 400   1,270   (1,323)
    2026-09-13 | thread:continue | ERROR invalid_request_error 400   2,210   (2,302)
    2026-09-14 | thread:continue | ERROR invalid_request_error 400   3,194   (3,344)
    2026-09-15 | thread:continue | ERROR invalid_request_error 400   3,197   (3,325)
    2026-09-16 | (no thread key on any request that day)                 0   (0)
    2026-09-17 | thread:continue | ERROR invalid_request_error 400   5,719   (5,974)
    2026-09-18 | thread:continue | ERROR invalid_request_error 400   7,050   (7,405)
   ```

   **22,926 translated `thread:continue` requests since 2026-09-11, and every single one failed** —
   a 100.0% failure rate, no successes at all. The matched `thread:create` population on the same
   days, models and route is 23,985 requests with 37 failures (0.15%), and **none of those 37 is an
   `invalid_request_error`** (21 service-unavailable, 1 rate-limit, 15 mid-stream failures that had
   already emitted data). Every continue failure is `invalid_request_error` / 400 with
   `emittedModelData:false` and `emittedDownstreamData:false`, and the error-message size histogram
   is `22,924 x 87 bytes, 2 x 94` — one template with a per-`call_id` body, exactly the message you
   quoted. Every `previous_message_id` on a translated continue matched `/^msg_\d{13}$/`, i.e. the
   literal `'msg_' + Date.now()` this PR removes.

   Three things worth saying about that data. **First, all 22,926 are on `openai-oauth`** — so this
   corpus is independent confirmation of the correction you posted in the comments, from a machine
   you have never touched. Your OAuth routes really do get continuations, they really are anchored
   on fabricated `msg_<ts>` ids, and here they fail 100% of the time. **Second, the before/after
   boundary is 2026-09-11T23:02Z and it is Claude Code's rollout, not one of our releases** — for
   the six weeks before it, translated requests carry no `thread` key at all — so no clodex version
   straddle explains the signal, and 2.14.0 landing on 09-17 moved nothing. The 09-16 gap is the
   same rollout flipping off for a day, which matches what you wrote about it being server-side
   gated. **Third, it is not decaying:** 7,050 failures on 09-18 alone. This is the strongest
   argument for landing the PR that exists in this repo, and it is independent of your machine.

4. **Mutation.** Reverting `translatedMessageId()` to `return 'msg_' + Date.now();` and running the
   **whole test file** (not `-t` isolation) reds exactly your three tests and nothing else —
   `Tests  3 failed | 93 passed (96)`. Run again over the **full suite** it is `3 failed / 2823
   passed`, i.e. nothing outside this change's own tests depends on the hunk — that is also the
   severability evidence. Widening to `'msg_' + randomUUID()` reds all three (the prefix is pinned
   independently of randomness); a constant `clodex_000…` and `'clodex_' + Date.now()` red only the
   third (uniqueness and the format are pinned separately). No sha256 digest pin is involved
   anywhere.

CI is green at this head (commitlint / test / windows-launcher).

## Concerns we checked and dismissed — so you do not have to re-derive them

These were all raised in review and all died. Recording them so they are not re-litigated later:

- **Nothing in clodex reads the old id shape.** Traced every hit of `msg_` and `message.id` in
  `src/` and `tests/`: on this head the only `msg_` in `src/` is your new doc comment; the only
  `message.id` reads are your three new tests. The one other production parser of translated
  `message_start` frames (`src/http-proxy/server.ts:100-103`) reads `message.usage` only, and
  `src/upstream-forward.ts:147-163` rewrites `message.model` only.
- **OAuth continuation and `previous_response_id` are a different identity.** The head-matching
  chain in `src/oauth/responses-websocket.ts` captures the upstream OpenAI *response* id from
  `responseIdFromEvent(event)` (`:1351`, assigned at `:1405`/`:1438`), strips ephemeral upstream item
  ids before snapshotting, and chains on `entry.responseId` (`:2236`, consumed at `:2736`). The
  Anthropic wrapper id is synthesized after all of that and never enters it. `lastUsedAt` head
  ordering uses the injected clock, not any id, so the known `Date.now()` tie hazard is untouched by
  this change.
- **No cache key, compaction check or trace correlation uses it.** `claudeSessionPromptCacheKey`
  (`src/sdk-adapter.ts:150`) hashes the session UUID; `openAiPromptCacheKey` (`:176`) hashes the
  system prompt and tool definitions; the OAuth prompt fingerprint explicitly removes `input` and
  `previous_response_id`; auto-compaction recognition reads the final user envelope. The `::ts::`
  tool-use separator only ever sees `tool_use.id`. There is no response cache keyed on this id.
- **No retry or ordering contract on the timestamp.** One id per `writeAnthropicStream` call, held
  for every frame of that response; the non-streaming id is allocated only after generation
  succeeds. Nothing in clodex maps, sorts or compares it. Dropping monotonicity costs nothing and
  removes same-millisecond collisions.
- **`clodex_` cannot accidentally satisfy some other Claude Code check.** We extracted all 652
  distinct `startsWith("…")` literals in the 2.1.276 bundle and tested each against `clodex_`: zero
  are a prefix of it. The only two anchored id regexes (`/^cmsg_[A-Za-z0-9_-]{1,128}$/` for Remote
  Control relay ids, and `/^msg_[A-Za-z0-9_]{1,64}$/` whose sole consumer sits behind
  `function jin(){return!1}`) do not match it; `deserializeMessages` only stamps an id when one is
  absent. Every other `message?.id` use in the bundle is equality-based grouping and is
  prefix-agnostic.
- **The `if(e.requestId)return!0` escape hatch really is unreachable on translated routes.** A probe
  stood up a fake OpenAI-compatible upstream that deliberately sends `request-id:`, drove a real
  translated route through `startProxyCatalog`, and confirmed the header does not survive to the
  client on either the streaming or non-streaming path (client saw only
  `content-type,cache-control,connection,date,transfer-encoding`). So the id prefix is the sole
  decider, as you wrote.
- **`/usage` accounting for translated turns changes shape — hygiene, not a defect, no change
  requested.** Claude Code's usage scanner keys records on `requestId`, then an `"id":"` beginning
  `msg_`, then the per-record transcript `uuid`. Post-PR, translated records fall through to the
  `uuid`, and one API response can emit several assistant records (one per content block). So a
  translated turn may be counted per block rather than per response. We did not execute the scanner,
  so this is inferred. It also does not matter: the pre-PR key was itself unsound (two turns in the
  same millisecond collided), and repo CLAUDE.md already states cost display is always inaccurate
  for routed third-party models.
- **A worry about a header leak into #253 turned out not to reach the client, so #253 does not have
  to defend against it.** The theory was that because the outer MITM's `copyResponse()` forwards
  `upstream.rawHeaders` unfiltered (`src/http-proxy/server.ts:287`), an OpenCode Go `request-id`
  response header could reach Claude Code and anchor a turn regardless of any body-id rewrite. It
  cannot, on clodex's routed passthrough: that route is served by `relayAnthropicMessages`
  (`src/upstream-forward.ts:202`), which reads the upstream with `fetch` and then writes its own
  response headers — `Content-Type`/`Cache-Control`/`Connection` on the streaming path (`:243`),
  `Content-Type`/`Content-Length` on the non-streaming one (`:290`), `Content-Type` on the error
  path (`:237`) — and never copies an upstream header. The `copyResponse` at
  `src/http-proxy/server.ts:725` is copying *that* sanitized inner response, not the origin's. The
  raw-relay `copyResponse` call site (`:439`) is the one that talks to `api.anthropic.com` directly,
  which is the upstream that legitimately holds threads. We have not obtained a real OpenCode Go
  response-header dump, so this rests on clodex's code path rather than on a vendor observation —
  but that is the only path a Go response can take to the client.

## Minor (non-blocking)

**M1 — "so it sends the full history" is too strong in a mixed-provider session.** `Kst()` walks the
transcript backwards and returns the *first* qualifying assistant message. When translated messages
stop qualifying, the walk does not stop — it keeps going to the newest message from a route that
does qualify. Reproduce with the extracted gate:

```js
// verbatim xQo/qst/Kst from the 2.1.276 bundle, Rl = "<synthetic>"
Kst([U(), A("msg_01Ant", {requestId:"req_real"}), U(), A("clodex_aaa"), U(), A("clodex_bbb"), U()])
// => "msg_01Ant"   (and the same without requestId — the msg_ prefix alone is enough)
```

Reachable sequence, proxy mode with the rollout on: a few turns on a real Anthropic model, `/model`
to a clodex alias (that turn is `config_changed` → `create`, fine), then a second turn on the alias
— config unchanged, history appended, engine says `append` → `continue`, `Kst()` skips the
`clodex_…` records and anchors on the last real Anthropic turn. `sliceFrom` then drops everything up
to and including it. Unlike the tail-anchored case this slice is usually well-formed, so it returns
**200 and silently loses the prefix**.

This is **not a regression and not a blocker** — on `main` the anchor is always the tail, so the
slice is strictly smaller; this PR only enlarges it. But the PR body's *"Claude Code does not anchor
on them, so it sends the full history"* and the identical sentence in the `translatedMessageId()`
doc comment (*"An id outside the anchor prefix makes Claude Code send the full history in the first
place"*) are stronger than what the gate does, and a wrong claim in one of our own comments is
exactly what propagates into the next PR.

**No push required, and we are not asking for one.** Editing the PR body costs you nothing (a clause
like "unless an earlier assistant message from a route that *does* hold a thread is still in the
transcript, which #253 closes") — do that if you like. The identical clause in the source comment
would cost a force-push and a full CI cycle on a green PR, which is not a fair trade for comment
text; **we will add it ourselves in the doc stamp we already owe** (see the last maintainer bullet).
The finding's real consequence is for us, not for you: #253's *refusal* half is load-bearing and
cannot be dropped in favour of its id-rewrite half.

**M2 — the version span is off by one release, in the safe direction.** The comment says "the gate
is in every build from 2.1.268 on" and the body says "every build I have from 2.1.268 to 2.1.276".
It is present from **2.1.266** and absent in 2.1.259 through 2.1.263:

```console
$ for v in 2.1.259 2.1.260 2.1.261 2.1.263 2.1.266 2.1.267 2.1.268 2.1.269 2.1.273; do
    f=$(ls ~/.cache/clodex-review-bundles/$v/*.js | head -1)
    printf '%s gate=%s anchorReasons=%s\n' "$v" \
      "$(grep -c -a 'isApiErrorMessage)return!1;if(e.requestId)return!0' "$f")" \
      "$(grep -o -a 'anchor_missing\|anchor_is_tail\|anchor_incomplete' "$f" | wc -l | tr -d ' ')"
  done
2.1.259 gate=0 anchorReasons=0
2.1.260 gate=0 anchorReasons=0
2.1.261 gate=0 anchorReasons=0
2.1.263 gate=0 anchorReasons=0
2.1.266 gate=1 anchorReasons=3
2.1.267 gate=1 anchorReasons=3
2.1.268 gate=1 anchorReasons=3
2.1.269 gate=1 anchorReasons=3
2.1.273 gate=1 anchorReasons=3
```

You wrote "every build I have", which was honest about your bundle set, so nothing here is
overclaimed — "2.1.266" is just free to write in the body and saves the next person a bisect.

**M3 — your third test is stronger than the PR body claims.** The body says the third test "can pass
by chance, since two `Date.now()` calls usually differ". It cannot: line 2443 asserts
`toMatch(/^clodex_[0-9a-f]{32}$/)` before the inequality ever runs, so the deletion mutation reds it
deterministically. Reproduced here:

```
 FAIL tests/sdk-adapter.test.ts > translated message ids > gives each translated message a random id, not a timestamp
 AssertionError: expected 'msg_1789747034046' to match /^clodex_[0-9a-f]{32}$/
 ❯ tests/sdk-adapter.test.ts:2443:25
```

Understating coverage is the right direction to be wrong in; worth correcting in the body so nobody
later reads that test as weak. Body-only, no push.

## Nit, and one thing we got wrong ourselves

**N1 — the commit summary.** `fix(translate): give translated messages ids Claude Code won't anchor
a thread on` leads with the mechanism, uses `thread` (a Claude Code protocol field) with no
plain-language gloss, and does not say what the user stops paying for. Measured against
CLAUDE.md's release-note rules, that is a miss.

**We owe you a correction here, because our first pass called this blocking and it should not have
been.** Our own CLAUDE.md says "Your summary line is the release note", and the PR template carries
a checkbox for it — so you did exactly what we asked. What neither document says is that a
standalone single-commit PR is squashed with a maintainer-typed `--subject`, which is the only text
release-please ever parses. Two merged precedents confirm it: #237 shipped as
*"log OpenAI usage limits to help diagnose depleted allowances"* and #213 as *"make OpenCode Go
models answer again and add DeepSeek V4.1 Flash"* — neither line appears on any commit on its branch
or in its PR title (and #237's own title broke the same rules this nit cites). So the line that
reaches users is written at merge, and asking you to force-push a green PR to change text that gets
overwritten would be charging you a full CI cycle for nothing. **Nit only. No action needed from
you.** We will fix the doc gap on our side.

## For the maintainer at merge

- **Land it.** Squash, and type the release note as the subject — something in the shape of
  `fix(openai): stop the wasted failed request after each tool call on non-Anthropic models (#252)`
  (95 chars). Keep it scoped to what was measured: do **not** write an unconditional "prevents
  retries after tool calls", since the anchor path only fires under first-party auth, in proxy mode,
  with the thread rollout on, and passthrough routes still have the bug until #253 lands.
  `gh pr merge 252 --squash --subject "…(#252)" --body ""` — the empty body is mandatory.
- **Land order is free.** #252 touches `src/sdk-adapter.ts` + `tests/sdk-adapter.test.ts`; #253
  touches `src/proxy.ts`, `src/upstream-forward.ts` and their tests. No overlap, no rebase needed in
  either direction.
- **No fallout on `main`.** Nothing in `.claude/harnesses/`, the patch transforms or any pinned
  constant references the old id shape, so nothing needs re-pointing after this merge.
- **Follow-up for us, not for this PR:** `.claude/docs/claude-code-internals.md` has nothing about
  the thread-anchor gate. Stamp `qst`/`Kst`, the `previous_message_id` construction site, the `mN()`
  preconditions, the 2.1.266 first-appearance and the M1 mixed-provider caveat into that doc, so the
  next contributor does not re-derive it from the bundle the way this one had to.

PR #253, opened 2026-09-18
fix(proxy): don't let Claude Code thread a conversation through a route that can't hold it

TweedBeetle:
On some OpenCode Go models (qwen3.8-max, and by id shape the two Qwen 3.7 models), a Claude Code session could forget everything that came before a tool call. The next reply looked normal and nothing showed an error, but the model had only seen the tool result. This PR stops Claude Code from treating any non-Anthropic route as if it kept the conversation on its side: replies no longer invite a partial request, and a partial request that still arrives is refused in a way Claude Code answers by resending the whole conversation.

## The failure

Claude Code 2.1.276 in proxy mode, a saved alias `qwen` for OpenCode Go `qwen3.8-max`. The prompt gave a secret word, asked for three `echo` commands as separate Bash calls, and then asked for the outputs plus the word. Input tokens per request, from the session transcript:

| request | before | after |
|---|---|---|
| 1 | 98,766 | 98,768 |
| 2 (after the first tool call) | **91** | 98,786 (99,355 cache read) |
| 3 | (none) | 98,804 |
| 4 | (none) | 98,822 |

Before the change, the second reply was `You have **14,900,567 tokens** left. How can I help you?`, and the task, the remaining commands and the secret word were all gone. Every request got a 200. After the change, the reply was `step-one, step-two, step-three — PELICAN`.

## Root cause

Claude Code anchors server-side thread continuation on the previous assistant message. This is the gate in the 2.1.276 bundle:

```js
var xQo="msg_";function qst(e){if(e.isApiErrorMessage)return!1;if(e.requestId)return!0;if(e.isVirtual)return!1;let n=e.message?.id;return typeof n==="string"&&n.startsWith(xQo)&&e.message?.model!==Rl}
```

When a message qualifies, the next request carries `thread:{type:"continue",previous_message_id}` and only the messages after the anchor. Threads are only attempted when `mN()` allows them: first-party auth, `ANTHROPIC_BASE_URL` unset or pointing at the Anthropic API itself (so proxy mode, not `--endpoint`), and `CLAUDE_CODE_TETHER_LIVE` or the `tengu_curious_tower` rollout flag.

The anchor is chosen by searching backward for the last qualifying assistant message, so a newer reply without a `msg_` id does not hide an older one. It matters beyond the Qwen case: a session that switches from a genuine Claude model to a routed one keeps the genuine reply's id as a valid anchor, and can send a continuation to the routed provider after the switch. Refusing continuations covers that case whatever the anchor's origin.

The passthrough relay forwards the upstream's message ids and the request body unchanged. Asked directly today, Go's `/v1/messages` returned these ids:

| model | id |
|---|---|
| `qwen3.7-max` | `msg_8ed0ab5b-…` |
| `qwen3.7-plus` | `msg_09e950db-…` |
| `qwen3.8-max` | `msg_4c571f9f-…` |
| `minimax-m3` | `06fc3fa1…` |
| `deepseek-v4.1-flash` | `e3bcf999-…` |

So Claude Code anchored on the Qwen replies, and Go received the fragment as if it were the whole conversation and answered it.

Translated routes have the same exposure from the other side. The SDK rebuilds the request without `thread`, so a continuation that reaches one gives the provider only the fragment. #252 stops clodex's own translated replies from qualifying as anchors. This PR refuses any continuation that still arrives.

## The change

For a routed `/v1/messages` request in proxy mode (`src/proxy.ts`), where "holds threads" means Anthropic-format passthrough to `api.anthropic.com`:

1. A request with `thread.type === "continue"` on a route that does not hold threads is refused before any format branch, with 400 and `error.details.error_code: "thread_unsupported_request"`, and nothing is forwarded. This covers passthrough to any other host, every translated route, an anchor left over from before the upgrade, an anchor from a genuine Claude reply earlier in a session that has since switched models, and a `msg_` id that got past (2), for instance in a `message_start` split across several `data:` lines, which the line-by-line rewrite does not parse. Claude Code's classifier `XOe` reads that code through `OLn` (`APIError.error` → `.error.details.error_code`) and maps it to `unsupported_request`. Its retry resolver then marks that agent and model stateless (`jQo`, leaving other agents and models alone), logs `[tether] …: resending this turn stateless; this thread stays stateless on this model for the session`, and resends with the full message array. The same markers are in every build from 2.1.268 to 2.1.276.
2. On passthrough from any other host, the relay replaces a `msg_`-prefixed message id with `'clodex_' + randomUUID()`, in the SSE `message_start` and in the JSON body, so Claude Code sends full history instead of anchoring. `relayAnthropicMessages` does this only when the caller sets the new `anchorSafeMessageIds` option. The proxy's passthrough sets it, so count_tokens, the completions relay and endpoint mode are unchanged. Lines with nothing to change pass through byte-for-byte.

Requests to `api.anthropic.com` by passthrough are untouched, because Anthropic's API is the one that holds threads. A `thread.create` is forwarded, because it carries the full conversation.

This complements #252 and doesn't depend on it. The two use the same id format.

## Tests

The new `thread continuations on routes that cannot hold a thread` block in `tests/proxy.test.ts` goes through the proxy. It checks that:
- a continuation to a non-Anthropic passthrough upstream gets the 400, and `fetch` is never called, whether it opens with a tool result or with user text, is streaming or not, and whether or not its anchor id starts with `msg_`
- a continuation on a translated route gets the 400
- a continuation on a translated route whose upstream URL is `api.anthropic.com` also gets the 400, because the SDK drops `thread`
- a continuation to Anthropic by passthrough is forwarded with its `thread` intact, and its `msg_` id comes back unchanged
- a full-history `thread.create` to a non-Anthropic upstream is forwarded with all its messages, and its `msg_` id comes back replaced

The new `relayAnthropicMessages anchor-safe message ids` block in `tests/upstream-forward.test.ts` checks that:
- the option replaces a streamed `msg_` id and a JSON `msg_` id
- an id that doesn't start with `msg_` passes through byte-for-byte
- without the option, a `msg_` id is kept on both paths, including when the model is being rewritten

Fifteen mutations, each run over the full files, and every one turned something red:

| mutation | red |
|---|---|
| guard removed | 3 (the three refusals) |
| holds-threads ignores route format | 1 (translated route to Anthropic) |
| every host holds threads | 2 (passthrough refusal, create id) |
| no host holds threads | 1 (Anthropic continuation) |
| proxy doesn't set the option | 1 (create id) |
| guard refuses any `thread` | 1 (create forwarded) |
| SSE id rewrite off | 1 |
| JSON id rewrite off | 2 |
| JSON id rewrite always on | 3 |
| SSE id rewrite always on | 1 |
| any id rewritten, not just `msg_` | 1 |
| refuse only tool-result fragments | 1 (user text) |
| refuse only non-streaming requests | 1 (streaming) |
| refuse only `msg_` anchors | 1 (other anchor id) |
| strip `thread` when forwarding to Anthropic | 1 (Anthropic continuation) |

No existing test changed. The relay's SSE fixture terminates its last event, and the test decodes events rather than searching for text. Full suite: `2835 passed | 20 skipped`. `pnpm typecheck` and `pnpm build` are clean.

## Runtime evidence

macOS 26.6.2 (arm64), Node 22.23.2, Claude Code 2.1.276, clodex proxy mode, a local branch carrying this commit (and #252):

- The before/after sessions in the table above, on `qwen3.8-max`.
- The same three-tool-call prompt on the final build, on Muse (translated, OpenCode Go Responses) and on Flash (passthrough, UUID ids): 4 requests each, every one carrying full history (large cache reads), and both answered with the secret word.
- `curl` through the running proxy with a `thread.continue` body got the 400 on both a passthrough alias (`flash`) and a translated alias (`muse`), and nothing reached Go. A `thread.create` on `flash` was forwarded and answered 200.
- The proxy's request diagnostics on my machine recorded 251 routed continuations between 2026-09-12 and 2026-09-18: 71 on OpenAI OAuth and 180 on OpenCode Go. 223 of them opened with a tool result. The other 28 opened with user text, which the provider can answer without error, so those turns are the ones where context may have been lost silently.

## What I could not verify

- Claude Code's own reaction to the 400. That comes from the bundles, not from a live run: I couldn't get a live session to send a continuation after the upgrade, because a resumed session started a new thread on its first request. If that reading were wrong, the turn would fail visibly with the 400 rather than succeed silently without its context. I can't say how Claude Code would retry in that case.
- Endpoint mode (`server/router.ts`) is unchanged. Claude Code only threads when `ANTHROPIC_BASE_URL` is unset, so continuations should not reach it.
- Whether any host other than `api.anthropic.com` holds Claude Code threads (a staging host, a corporate gateway). I treat none as doing so. For such a route the cost would be full-history requests where a delta would have worked.
- The Qwen 3.7 models return the same id shape, but I only ran a session on 3.8.

## Upgrade and rollback

Nothing is stored. A session carrying an old `msg_` anchor gets the 400 once on that model, then continues stateless. Reverting the commit restores the old relay.


bman654 (approved):
## Review: approve

Reviewed independently from several angles, then adversarially re-checked, all on `6b03c98`. **No blocking findings.**
Everything below the verdict is optional, and one concern that was drafted as a change request is
written up here as a demotion so you do not have to re-derive it.

This is a careful PR. The failure is real, the mechanism is identified correctly down to the bundle,
and the "What I could not verify" section named the right unknowns instead of papering over them. The
one thing in it that you flagged as unverified — Claude Code's reaction to the 400 — is the thing we
verified for you, and it holds.

### What we confirmed independently

- **The 400 body is accepted and does what you said.** We cut the classifier, the nested-code reader,
  the retry resolver and the stateless marker (`XOe` / `OLn` / `un` / `jQo` in 2.1.276; `x5n` /
  `A5n` / `Le` / `Fts` in 2.1.268) out of **both** bundles and *executed* them against your exact
  `THREAD_UNSUPPORTED_BODY`. Classification `unsupported_request`, log line verbatim
  (`[tether] unsupported_request: resending this turn stateless; this thread stays stateless on this
  model for the session`), return `retry:tether-stateless`, stateless key stored as
  `` `${agentId ?? "root"}\x00${normalizeModel(model)}` ``, and the **next** request built from the
  full message array (wire message count 1 → 2). The custom retry token also survives
  `maxRetries: 0`, so the recovery is not budget-dependent. Your reading was right in every detail,
  and the markers are present in 2.1.268, 2.1.273, 2.1.274 and 2.1.276 — the range your body claims.
- **It is safer than you claimed.** Your stated worst case ("one visible 400 on a turn") is
  pessimistic. If the `error_code` string were ever wrong, an unclassified 400 on a continuation
  still falls to `retry:tether-replay`, which forces `{thread:{type:"create"},sliceFrom:0}` — that
  path is in every build we checked, 2.1.268 through 2.1.276. On **2.1.276 only** there is a further
  net: `lr()` resends any unrecognised 400 on a threaded turn statelessly (the string
  `unrecognised HTTP 400 on a` is absent from 2.1.268/273/274, so do not count on it on older
  clients). The stateless mark is session-local and per agent+model, not persisted, so the cost is
  one refused turn per model per session, not one per turn.
- **The `request-id` bypass does not fire.** `qst()` short-circuits `true` on `e.requestId` before it
  ever looks at the id, and `e.requestId` is exactly the `request-id` **response** header
  (`withResponse()` → `t.headers.get("request-id")`, and `requestID = headers.get("request-id")` on
  the SDK's error path). Three lenses independently *executed* the relay with an upstream emitting
  `request-id` and `x-request-id` on the JSON, SSE and error paths — one of them through real
  `node:http` against `startProxyCatalog` — and the header never reaches the client on any path.
  `relayAnthropicMessages` writes its own header set on all four exits and never copies
  `upstreamRes.headers`. The id-rewrite half of this PR is live, not inert, including for OpenCode
  Go.
- **The headline number is corroborated by the bundle, not just plausible.** 91 input tokens needs
  more than a short message array, and the continuation path supplies it: `Pn()` returns a reduced
  `system` and `tools: undefined` when the stored tools hash matches, on top of `sliceFrom: B+1`. So
  a lone `tool_result` plus a trimmed system really is ~91 tokens. Not a cache artefact (you report
  `cache_read_input_tokens` separately) and not local estimation (`count_tokens` is forwarded on
  anthropic-format routes).
- **The refactor is behaviourally identical on the pre-existing model-rewrite path.** Byte-compared
  (hex) against main's `anthropicSseModelRewrite` on 18 differentiating streams — CRLF, bare CR,
  multi-line `data:`, mid-JSON and mid-multibyte chunk splits, non-string `model`, missing `id`,
  `message: null`, unterminated tail — and against main's whole `relayAnthropicMessages` on 28
  body/option combinations comparing body bytes, status *and* the header object. Identical
  everywhere. Your `changed` flag does correctly restore what the removed outer conditions used to
  guarantee; a pretty-printed non-`msg_` JSON body still comes back byte-for-byte with
  `anchorSafeMessageIds: true`.
- **All 11 mutations in your table are still red at this head.** Re-run independently by two lenses,
  full-file, never `-t` isolation (one re-ran your table verbatim; one ran an overlapping
  13-mutation set of its own). **Three** rows under-report, because they predate `6b03c98`: guard
  removed is 6 not 3, every-host-holds-threads is 5 not 2, and SSE-id-rewrite-always-on is 3 not 1
  (it also reds two framing tests). Nothing to fix; noting it so a future re-runner does not think
  the table disagrees.
- **Both mandatory host-allowlist directions are pinned by behaviour**, which is what settles the
  demotion below. Our own runs in a clean worktree:

  ```
  upstreamHoldsThreads → hostname !== ''            →  Tests  5 failed | 77 passed (82)
  upstreamHoldsThreads → 'api.anthropic.invalid'    →  Tests  1 failed | 81 passed (82)
  ```
- **#252 independence holds.** Zero file overlap (`src/sdk-adapter.ts` + its test vs `src/proxy.ts`,
  `src/upstream-forward.ts` + theirs), trial merge is clean (`git merge-tree`, no conflicts), and
  #252 adds 3 tests on top of this head's 2835.
- Baseline at head, our env (`CLODEX_HOME=$(mktemp -d)`, `CLAUDE_CODE_ENTRYPOINT=cli`): typecheck
  clean, build clean, full suite 2835 passed / 20 skipped, changed files 82/82.

### Demotions — concerns we raised and then dismissed

Stated out loud so they do not come back at you later, and so the nits that survive carry weight.

1. **"Tests do not pin the exact `api.anthropic.com` trust boundary" — drafted MAJOR, demoted to a
   nit.** The mutation reproduces: `hostname === 'api.anthropic.com'` → `.endsWith('anthropic.com')`
   leaves 82/82 green. But the *mandatory* over-scope and under-scope mutations are already pinned by
   behaviour (5 red and 1 red, above), so this is a third, finer-grained mutation inside the trusted
   host's own registrable domain, not the required one. The specific extra test case the finding
   asked for also fails on its own terms: `api.anthropic.com.example.test` is `false` under the
   shipped predicate **and** under the `.endsWith` mutation that motivated the ask, so it cannot
   discriminate that mutation. And `staging.anthropic.com` is NXDOMAIN, so
   `validateCustomEndpointUrl()` refuses it as a custom endpoint — the only hosts the `.endsWith`
   mutation actually admits are real Anthropic-owned siblings (`api-staging.anthropic.com` resolves
   and answers `/v1/models` with a 401 `authentication_error`), where whether the mutated predicate
   is *wrong* turns on the exact vendor fact your body already discloses as unobserved. Asking you to
   freeze a policy guess as an oracle is not a blocker. It survives only as N1 below, in policy
   wording.
2. **The `request-id` bypass — refuted, closed.** See above. This was the sharpest objection we could raise against the change
   going in; it is not a defect at this head.
3. **"If the `error_code` string is wrong the user eats a failed turn" — refuted by construction.**
   The string is right and the fallbacks are good. We will not be posting that class of ask.
4. **Guard sits after the inbound-token check but before the static API-key/401 check.**
   (`src/proxy.ts`: invalid proxy token → 401 at ~397, OAuth resolved at ~480, guard at ~533,
   `Missing API key` at ~542.) Not an auth bypass and not useful disclosure: routes and capabilities
   are local trusted config, invalid local proxy tokens are rejected before routing, and OAuth
   tokens resolve earlier. Worst case is a stale keyless route costing one extra local round trip
   before the normal 401.
5. **Guard placed after the `count_tokens` early return** (~497, ahead of the guard). Correct as-is.
   The SDK's `countTokens` posts `{model, messages, tools, betas?, thinking?}` — there is no `thread`
   field on that request, so nothing can slip past.
6. **Endpoint mode.** Genuinely outside the threading path, for the reason you gave, verified:
   `Hw()` returns `IE(ANTHROPIC_BASE_URL)`, which is `host === 'api.anthropic.com'`, and endpoint
   launch sets `http://127.0.0.1:<port>`. (One wording correction in N5.)
7. **`thread.create` forwarding is not a new exposure.** main already forwarded `create` on the
   first turn of every session to the same upstreams.
8. **Noted, no action: every non-Anthropic passthrough stream now traverses the line transform it
   used to bypass**, because `anchorSafeMessageIds: !holdsThreads` is set for all of them. Measured
   consequence is that invalid UTF-8 bytes are now substituted (`80 ff` → `efbfbd efbfbd`) rather
   than relayed. For Anthropic-format SSE (always UTF-8 JSON) that is theoretical, and the
   line-boundary buffering already applied to every override route. Recorded so nobody rediscovers
   it as a mystery later.

### Minor / nit — all optional, none blocking

**N1 — a direct `upstreamHoldsThreads()` table would close three surviving mutations at once.**
`grep -rn upstreamHoldsThreads tests/` returns nothing; it is exercised only through two hostnames
(`api.anthropic.com`, `opencode.ai`). Three weakenings stay 82/82 green:

```
hostname === 'api.anthropic.com'  →  .endsWith('anthropic.com')   →  82 passed
hostname === 'api.anthropic.com'  →  .includes('anthropic.com')   →  82 passed
catch { return false }            →  catch { return true }        →  82 passed
```

Repro (from a clean worktree at this head):

```sh
cp src/upstream-forward.ts /tmp/uf.orig
# apply one of the three edits above to src/upstream-forward.ts
CLODEX_HOME=$(mktemp -d) CLAUDE_CODE_ENTRYPOINT=cli \
  pnpm exec vitest run tests/proxy.test.ts tests/upstream-forward.test.ts
cp /tmp/uf.orig src/upstream-forward.ts
```

The fail-closed `catch` is the one we would most like pinned, because it fails open in *both*
halves at once (continuation forwarded *and* the id left anchorable). The cheap correct form is a
pure-function table in `tests/upstream-forward.test.ts` — no DNS involved, so hostname choice is
free:

```ts
expect(upstreamHoldsThreads('https://api.anthropic.com')).toBe(true);
expect(upstreamHoldsThreads('https://API.Anthropic.Com')).toBe(true);
expect(upstreamHoldsThreads('https://opencode.ai/zen/go')).toBe(false);
expect(upstreamHoldsThreads('https://some-other.anthropic.com')).toBe(false);
expect(upstreamHoldsThreads('not a url')).toBe(false);   // fail closed
```

Please word the fourth line as **policy** ("only the exact host is trusted; everything else takes the
safe default"), not as a vendor fact — nobody here has observed whether an Anthropic sibling shares
thread state, and a maintainer who later learns that it does should be able to change the policy
without deleting an assertion that claims to know. That one line closes both the `.endsWith` and the
`.includes` mutation, so a `api.anthropic.com.example.test` row is not needed (it is inert against
`.endsWith`, though it would catch `.includes`); add it only if you want the lookalike shape spelled
out.

**N2 — "only re-serialize when something changed" is unpinned on the JSON path.** Mutating
`if (changed) text = JSON.stringify(message);` to an unconditional re-stringify leaves 82/82 green,
because every JSON fixture in both files is itself `JSON.stringify` output, so the round trip is
byte-identical. The shipped code is right (we proved it with a pretty-printed body). Making one
existing fixture's `raw` a pretty-printed string — e.g. `tests/upstream-forward.test.ts:363` — makes
the property visible: whitespace, `1.0` → `1`, integer-like key reordering and `Content-Length` all
ride on it.

**N3 — small SSE/JSON asymmetry introduced by the refactor.** The SSE branch now tests
`rewrite.model !== undefined`; the JSON branch kept `options.responseModelOverride` truthiness. With
`{responseModelOverride: '', anchorSafeMessageIds: true}` they disagree (SSE blanks the model, JSON
does not). **Unreachable today** — `proxy.ts` only supplies `responseModelOverride` when a route
resolved, so it is either a real alias id or `undefined`, never `''`, and `src/server/router.ts`
never sets `anchorSafeMessageIds` — so this is hygiene only. It is worth the one-word fix because the
comment right above that block exists precisely because the two directions of the same relay once
disagreed about what they were allowed to touch.

**N4 — nothing pins "clodex emits no `request-id` response header".** That property is what makes the
id rewrite decide anything, and it is asserted by no test. Forwarding upstream response headers
(`anthropic-ratelimit-*`, `retry-after`, an upstream `request-id` for support tickets) is a plausible
future change that would silently make the rewrite inert — the 400 guard would still hold the
correctness line, but the rewrite exists to avoid needing it. One line in the existing proxy test
names the invariant: `expect(res.headers['request-id']).toBeUndefined();`.

**N5 — one body sentence, and this one is partly on us.** "Endpoint mode (`server/router.ts`) is
unchanged" is right about `clodex server --endpoint`, but `clodex claude --endpoint` also calls
`startProxyCatalog()` (`src/cli.ts` returns early to the HTTP-proxy command only when
`bridgeMode === 'proxy'`, and otherwise falls through to `launchClaudeViaCatalog`, which starts the
catalog proxy), so the guard is armed and ids are rewritten there too. Your **conclusion** is still
correct, for the other reason you gave: `buildChildEnv()` sets
`ANTHROPIC_BASE_URL=http://127.0.0.1:<port>`, `Hw()` is false, and Claude Code never threads in
either endpoint flavour. The file mapping you needed is in `.claude/docs/launch-and-wrapper.md`
("`src/proxy.ts` for the claude launch path, `src/server/` for the standalone gateway"), but the
top-level `CLAUDE.md` summary describes endpoint mode with no file mapping at all, which reads as if
endpoint mode lives entirely in `src/server/`. We will tighten that summary. Only the sentence needs
a tweak; no code change.

**N6 — optional design suggestion: strip `thread` from a forwarded `create`.** For an upstream that
holds no threads, the field can only be ignored or rejected; `holdsThreads` has already identified
the only upstream that acts on it. A strict Anthropic-compatible upstream that 400s on an unknown
top-level field with a message that does not match Claude Code's
`/\bthread\b.{0,3}(param|field|request|type|:)/i` pattern is classified `null`, and on a `create`
that falls through to `lr()`, which retries stateless *without* marking the model — so it repeats
every turn. Not a regression (main forwards it too), and you observed a 200 on `flash`, so this is
your call. The existing create test (`tests/proxy.test.ts:2096`) asserts the forwarded `messages`
length, the status and the returned id shape, but not that `thread` was forwarded, so the change is
test-compatible as written.

**N7 — follow-up after both land, not a change request here.** `translatedMessageId()` in
`sdk-adapter.ts` (#252) and `anchorSafeMessageId()` here encode the same *invariant* ("an id Claude
Code anchors on starts with `msg_`"), each with its own prose explanation — 17 lines above
`translatedMessageId()`, and here spread across the `upstreamHoldsThreads` doc comment and the
`anchorSafeMessageIds` option doc. The signatures legitimately differ (mint vs. conditionally
replace), so this is not a copy-paste utility — but if Claude Code ever widens its gate, the fix
lands twice and nothing goes red if only one is done. A small `src/anchor-ids.ts` with the
explanation written once would settle it. Only actionable once #252 and #253 are in the same tree;
no cycle risk in either direction.

### For the maintainer at merge time

- **Land #253 first, then #252.** Either order merges clean (trial merge verified), so this is a
  preference: #253 is the one that stops *silent* context loss (a 200 with the conversation gone),
  and landing it first means #252 ships into a tree where a stray anchor is already backstopped.
  #252 then removes #253's one remaining wasted round trip on translated routes. Neither absorbs the
  other.
- **Rewrite the squash subject.** It is the changelog entry, and the current title is mechanism
  ("thread a conversation through a route that can't hold it"). The PR body's own first sentence is
  the better release note. Suggested, 74 chars:
  `fix(proxy): stop models from forgetting the conversation after a tool call`
- **Squash with an empty body** (`gh pr merge --squash --body ""`), so the PR title does not land in
  the merge-commit body and generate a duplicate changelog bullet.
- **No fallout on main.** No harness constants, patch anchors or pinned test counts depend on this
  change; `src/patch-transforms.ts` and the review harness are untouched. CI at this head is green on
  all three jobs (`test`, `commitlint`, `windows-launcher`); the PR is `BLOCKED` only on the missing
  approval, so leave the approval even if you admin-merge.

Issue #249, opened 2026-09-18
Let users add OpenCode Go models themselves, with a live transport probe


bman654:
## The problem, from a user's point of view

If OpenCode Go ships a new model today, an OpenCode Go user cannot use it today. They have to wait for someone to open a PR against clodex, get it reviewed and merged, and then wait for a release. An OpenAI user has no such wait — they can point at a newly published model as soon as it exists, because nothing in clodex gates which OpenAI model ids are allowed.

That asymmetry is not a policy decision, it is a side effect of how the two providers are configured, and it bites hardest exactly where it is least affordable.

## What made this concrete

PR #248 mapped `union-alpha`, a free "cloaked" model on OpenCode Go. The contributor did the work properly: they probed all three endpoints against the live gateway, recorded 200 on `/v1/messages` and 500 on the other two, and wrote it up.

models.dev added that model on 2026-09-16T14:44Z and deleted it on 2026-09-17T23:55:46Z — `chore: remove expired Union Alpha model`. It existed for about **33 hours**. The contributor fetched at 2026-09-17T20:50Z, inside the window, so every number in that PR was correct when taken. The model expired roughly three hours after the PR was opened, and well before it could have been reviewed, merged and released.

That is the real lesson. For this class of model — free, undisclosed operator, short-lived, currently `omen-alpha` and `ox-alpha-free` in the same feed — **the PR pipeline is slower than the model's lifetime**. No amount of review discipline fixes that. The mechanism has to change.

## Proposal

Ship a user-facing way to add an OpenCode Go model, and retire the maintainer-curated list.

Something like:

```
clodex providers add-model opencode-go union-alpha
```

At the moment the user runs it, clodex can do what a PR author currently does by hand, but better, because it has the user's credentials:

1. **Validate the id** against the provider's own `/v1/models` and against models.dev.
2. **Probe the transport live** — try `/v1/messages`, `/v1/chat/completions`, `/v1/responses` with a minimal request and keep whichever answers. This is the one fact that genuinely requires credentials, and it is the fact we currently ask a contributor to establish by hand and then take on trust.
3. **Persist the result** as a user-local model entry, so it survives upgrades and needs no clodex release.

This resolves both gaps found while reviewing #248, and the root cause:

- No maintainer-curated list means nothing to go stale, and no `Transport-mapped models missing from models.dev` failure (today one delisted id makes `pnpm update:opencode-go` exit 1 and refuse to regenerate the catalog for all 18 other models).
[...]

PR #255, opened 2026-09-18
feat(providers): add any OpenAI-compatible server, like OpenRouter, from providers add

marcelstenkewitz:

`clodex providers add` can now connect any service that speaks the OpenAI API, such as OpenRouter,
Together, LM Studio or vLLM. You give it a name, the server's address and an API key (optional for
a server on your own machine); it lists the server's models to check the connection and saves it,
and its models can then be favorited and given short names like any other. Before this there was no
way to add one: the menu offered OpenAI, OpenCode Go and ChatGPT sign-in only.

**Two questions for the maintainer first, because this may not be wanted.** The menu entry for a
custom endpoint was removed on purpose in `d01d0eb`, when clodex was trimmed to a Claude-to-OpenAI
bridge, so this PR *offers* it back rather than assuming it is welcome; closing it is a fine
answer. And `CONTRIBUTING.md` asks for an issue before a larger change: this is one new 84-line
source file plus a menu entry, but it does add a provider class the README's support-tier table
does not classify. I left the table alone, since a tier is your commitment to make, not mine. If
you would rather have the issue first, say so and I will open it.

## Problem and root cause

The registry side already exists: `addCustomEndpointProvider` (`src/registry/custom-endpoint.ts`)
and every reader of `custom-*` providers are in the tree with tests, but nothing calls the add
function, so a user cannot reach it. Reachability of the new path: any user who runs
`clodex providers add` (or picks "Use an OpenAI API key" on first run, which opens the same menu)
and pastes a base URL. They control the URL and the key; the server at that URL controls what it
answers, which is why most of this PR's review rounds were about not trusting that answer.

## The change

- `src/providers-custom-add.ts` (new, 84 lines): the prompts (name, base URL, a confirmation for
  plain `http://`, key) and the call to `addCustomEndpointProvider` with `kind: 'openai'`. It is a
  separate file because `providers-command.ts` is already over 1,000 lines.
- `src/providers-command.ts`: the menu entry "Custom OpenAI-compatible server" and its dispatch.
  The menu now always has an entry, so "All built-in providers are already configured" could no
  longer be reached and is removed.
- `src/registry/custom-endpoint.ts`: the add function now reports `credentialCleanupReconciled`,
  as the template add and provider removal already do, so the command knows whether to retry
  queued credential deletes.
- `src/registry/server-text.ts` (new, 25 lines) and the two model-list fetchers
  (`parseModelList`, `fetchAnthropicModels`): a model whose id or name contains a control
  character (C0, DEL, C1; checked on the trimmed string that is stored) is skipped where a server's
  list enters, so `providers add` and both `refresh-models` branches are covered by one rule. The
  custom flow prints a server's error text with control characters replaced by spaces.
- README, `clodex claude --help` provider list, `.claude/docs/registry-and-server.md`.

**Deliberately left out:** the Anthropic-format kind stays unexposed; per-endpoint custom headers;
any change to the URL guard (see below); the support-tier table row.

## Evidence

Environment for everything below unless stated: macOS (Darwin 24.6), Node 22.19.0, pnpm 10.34.5,
isolated `CLODEX_HOME="$(mktemp -d)"`, no proxy variables in the shell. Base is `origin/main`
`6d5b2bd`.

- `pnpm typecheck && pnpm test && pnpm build`: clean at the pushed commit: 123 test files passed / 2 skipped (the two Windows-only files), 2,894 tests passed / 20 skipped, rc 0.
- **Feature deletion and mutations, full-file runs, each against a copy of the tree:** 20 mutations
  of the command and registry change (menu entry, dispatch, kind, always-allow-http, `reconciled`
  always true / always false, the id hint, `onResult`, the http prompt and its default, cancel at
  name and at key, key trim, a refusal reported as success, the success line, both `validate`
  callbacks, the `/i`), 11 of the first control-character fix, 16 of the move to the fetch
  boundary (filter removed or narrowed in each fetcher, each range edge, no trim, delete instead
  of space, hint or error printed raw), and 6 of the trimmed-name fix. All die in the test meant
  to catch them. Two were instructive on the way: one "died" for the wrong reason because its
  pattern matched the template flow's identical line first, and one survived until a case with a
  clean name and a bad id was added (a model's name defaults to its id).
- **Real terminal (tmux) against a local fake OpenAI-compatible server, no key:** the menu shows
  the entry, the flow ends `Connected · 2 models`, exit 0, `providers list` shows
  `custom-local-server`, the model manager finds the models, and
  `models --alias kimi-k3=clodex:custom-local-server:moonshotai/kimi-k3` saves.
- **Real OpenRouter, read-only:** its `/models` through the same `fetchTemplateModels`: 445 models,
  all `modelFormat: openai`. 96 ids contain `:`; both id parsers split on the first colon, so
  `clodex:custom-openrouter:vendor/model:free` round-trips (measured by a reviewer, not just read).
- **One real use:** OpenRouter was added interactively with a build of this branch (`ae89b33`) on
  macOS; afterwards `providers list` printed `OpenRouter (custom-openrouter) — 445 models, auth:
  keychain (API key)` and two of its models are saved as favorites with aliases. The interactive
  session itself was not captured, only its result.
- **Reviewed adversarially five times before this push** by a separate Claude Opus agent with no
  access to my reasoning, each round on the delta the last one caused; no round found a blocker.
  Round 1: the README overstated the URL guard, `reconciled` was inferred rather than published,
  four surviving mutations. Round 2: a hostile server's error body and model ids reached the
  terminal and `providers.json` (measured in a real pty: screen clear, title change, a forged
  line under clodex's prefix). Round 3: the fix sat at the add call site, so `refresh-models`
  bypassed it (measured through the real refresh function against a local HTTPS server). Round 4:
  the moved fix checked a name before `trim()` and so dropped a model whose name merely ended in a
  newline, for every template that lists models (measured: 2 of 4 models stored instead of 4).
  Round 5 (that fix only): the two source lines hold. A 1,345-row fuzz (every code point
  0x00-0x100 plus the Unicode separators, at three positions, through all three fetch paths)
  stored no control character, a whitespace-only name falls back to the id, and 20 mutation runs
  left no survivor. Its one finding was a README sentence that still described an untrimmed check;
  that sentence and one phrase in the registry doc are reworded in this commit, and that wording
  is the only thing here no reviewer has read. It also noted that a name of exactly `models/` is
  stored empty after the existing prefix strip (every label path falls back to the id; existing
  behaviour, not changed).

**What I could not or did not verify:**

- The OS keychain write is mocked in tests (that code is unchanged); the one real write above was
  on macOS only. Nothing was run on Windows or Linux.
- `tests/credential-helper.test.ts > force-kills a helper that exceeds the runtime limit` is flaky
  on pristine `main`: it failed 2 of 4 full runs there (the test reads the pid file after it
  exists and before the helper writes to it) and passes alone 3 of 3. Not fixed here.
- `src/registry/add-template.ts:292` (`result.credentialCleanupReconciled = true`) is not pinned by
  any existing test; deleting it leaves the suite green. Upstream's, noted, not fixed here.

**Known gaps, none introduced by this PR unless it says so:**

- The URL guard (`validateCustomEndpointUrl`, unchanged) refuses loopback, private, link-local,
  unique-local, carrier-grade-NAT and the well-known cloud-metadata addresses. It lets `0.0.0.0`
  and `::` through (the `unspecified` range; `https://0.0.0.0:<port>` reaches a listener bound to
  127.0.0.1, measured), as well as broadcast, the reserved ranges and `192.0.0.192` (Oracle's
  older metadata address). Two lines in `isBlockedIp` would close `unspecified` and `broadcast`;
  I did not block `reserved` because that would also refuse 198.18.0.0/15, which fake-IP DNS
  proxies hand out for real hosts (my inference, untested). The README now names what the guard
  does refuse. Happy to add the two lines here or separately.
- The control-character skip is silent when partial, and a list that is entirely skipped reads
  "Connected but no models were returned." (OpenAI-format) or "Could not list models (HTTP 200)."
  (Anthropic-format, reachable only by refreshing a hand-written registry; an empty `data: []`
  already read that way). It is write-time only: a `providers.json` written by an earlier build
  keeps what it has until refreshed. The raw error hint is still built by `fetchTemplateModels`;
  only the custom flow cleans it before printing, since the other templates' hosts are fixed.
  `parseOpenAiModelEntries` (the ChatGPT OAuth catalog, fixed host) has no such check.
- Bidirectional, zero-width, `U+2028` and BOM characters pass: they can mislead a label, they
  cannot act on a terminal. Id length and model count are unbounded (3,001 models measured, 1.34 MB
  of `providers.json`), and there is no length cap on the provider name.
- An `http://` or anonymous custom provider can never refresh its models (existing behaviour).
- First run prints "OpenAI provider ready" whichever provider was added from its menu (already
  true for OpenCode Go).
- URL normalisation drops userinfo and query (existing).

- [x] `pnpm typecheck && pnpm test && pnpm build`
- [x] Feature-deletion mutation run (full file, not `-t`) — the intended test fails without the fix
- [x] Manual smoke, if a launch path changed — not run; no launch path is changed. No provider was
      smoke-tested through a launch, OpenRouter included.
- [x] Commit summary line reads as a release note for a non-technical user

A failed add saves nothing: a refused URL, a failed connection check or an empty model list ends
the flow before anything is written. The key is journaled before it is stored
(`journalCredentialWrite`), so a key saved ahead of a failed registry write is left to the
existing pending-delete reconcile, which the add runs on both outcomes; the command retries it
when the result says cleanup is still pending. That is my reading of
`custom-endpoint.ts:206-297`, which this PR does not change apart from the reported flag. Nothing persisted changes shape: a custom provider is the same
`providers.json` entry and keychain item `addCustomEndpointProvider` already wrote, so there is
nothing to migrate, and removing the provider with `clodex providers` undoes it. Reverting this PR
leaves existing `custom-*` providers readable, as they are on `main` today.


bman654 (changes requested):
## Review — changes requested (one ask)

Reviewed head `7744ddf` against base `6d5b2bd`, reviewed from four independent angles, each finding then
adversarially re-checked. Six of the seven MAJOR findings did **not** survive that re-check; they are listed below under *Checked and dismissed* so nobody re-derives them. One
survived, and the maintainer has since set the policy that resolves it.

### Answering your two questions up front

**Yes, this is wanted.** You were right that `d01d0eb` removed the menu entry deliberately, and
right to ask rather than assume. The answer is to take it: the registry half already ships on
`main` — `addCustomEndpointProvider`, the materialization, model-source, refresh and catalog
readers, and `validateCustomEndpointUrl` are all live and consuming hand-written `custom-*`
records. Custom providers already work for anyone willing to edit `providers.json`; this makes
them discoverable. The alternative to wiring it up is deleting that half.

**The tier is community-supported**, and you were right that the tier was ours to set rather than
yours. Please add the row to the README support-tier table in this PR.

**No separate issue needed.** You offered to open one; this review is that conversation.

### What we verified independently

- **Gate reproduced exactly**, twice, in separate worktrees: `123 passed | 2 skipped` files,
  `2894 passed | 20 skipped` tests, `tsc --noEmit` rc 0, build clean. The PR body's numbers are
  right to the test. Pristine-`main` control: `121 | 2` files, `2841 | 20` tests — the `+2`/`+53`
  delta matches the diff.
- **Mutation work re-run, not taken on trust.** 42 independent mutations across the command,
  registry, and control-character paths, all as full-file runs, restored from a `cp` snapshot (never
  `git checkout`). 38 died, each in the test
  whose *name* describes it; four survived: the two real coverage gaps below, one cosmetic
  prompt-wording mutation, and one provably equivalent mutation. Both directions fire on
  the control-character rule: reverting the guard reddens the positive tests, narrowing it by one
  code point reddens the parameterised negative ones.
- **The new command test file includes a real integration**, not a hand-built fake: it drives
  `runProvidersAdd()` and `runProvidersCommand(['add'])`, invokes the real `validate` callbacks, and
  asserts bytes on the wire against a real `node:http` listener with the real registry on a temp `CLODEX_HOME`. No
  clock, no timers, no randomness.
- **Persisted shape is byte-identical to `main`'s.** The same deterministic anonymous custom add run
  at base and head produced `providers.json` files with the same sha256. Nothing to migrate;
  reverting leaves existing `custom-*` providers readable.
- **Colon-bearing ids, catalog cap, patch identity.** `vendor/model:free` round-trips through both
  id parsers, into the proxy route, into a patch config, and into the executed Agent-tool enum of a
  fixture patched with the real transforms. The constructed 445-model provider still yields exactly 20 patch
  identities and rejects the 21st favorite with `{ reason: 'cap' }`.
- **Terminal safety holds at the real boundary.** Driving the actual CLI under `tmux` against a
  hostile server: zero OSC title changes, zero `ESC [2J`, zero 8-bit CSI, CR/LF rendered as spaces,
  the four attack rows absent from `providers.json` and from the favorites picker. Redirects are
  refused (`redirect: 'manual'`) and the credential never reaches the redirect target. The
  credential is stored only as an `authRef`; `providers.json` holds no secret.
- **Docs checked claim by claim** — base-URL definition, `custom-<name>` id, `clodex:custom-…:<model>`
  addressing, "checked before anything is saved", the trim rule, both fetch boundaries, the registry
  doc's `templateId`/`authRef` values. All accurate.

Credit where it is owed: the body's *Known gaps* list is accurate everywhere we could test it
(`0.0.0.0`/`::`/broadcast/`192.0.0.192` pass the guard exactly as stated; hex, decimal and
IPv4-mapped loopback spellings are refused, which the author did not claim either way), the
disclosed `refresh` limitation is correct *including* its "(existing behaviour)" parenthetical, and
`src/registry/add-template.ts:292` really is unpinned on `main` while this PR's analogous line
dies under mutation. Several things that would have been findings are not, because the author wrote
them down first.

---

## BLOCKING (1)

### B1 — a server that publishes its own context limit has it discarded, and the guess is never shown

`parseModelList` (`src/registry/fetch-template-models.ts:137-141`) resolves the window as
`row.context_length ?? row.contextWindow ?? row.context_window ?? resolveContextWindow(id)`. vLLM —
named in the menu hint this PR adds — serialises `max_model_len` in each OpenAI `ModelCard`, and
that key is not in the chain, so an honest server answer is thrown away in favour of a regex on the
id. Nothing prints the resolved window, so the user gets no signal that a guess happened.

Reproduced independently through the real `addCustomEndpointProvider` against a real
`127.0.0.1` listener, with no mocks:

```
ADDED true
STORED id=llama-3.2-3b-instruct contextWindow=131072   # server said max_model_len: 8192
STORED id=qwen3-0.6b            contextWindow=262144   # server said max_context_length: 4096
STORED id=server-explicit       contextWindow=32768    # server said context_length: 32768  ✓
STORED id=my-finetune-v1        contextWindow=200000   # server said nothing
```

Why it is behavioural and not cosmetic: `modelsCache.models[].contextWindow` reaches
`src/env.ts:183` (`CLAUDE_CODE_MAX_CONTEXT_TOKENS` for the launch model) and `src/patcher.ts:229,343`
(the window baked into the patched binary). Claude Code reserves 33,000 below whatever it is told,
so a 16x **over**-report means auto-compaction never fires and the session dies on an upstream
context-length error with no explanation. The other direction was measured live: the same door
pointed at `https://opencode.ai/zen/go/v1` (no key, metadata-free rows) stores 200,000 for `kimi-k3`
against the curated 1,048,576 — 848,576 tokens of context silently lost, and `--context` cannot
repair it, because `resolveContextStop` (`src/context-modes.ts:80`) uses the persisted window as the
ceiling and only clamps downward.

Why this is in scope: `parseModelList` is in a file this PR edits, and the new menu explicitly
advertises vLLM. The body does not disclose inferred context windows. This is not a regression
against `main`, but the PR turns arbitrary custom discovery into a supported-looking menu path. Its
only real provider exercise, OpenRouter, could not expose the problem because OpenRouter publishes
`context_length`.

**Ask — and the maintainer has set the policy, so this is now a concrete spec rather than an open
question.** The rule is *refuse rather than guess*:

1. **Probe the server's published limit, including vLLM's `max_model_len`.** The chain becomes
   `context_length ?? contextWindow ?? context_window ?? max_model_len`. A parser test should store
   8192 when the server reports `max_model_len: 8192`. We are not asking you to claim this fixes LM
   Studio: its `max_context_length` lives on the native `/api/v1/models`, and its OpenAI-compatible
   `/v1/models` exposes no context metadata.

2. **If the probe cannot determine a window, the add FAILS** rather than falling back to the
   id-based guess. Tell the user exactly what to do: re-run with `--context` to supply their own
   value. Refusing beats silently persisting a number we invented, because
   `modelsCache.models[].contextWindow` is what reaches `env.ts:183` and the patched binary.

3. **`--context` wins over the probe.** When the user supplies it, that value is used even for
   models that did publish one. Warn when it differs from a probed value, naming whether it is
   higher or lower, so an accidental override is visible.

**Scope this to the custom-endpoint flow.** `fetchTemplateModels`/`parseModelList` are shared with
`add-template.ts` and the refresh paths, which serve the curated providers. Do not change what the
built-in templates do — a blanket refusal there would break `openai`, whose catalog legitimately
carries no context metadata and is backfilled from curated data.

**One case we could not resolve for you, so please raise it rather than guess.** A discovered
catalog can be large and heterogeneous — real OpenRouter returns 445 models with very different
limits, most publishing `context_length` and some not. A single `--context` applied to all of them
would be wrong for most. Our reading of the rule for that case is: refuse the add only when the
probe found nothing for a model, let `--context` supply the window for exactly those, and warn
listing how many models it covered. If you would rather `--context` override every model in the
provider, say so on the PR and we will confirm the intent before you build it.

A parser test pins the `max_model_len` field; the command-flow tests can pin the refusal and the
override warning.

*(Context for why the policy is what it is: issue #249 reached the same conclusion independently
for metadata-free discovery — refuse or require an explicit `--context` instead of guessing. Both
paths now share one answer.)*


---

## MINOR / NIT (none blocking)

- **N1 — the two guard lines you already offered.** `isBlockedIp` lets `unspecified` through, and
  `https://0.0.0.0:<port>` does reach a listener bound to `127.0.0.1` — reproduced three times in
  our own testing, including through a stable public DNS record with the unmodified system resolver:
  guard `ok:true`, and the add reaches a loopback listener.
  Your disclosure of this is exactly right, and `if (range === 'unspecified' || range === 'broadcast')
  return true;` closes it. Taking the two lines here is fine by us; so is a follow-up. Leaving
  `reserved` alone for the fake-IP-DNS reason is a defensible call.
- **N2 — `Connected · N models` proves nothing about the key when `/models` is public.** With an
  empty key against a paid gateway the flow reports `added: true`, 37 models,
  `authRef: 'none:anonymous'` — and the user learns about the 401 at first launch. The repo already
  knows this shape: the OpenCode Go template carries `verifyCredential` with a comment saying exactly
  why. A generic equivalent is not possible; the cheap honest fix is to say
  `Connected (no key sent) · N models` when the key was empty.
- **N3 — test name promises more than the test sets up.** `tests/providers-custom-add.test.ts:107`
  is named *"…and is offered even when every built-in is configured"* but runs against an empty
  registry. Restoring the deleted `options.length === 0` early return leaves the **whole suite
  green** (verified on the full 2,894-test run), so the guarantee is pinned by nothing. Seeding the
  three built-ins inside that test — 12 lines — fails under that mutation and passes at head. The
  shipped code is correct; only the pin is missing.
- **N4 — one of the two early returns in `addCustomEndpointProvider` is unpinned.** The
  model-discovery refusal asserts `credentialCleanupReconciled` is undefined; the sibling URL
  refusal (`custom-endpoint.ts:170`) does not, and a wrongly-`true` flag makes
  `providers-command.ts:88-102` skip `reconcileCredentialCleanup()` silently. One added line in
  `tests/custom-endpoint.test.ts`.
- **N5 — README clause.** The new sentence's loopback parenthetical (`127.0.0.0/8`, `::1`) is
  literally accurate and `0.0.0.0` is genuinely a different range, so we are **not** calling it
  wrong (we tried; see the dismissals). But since the body documents the `unspecified` gap, one
  clause saying the guard keys on address class — and that `0.0.0.0`/`::` are not in the refused set
  — would stop a reader concluding that no local listener is reachable.
- **N6 — one reversed sentence in the PR body**, for the record rather than the code: the failure
  section says the command retries when the result says cleanup is pending. It does not retry in
  *that* invocation; it warns, and the next provider command retries. The de-duplication is
  sensible, only the wording is backwards.
- **N7 — prose nit, and a correction to our own review.** The commit summary passes the release-note
  bar as written (86 chars, names the outcome, no internal vocabulary, self-contained). One of our
reviewers
  suggested `like OpenRouter or LM Studio` to reach the local-server audience — do **not** take it:
  the summary becomes 99 characters; adding the required ` (#255)` suffix to the squash subject
  pushes the merge subject to 106, over commitlint's 100. Leave the line alone.

### Follow-ups we would file against `main`, not against this PR

- **Anonymous / plain-HTTP custom providers can never refresh.** Real and reproduced, and the body
  discloses it with an accurate "(existing behaviour)". Two causes, both at `refresh-models.ts:374`
  and `:648`, both untouched by this diff, both reproducing character-for-character on `6d5b2bd`.
  What deserves an issue is the *message*: an intentionally anonymous provider is told to "add this
  provider again with a real key before launching" when anonymous launch is supported — and
  `uniqueProviderId` would make "add again" create `custom-x-2` and strand the favorites.
- **First-run's `runProvidersAdd()` is not wrapped in `runWithCredentialCleanup`.** A registry-write
  throw after the credential is provisioned orphans the key. Reproduced — and reproduced identically
  on base `6d5b2bd` through first-run → built-in OpenAI (provision 1, delete 0, one pending journal
  entry), so it is `main`'s posture on the default fresh-install path, not something this PR adds.
- **`tests/credential-helper.test.ts > force-kills a helper…`** — the author reports 2/4 failures on
  pristine `main`; we saw 0/8, but the mechanism is real and machine-dependent: the fixture writes
  the pid with a plain `writeFileSync`, so a reader can see the file created and still empty (4
  non-numeric reads in 400 tight-spin trials). `main`'s test bug; fix is a `rename` in the fixture or
  polling for a numeric pid.

---

## Checked and dismissed

Stated out loud so they are not re-derived. Each of these was raised as a MAJOR and did not
survive refutation.

- **"The SSRF guard validates a different address than the one connected to" (TOCTOU).** The split
  is real — `validateCustomEndpointUrl` resolves, then `fetch` resolves again — and a
  dual-answer resolver harness does land a `Bearer` key on a loopback listener. Demoted to a note
  for three reasons: the relevant validation/fetch code is byte-identical to `main`, and the body
  scopes guard changes out; the entire incremental capability is already available with **one stable
  public DNS answer** via the disclosed `0.0.0.0` gap, i.e. strictly weaker preconditions and no
  rebind at all; and the proposed remedy (pin the validated answer to the socket) does not close that path, because
  `0.0.0.0` *is* the connected answer. Both proofs also needed `NODE_EXTRA_CA_CERTS` — without an
  already-trusted cert only a TCP connection lands and the add fails. Attacker picks the IP only
  (`redirect: 'manual'`, fixed path, victim-typed port), and there is no channel back. N1 is the fix
  that matters.
- **"The new door makes the OpenCode Go allowlist advisory."** The mechanism is real (37 live ids
  vs 18 curated; all routed over `/chat/completions`), but it is an explicitly user-selected custom
  identity, not the built-in Go identity, and #249 expressly permits user-supplied overrides. The
  finding's counts were also wrong (18 curated not 19, five Anthropic-format not four, 17 usable
  uncurated not 18 — the blacklist removes two), and no valid-key probe shows those models actually
  fail today; with an invalid key both transports return the same 401. A "this looks like OpenCode
  Go — use the built-in entry" hint is optional polish, nothing more.
- **"The README sentence overstates the URL guard."** Refuted: the sentence defines loopback as
  `127.0.0.0/8, ::1` and every address in that class is refused. `0.0.0.0`/`::` are `unspecified`
  and stay `unspecified` through `dns.lookup`; it is Darwin's *connect* that maps an unspecified
  destination to the local host. The finding silently widened "is, or resolves to" into "eventually
  reaches". Kept as N5 (a clarifying clause), not as a defect.
- **"Anonymous / local-HTTP providers can never refresh" as a change request.** Mechanism
  reproduced; attribution wrong. Zero lines of this PR touch the failing path, base returns the
  identical two error strings, `main` already ships this population (`d01d0eb` *removed* a shipped
  add flow; `'local'` is in `PLACEHOLDER_KEYS`; `isLegacyAnonymousCustomEndpoint` exists precisely
  to handle it), and main's own hint has named "Ollama, LM Studio, vLLM" since v0.4.0. "The
  advertised path is unavailable" is also overstated — real OpenRouter refresh at this head returns
  `ok: true`, 446 models. Follow-up issue above.
- **"First-run custom adds bypass the credential-cleanup wrapper" as a change request.**
  `runProvidersAdd()` calls `runProvidersAddWithCleanupState()` with no state *verbatim on `main`*,
  and the diff does not touch `src/first-run.ts` at all. Reproduced identically on base. Follow-up
  issue above.
- **"No support tier fits, therefore the table omission is a defect."** Demoted to the product
  decision at the top: `README.md:294` scopes "Providers supported" to three integrations and
  documents custom servers separately, which is an internally consistent boundary, and the body
  foregrounds the policy question twice. A third row is clarity, not a fix.
- **A hostile server can echo the API key into terminal scrollback** (the first 200 characters of an
  error body become the printed hint). Real and measured in a real PTY, but the endpoint the user
  chose already *has* that key; the incremental audience is local scrollback and a mode-0600 trace
  log. Non-blocking. Worth knowing: the trace redactor ignores registered secrets shorter than 8
  characters, so a short key reflected without `Bearer` syntax is written verbatim under
  `CLODEX_TRACE=1`.
- **Equivalent-mutation survivors.** `!result.added || !result.provider` → `&&` survives and is
  correct (every `added: false` return omits `provider`); the meaningful mutation of that line,
  `return 1` → `return 0`, dies with 9 failures. Prompt *wording* mutations survive; swapping the two
  prompts wholesale dies with 10.
- **File seam and architecture.** A new 84-line file instead of growing a 1,090-line
  `providers-command.ts` is the right call, as is 25-line `server-text.ts` shared by two unrelated
  modules. No hand-rolled translation, no launch path, no `~/.claude/settings.json`, no proxy body
  rewriting, no patcher change.

---

## Merge-time notes (for whenever it lands)

- Squash subject:
  `feat(providers): add any OpenAI-compatible server, like OpenRouter, from providers add (#255)`
  — 93 characters with the ` (#255)` suffix, inside commitlint's 100. Do not lengthen it (see N7):
  the `or LM Studio` variant lands at 106 with the suffix. Merge with `--body ""` as always.
- The branch is two commits behind `main` (`9e1e279`, `chore(main): release 2.14.1`); neither #254
  nor #256 touches a file in this diff and `git merge-tree` reports zero conflicts, so no rebase is
  needed — and if one is wanted, merge `main` into the branch rather than asking for a rebase.
- The README support-tier row (community-supported) should land in this PR, not after.
- No migration, no new persisted shape, no launch-path change, so the manual smoke test is genuinely
  not owed here — the unticked box in the checklist is honest.

bman654:
**Amending B1 — please don't build the add-time refusal I asked for.** My ask was the wrong shape,
and I'd rather correct it now than have you spend a round on it. The narrowed ask is smaller.

## Why the ask was wrong

I framed the context-window rule as "if the probe can't determine a window, the add fails." That
treats `providers add` as if it adds one model. It adds a provider, and its whole catalog comes
with it — real OpenRouter is 445 models. Refusing the add because some subset lacks metadata would
block a provider over models the user will never touch.

More to the point, an unknown window is **inert while it sits in `providers.json`**. It only does
damage once a model becomes a favorite:

- `buildPatchModelConfig` (`src/patcher.ts:196`) iterates **favorites only**, and an alias whose
  target is not a favorite is rejected as `target-not-favorite`.
- `src/env.ts:183` sets `CLAUDE_CODE_MAX_CONTEXT_TOKENS` for the **launch** model.
- `MAX_MODEL_CATALOG` is 20, so at most 20 of those 445 are ever exposed.

So the add is the wrong gate. It is both too aggressive (blocks a bulk add over unused models) and
too narrow (a curated provider with one metadata-free model isn't affected at all).

And the right seam already exists. `buildDesiredPatchConfig` collects `unknownWindows`, and
`clodex patch` already reports:

```
No context window metadata for <id> — Claude Code will assume the 200k default.
```

That is per model, at the moment the window starts to matter, and it already covers every provider
rather than only custom ones. Tightening that into a refusal-plus-explicit-`--context` is the
maintainer's call and belongs in its own change, so it lands once for everyone instead of twice.

## What we're actually asking for in this PR

1. **Read vLLM's `max_model_len` in the probe**, so the chain is
   `context_length ?? contextWindow ?? context_window ?? max_model_len`. A parser test that stores
   8192 when the server reports `max_model_len: 8192` is enough. This is the one part your
   advertised vLLM support genuinely needs, and it's a strict improvement on any policy we pick
   later. As before: not a claim about LM Studio, whose `max_context_length` is on the native
   `/api/v1/models`, not the OpenAI-compatible one.
2. **Add the README support-tier row: community-supported.** That answers your second question —
   the tier was ours to set, and that's it.

**Dropped from the ask:** the add-time failure, the `--context` flag and its override warning, and
the "warn that a guess happened" wording. None of that is yours to carry. The id-based estimate
stays as it is on `main` for now, and the unknown-window policy gets decided separately across all
providers.

**Also dropped:** the heterogeneous-catalog question I asked you to raise. It was a consequence of
the wrong framing, and it's answered by moving the decision to the favorite/patch seam.

Everything else in the review stands, including the dismissals. Sorry for the churn — the 445-model
case is what exposed it, and that came out of your own OpenRouter evidence.

bman654 (approved):
**Approving — both asks are done, and I verified them rather than taking the commit message for it.**

**vLLM's `max_model_len`.** The chain is now
`context_length ?? contextWindow ?? context_window ?? max_model_len ?? resolveContextWindow(id)`,
with the new key in the right position. I mutated it in both directions against the full file:

| mutation | result |
| --- | --- |
| remove `row.max_model_len ??` | your new test reds |
| move it ahead of `context_length` | your new test reds |

The second one is the one I care about: your fixture gives a model both `context_length: 32768` and
`max_model_len: 8192` and asserts 32768, so you pinned *precedence*, not just presence. That is the
part a later refactor would quietly break.

**The support-tier row** reads as asked, community-supported.

**Gates at `24e41f0`:** `pnpm typecheck` clean, 123 test files passed / 2 skipped, 2,895 tests
passed / 20 skipped, `pnpm build` success. Merges onto `main` with no conflicts.

---

Thank you for this. A few things were genuinely unusual and worth naming, because they changed how
this review went:

- **You asked before assuming.** You spotted that `d01d0eb` removed this deliberately and opened by
  inviting us to close the PR. That is the right instinct, and it meant the product question got
  answered first instead of after you had absorbed a round of code review.
- **You disclosed your own gaps** — the `unspecified` range in the URL guard, the silent partial
  skip, what you could not test, and an unrelated flaky test on `main` that you explicitly declined
  to charge to this PR. Reviewing a PR whose author has already told you where the bodies are is a
  completely different exercise, and a faster one.
- **Your evidence was reproducible.** Six of the seven significant findings raised against this PR
  did not survive checking, and in several cases it was your own write-up that refuted them.

And an apology for the churn: my first version of the context-window ask told you to fail the add
when the probe found nothing. That was wrong, and it was your own OpenRouter measurement — 445
models in a single add — that made it obvious. An unknown window is inert until a model is
favorited, so the decision belongs at the favorite/patch seam where `unknownWindows` already
exists, not at the add. You did not build the wrong thing, which is the outcome I was hoping for
when I retracted it.

The unknown-window policy is now being worked out separately, across every provider rather than
just this path. Nothing there is yours to carry.

Merging with the squash subject:
`feat(providers): add any OpenAI-compatible server, like OpenRouter, from providers add (#255)`

PR #261, opened 2026-09-21
fix(proxy): route non-intercepted CONNECT tunnels through the outbound proxy

marcelstenkewitz:
Ni## Problem

Clodex's proxy mode selectively MITMs `api.anthropic.com` and blind-tunnels every other host. The CONNECT handler dialed those targets with a direct `net.connect(target.port, target.host)` and never consulted the configured outbound proxy — even though the module already chains its *own* fetches through `outboundHttpProxyAgent` / `outboundProxyUrlForTarget`.

Consequence: on a host behind a corporate proxy, or when pointing an upstream mitmproxy at Claude Code's WebFetch traffic, clodex's CONNECT tunnels bypass `HTTPS_PROXY` entirely. (Verified live: an upstream `mitmproxy --map-local` saw none of Claude's outbound requests because clodex stripped the proxy on the CONNECT path.)

## Fix

When a CONNECT request is **not** intercepted (the `shouldInterceptConnect` false branch) **and** an outbound proxy is configured for that target, establish the tunnel through the proxy instead of a direct `net.connect`, reusing `outboundHttpProxyAgent` so `HTTPS_PROXY`/`https_proxy`/`HTTP_PROXY`/`http_proxy` and `NO_PROXY` semantics match the existing fetch chaining exactly. The existing piping, `head` replay, and 502/close semantics are preserved.

With no outbound proxy configured, or when `NO_PROXY` bypasses the host, the code path is the **unchanged** direct `net.connect` — no behaviour change for existing users.

## Verification

- Added two focused tests in `tests/http-proxy-server.test.ts`: the tunnel goes through a configured upstream proxy (the upstream sees the CONNECT), and with no proxy configured it still dials direct.
- `pnpm vitest run tests/http-proxy-server.test.ts`: 41 passed, 1 failed — the single failure (`Anthropic passthrough retry`, a 10 s timing-sensitive case) also fails on clean `main` (`f1c74ac`), so it is pre-existing and unrelated.
- Proven end-to-end on macOS: `claude → clodex (this patch) → mitmproxy --map-local robots.txt → internet` serves the mapped `robots.txt` for a non-Anthropic host while the root path and `api.anthropic.com` MITM behave exactly as before.


bman654 (changes requested):
Good find, and the fix is the right shape — the CONNECT path genuinely was the one place in this module that didn't chain through `outboundHttpProxyAgent`. Two things to address before I can take it.

## Blocker — a self-pointing `HTTPS_PROXY` recursively tunnels CONNECT through clodex until FDs are exhausted

The new branch at `src/http-proxy/server.ts:1184` reuses the outbound-proxy *mechanism* but not the loop guard that goes with it. The Anthropic passthrough path calls `proxyUrlTargetsListener` before it builds an agent (`src/http-proxy/server.ts:1276`) precisely because, as the header comment in `src/outbound-proxy.ts` puts it, "a server can still inherit a previously exported bridge URL from its shell." The new CONNECT path skips that check, so when the configured proxy *is* this listener, every non-intercepted CONNECT tunnels into clodex, which is again non-intercepted, and so on.

Measured against a pristine build of this branch, driving the real `clodex server` command:

```
curl: (56) CONNECT tunnel failed, response 502
baseline_fds=19   peak_fds=8154   elapsed_ms=2176
```

The identical script against `origin/main` returns `http_code=200` in 171 ms. A separate bounded reproduction observed nine recursive inbound connections for a single target connection before it cleared the environment.

This is reachable without anything exotic: `clodex server --proxy` prints `HTTPS_PROXY=http://127.0.0.1:<port>` as setup instructions (`src/http-proxy/index.ts:152`) and the port is fixed by default (`DEFAULT_SERVER_PORT = 17645`, `src/constants.ts:58`). Export that line as told, then restart the server from the same shell, and it inherits a proxy URL addressing its own listener. Ordinary `clodex claude` launches don't self-target, since that listener URL only reaches the child's environment — this is `clodex server`-specific. The process does recover once EMFILE unwinds into a 502, so it's transient rather than permanently wedged, but it's still an availability regression on a path that works today.

Reusing `proxyUrlTargetsListener` here, with a regression test, should be all it takes.

## Should-fix — the two added tests don't discriminate

The upstream test proxy at `tests/http-proxy-server.test.ts:397` only acknowledges the CONNECT; no tunnel data is ever exchanged. Nothing downstream of the handshake is pinned, so almost any mutation of the new code stays green:

| Mutation | Result |
| --- | --- |
| Delete both `pipe()` calls | 42/42 passed |
| Delete CONNECT head replay | 42/42 passed |
| Drop the `proxyConnectStatus` guard | 42/42 passed |
| Drop the client-close cleanup | 42/42 passed |
| Drop `sockets.add(connected)` | 42/42 passed |
| Revert the feature entirely | 1 timeout, 41 passed |

Only the first test detects anything, and only route selection; the second stays green because it covers the pre-existing direct path.

Two hand-built predicates do discriminate, and confirm the implementation is correct as written — worth promoting into the committed tests:

- Replay a `head` and then exchange bytes in both directions. Passes as-is; fails under either the pipe or the head-replay mutation.
- Have the upstream proxy answer the CONNECT with `407`. The client sees exactly `502 Bad Gateway`; with the `proxyConnectStatus` guard removed it wrongly sees `200 Connection Established`.

## Nits (not blocking)

- A proxy that accepts TCP and then sends neither bytes, FIN, nor RST leaves one pending outbound socket after both client and proxy shut down (`PENDING_OUTBOUND_AFTER_CLOSE=1`, `destroyed=false`). Late responses, non-200s, and errors all clean up correctly, and the CLI's explicit `process.exit` keeps shutdown from hanging. Real bookkeeping regression from the direct path, too narrow to hold the PR for.
- A malformed proxy URL used to warn once at startup; per-CONNECT agent construction now repeats `clodex: HTTP(S)_PROXY cannot be used for a CONNECT tunnel...` indefinitely in standalone-server logs.

## Checked and refuted — no action needed

So these don't get chased on a re-roll, the following were all examined and are fine:

- Passing an `IncomingMessage` where `HttpsProxyAgent.connect()` expects a `ClientRequest` is safe on the pinned `https-proxy-agent@9.1.0`; it only uses `emit()` and `once()`.
- `proxyConnect` fires synchronously before `connect()` settles, for both 200 and non-200 responses. Parse failures reject, and a missing event fails closed.
- The non-200 early return destroys the real proxy socket inside `https-proxy-agent`; the ordinary catch paths don't leak.
- `secureEndpoint: false` is correct for a raw tunnel.
- Head replay and the client-facing `200` are correctly ordered — verified at the byte level.

Fix the guard, promote those two predicates into tests, and I think this is ready.

marcelstenkewitz:
Both addressed in 22b7959.

## Blocker — self-pointing `HTTPS_PROXY`

Reproduced your measurement, then fixed it the way you suggested. The CONNECT path now
resolves the proxy URL with `outboundProxyUrlForTarget` first and runs it through
`proxyUrlTargetsListener` against the bound listener before it builds an agent — the same
order the raw Anthropic passthrough uses. On a match it falls through to the unchanged
direct `net.connect` and says so once:

```
clodex: HTTP(S)_PROXY points at this proxy; tunneling CONNECT direct
```

The listener address is captured right after `listenTcpServer` resolves, so it is set
before any connection can be accepted.

Regression test: `dials CONNECT targets directly when HTTPS_PROXY names this proxy` starts
the bridge on a reserved port, exports that exact URL, and asserts the tunnel works,
`HttpsProxyAgent.prototype.connect` was never called, and the line above was logged. With
the guard removed it fails (recursion, never reaches the echo).

## Should-fix — the tests now discriminate

Both hand-built predicates are committed, and the weak pair is gone. The upstream test
proxy really tunnels now (`startTunnelingProxy`), and the target echoes `echo:<chunk>` so
the two directions are separable.

| Mutation | Test that fails |
| --- | --- |
| Remove the loop guard | `dials CONNECT targets directly when HTTPS_PROXY names this proxy` |
| Delete both `pipe()` calls | `routes passthrough CONNECT tunnels through HTTPS_PROXY, head and both directions` |
| Delete CONNECT head replay | same |
| Drop the `proxyConnectStatus` guard | `answers 502 when the outbound proxy refuses the CONNECT` |
| Build the agent per CONNECT again | `warns once about an unusable proxy URL and keeps tunneling direct` |

The 407 case asserts the client sees `502 Bad Gateway` and *not* `200 Connection
Established`, so the guard can't be dropped silently. Reverting the feature entirely still
fails the first test, since the upstream proxy records no authority.

## Nits

- **Repeated malformed-proxy warning — fixed.** The agent is now memoized per resolved
  proxy URL rather than constructed per CONNECT (`agent.connect()` opens a fresh socket on
  every call, so sharing is safe), and the cached agents are destroyed in `close()`
  alongside the passthrough agent. One warning per unusable URL, as before this branch.
- **`PENDING_OUTBOUND_AFTER_CLOSE=1` — left alone, deliberately.** That socket is owned
  inside `https-proxy-agent` while `connect()` is still pending, so there is nothing for
  `close()` to reach without reaching into the library. Agreed it is too narrow to hold
  the PR for; happy to take a follow-up if you'd rather have it tracked.

## Verification

Node 24.14.1, pnpm 10.34.5.

- `pnpm typecheck` clean, `pnpm build` succeeded.
- `pnpm vitest run tests/http-proxy-server.test.ts`: 44 passed, 1 failed — still only
  `Anthropic passthrough retry > replays a request whose pooled upstream socket was reset
  before any reply`. Confirmed pre-existing on a worktree at `f1c74ac`: 39 passed, same 1
  failed.
- `pnpm test`: 2,913 passed, 3 failed. The two beyond the above are
  `responses-websocket > new-connection pacing`, also failing on `f1c74ac` (197 passed, 2
  failed).
- Each mutation in the table was applied to a clean tree and run individually; every one
  turned its intended test red.

bman654 (approved):
Round 2 clears round 1's blocker, and the rewritten tests are the real thing. I reproduced your
mutation table row for row — **6/6 hold**, and the full-revert row is stronger than you claimed:
it reds all four new CONNECT tests, not one. The loop guard itself is verified against the real
`clodex server` command: one non-intercepted CONNECT went from **peak 29,562 descriptors at
`91841f6` to 59 at `22b7959`**, and both guard mutations (delete it, widen it) turn the new tests
red.

Approving and merging. Two small things came out of the panel that I applied myself on your branch
(`20912f5`) rather than sending you round three — both are written up below so the reasoning is on
the record. Everything else I looked at is cleared, and I've written that out too so it doesn't get
chased again.

## Applied 1 — the guard was not armed yet when the listener starts accepting

`src/http-proxy/server.ts:1307` assigns `listenerAddress` *after* `await listenTcpServer(...)`
returns. But `listenTcpServer` resolves the `listen()` callback first (`src/listener-ready.ts:126-142`)
and only then runs a real TCP probe round trip (lines 144-151). The socket is already accepting
throughout that await, so a CONNECT arriving in that window is handled with
`listenerAddress === undefined`, which makes `selfTargeting` false and switches the guard off.

Measured over 1,000 real `listenTcpServer` calls, the window is sub-millisecond
(p50 0.146 ms, p99 0.332 ms, max 2.517 ms — the 1000 ms constant is a failure budget, not the
natural width). With a client retrying at event-loop speed across a restart, 4 of 100 starts had a
root CONNECT select the outbound agent inside the window, and the deepest natural recursion from one
client was 2 proxy hops. So this is narrow, and it is not reachable from `clodex claude` (which
awaits `startHttpProxy` before spawning) or by a first-time or discovered client — it needs a
fixed-port standalone server restarting while a configured client is already retrying.

I fixed it anyway because it is one line: the handler now reads `proxyServer.address()` instead of
the cached variable, and the late assignment is gone. The handler cannot run before the server is
listening, so a live CONNECT always has an address. Measured before the change: 30 starts × 400
racers executed 242 root handlers while the cache was still unset, and after it **zero** of them
selected an outbound agent.

The new `tests/http-proxy-connect-guard-startup.test.ts` pins it. A natural window is ~0.15 ms, far
too tight to hit reliably, so the test holds it open deliberately — a mocked `listenTcpServer` runs
the real one and then blocks, leaving the socket accepting while `startHttpProxy` is still pending.
Restoring the cached-address version turns it red.

## Applied 2 — both new handler-time warnings were swallowed on `clodex claude`

`src/http-proxy/server.ts:1208-1213` uses a bare `console.error`. `launchClaude` replaces
`process.stdout.write` and `process.stderr.write` for the child's lifetime (`src/launch.ts:44-64`),
and it does so *after* `startConfiguredHttpProxy` returns (`src/cli.ts:1217` vs `:1285`).

I checked the obvious defence — that the sibling Anthropic-passthrough warning right below is also a
bare `console.error` — and it does not hold: that one fires during startup at `server.ts:1309-1319`,
before the mute is installed, so it is visible. Anything emitted from the CONNECT handler is not.
Driving the real `clodex claude` with a fake child that opens one non-intercepted CONNECT and an
inherited malformed proxy URL: the terminal showed **1** warning (the startup one), while
`--debug-file` captured the handler-time one as a swallowed `[parent] ...` line.

This is the fourth time this class has bitten us, which is why `src/parent-notice.ts` exists. Both
now go through `emitParentNotice`: the self-target warning, and the malformed-proxy-URL warning in
`outboundHttpProxyAgent`, since the handler can now trigger that one too. The three tests that
asserted them moved from a `console.error` spy to the notice channel; putting either warning back on
`console.error` reds them.

## Checked and cleared — no action needed

- **Sharing one agent across CONNECTs is safe, as your comment says.** 12 concurrent `.connect()`
  calls yielded 12 distinct sockets (129 ms, versus 1,440 ms if they serialised), and the agent's
  `sockets`/`freeSockets`/`requests` stay empty because `.connect()` bypasses agent-base's pooling.
  Keying the cache on the resolved proxy URL while constructing from the target URL is sound — the
  agent is a pure function of the proxy URL across ten target shapes (ports, IPv6, punycode,
  NO_PROXY, embedded credentials), and `HTTP_PROXY` is unreachable here since the target is always
  `https://`.
- **`agent?.destroy()` in `close()` is inert but harmless.** Node's `Agent#destroy` only walks
  `freeSockets`/`sockets`, both empty, so it destroys 0 of 12 live tunnels — the `sockets` sweep is
  the real cleanup. Good news half: it does *not* tear down tunnels that are still piping. The `?.`
  is load-bearing and is pinned.
- **The 200-then-RST path is handled.** A hostile proxy that answers 200 and then resets is consumed
  by `connected.once('error')` with `tunnelEstablished=true`, destroys the client, and raises no
  uncaught exception — six consecutive clean runs.
- **No ambient-proxy dependence.** The describe-level `beforeEach` clears all six proxy variables;
  results are identical under a dead proxy, a live local proxy, and none. No `Date.now()` ties, no
  digest-pin standing in for an assertion, no vacuous passes, no fixed-port collisions.
- **Your reported failures don't reproduce here, and this branch isn't the cause either way.** Head
  is 45/45 across 11 consecutive runs, base `f1c74ac` is 40/40 across 3, and the full suite is
  2,916 passed / 0 failed. Your three look like load flakes rather than stable pre-existing facts —
  nothing for you to do.
- **`PENDING_OUTBOUND_AFTER_CLOSE=1` is unchanged by round 2.** An in-flight `agent.connect()` at
  `close()` still leaks its upstream socket because `upstream` is only assigned in `.then()`.
  Measured identically at `91841f6`, so the memoization neither caused nor worsened it, and it adds
  no new leak. Still too narrow to hold the PR for, exactly as I said in round 1.
- **Going direct rather than refusing is the right fallback.** It matches the sibling passthrough,
  and refusing would break all non-Anthropic web traffic for no benefit.

## Non-blocking notes

- **Lifecycle test parity (follow-up, not a reroll).** The teardown/error half of the new branch is
  still unprotected: deleting the client-close upstream destroy, `connected.once('close')`, the
  trailing `.catch` 502, or the `tunnelEstablished` assignment each leave the suite 45/45 green, and
  none of those is an equivalent mutant. The direct branch *does* have sibling tests for three of
  them. I'm not blocking on it: the shipped code is correct on every one of those paths, `main`'s own
  direct branch is equally unprotected on the error paths (deleting its entire `upstream.once('error')`
  handler also stays 45/45), and round 1 named two predicates, which you promoted along with three
  more. If you want to close the biggest part cheaply, parameterize the two existing direct teardown
  tests over both branches.
- **The `warnedConnectSelfProxy` latch has no test.** Deleting it leaves 45/45 green; the test named
  "warns once" pins the *other* warning, the malformed-URL one latched by the agent cache. Worth a
  line if you revisit this. Gotcha if you write it: startup emits a different message sharing the
  prefix `clodex: HTTP(S)_PROXY points at this proxy;`, so match the full string, not a prefix.
- **The guard matches loopback spellings, not destinations** (`src/outbound-proxy.ts:98-111`). A
  proxy URL naming a loopback *alias* — `[::ffff:127.0.0.1]`, `localhost.`, or an `/etc/hosts` name
  like `kubernetes.docker.internal` — misses the guard and reproduces the blow-up. **No ask**, and
  it should not hold this PR: reaching it requires copying clodex's own URL, keeping its port, and
  independently rewriting the host to an alias; nothing clodex prints, stores, or documents ever
  emits one. And it isn't new — the same unchanged predicate already permitted an equivalent
  descriptor-exhausting loop on `main`'s Anthropic passthrough under the complete printed
  environment (2,052 descriptors under a 2,048 cap). I'll file it as a follow-up. If anyone picks it
  up: do **not** add a post-`agent.connect()` `remoteAddress` check — the recursion has already
  entered the handler while that promise is still pending, so it cannot help.

PR #263, opened 2026-09-21
feat(opencode-go): route Responses-only Go models (Muse Spark) through @ai-sdk/openai

TweedBeetle:
OpenCode Go serves `muse-spark-1.2-contributor` and `muse-spark-1.3-contributor` only on `/v1/responses`. Chat completions and the Anthropic Messages path both answer 500 for them, and Go's own docs list them under `@ai-sdk/openai`. Today the catalog has no way to say that, so the two Muse models are unreachable through clodex.

This adds an `openai-responses` transport to the catalog updater and wires the two pieces that transport needs.

**Pinning a model's package separately from the provider's.** The provider record's pin decides what discovery and refresh will accept, and it still refuses `@ai-sdk/openai`, which I did not widen. So `openCodeGoPinnedModelApiUrl` is a second, wider lookup used for a catalog model's package: a model carrying `@ai-sdk/openai` resolves to the Go `/v1` base, everything else falls through to the existing provider pin.

**Sending the base URL on API-key `@ai-sdk/openai` routes.** `createLanguageModel` passed `baseURL` on the OAuth branch only. Without it on the API-key branch, `@ai-sdk/openai` defaults to `api.openai.com`, so the request goes to OpenAI carrying the OpenCode Go key. That is the half worth a look: it sends a third party's key to the wrong host, and it does so silently, because the response is a normal OpenAI error.

Tests: `tests/provider-factory.test.ts` gains a case pinning the base URL on the API-key branch, and `tests/opencode-go.test.ts` covers the new transport through the updater and the catalog invariants.

Verified live against an isolated proxy on the two Muse models: a single tool call, multi-turn recall after a tool call, and two tool calls in one turn, all 200 through `opencode-go`. Since then both models have run as my daily Claude Code alias for about a week on top of #252 and #253.

Environment: macOS 26.6.2 (arm64), Node 22.23.2, pnpm 10.34.5. The four test files touching this area: 136 passed.


bman654 (approved):
The defect this fixes is real and I confirmed it at the byte level: on `main`, an API-key route
whose package is `@ai-sdk/openai` ignores its configured base URL, so a local listener standing in
for the SDK sees `POST api.openai.com` carrying the OpenCode Go key. At this head the key goes only
to the Go host. Good catch, and the PR body describes it accurately rather than overstating it.

The catalog data also checks out completely. Reconstructing the 2026-09-13 models.dev feed and
re-running your updater regenerates `src/data/opencode-go-models.json` **byte-for-byte identical**
to the committed blob (sha `cb07b5dc…` both sides), so nothing was hand-edited; a semantic diff of
base versus head shows 18 → 20 models, two added, **zero changed, zero removed**, order preserved.
Both Muse entries' context window, pricing, reasoning flag, modalities, family and upstream id match
the upstream records.

Approving and merging. Two small things came out of the panel that I applied myself on your branch
(`12e6b9e`) rather than sending it back to you; both are written up below.

## Applied 1 — three docs stated the opposite of what this ships

The PR makes Responses-only models reachable and updates no documentation, so all three of these are
now false:

- `docs/opencode-go.md:15` — "The provider uses one credential with **two** upstream wire protocols",
  followed by a two-bullet list. There are three now; the third is
  `https://opencode.ai/zen/go/v1/responses`.
- `docs/opencode-go.md:26` — "Clodex **intentionally excludes** those entries from this provider …
  The supported catalog contains **only** Anthropic Messages and Chat Completions models."
- `README.md:305` — "OpenCode Go exposes its Anthropic Messages and Chat Completions models;
  Responses-only entries are intentionally excluded."

Worth getting right beyond tidiness: a wrong claim in our own docs has twice propagated into a later
contributor PR. The protocol list now has a third entry naming `/v1/responses`, and the scope
paragraph says a model reaches the catalog once its transport is verified — and that a catalog entry
declares the Responses path by carrying `@ai-sdk/openai`, while the provider record still names
`@ai-sdk/openai-compatible`. `docs/opencode-go.md:40` needed no change; it already anticipated
Responses as a routing change.

## Applied 2 — pinned the reasoning-effort value

`compatibility.supportsReasoningEffort: false` had no test — flipping it to `true` left 130/130
green. I checked whether the value is right, and it is: `false` is the only self-consistent setting,
and it is also **inert on the wire**, because `effortProviderOptions` admits only recognised
OpenAI/Codex families and Muse matches neither, so the translated request is byte-identical either
way. The one real delta is a field in the debug `GET /models` dump. Added one assertion per entry in
`tests/opencode-go.test.ts`; flipping the flag now reds it.

## Checked and cleared — no action needed

- **Widening the model-level pin does not widen anything that carries a credential.** The provider
  record's pin is untouched: `refreshApiListProvider` and `fetchTemplateModels` both key on the
  *provider's* npm and still refuse `@ai-sdk/openai`, verified against hand-built registries with a
  drifted `templateId`, an attacker-controlled `api.url`, and a forged per-model `apiUrl`. The forged
  per-model URL is inert because `openCodeGoPinnedModelApiUrl` returns the pinned constant and
  ignores it; I traced every other consumer of that field and none reads the forged value.
- **A refresh does not silently un-fix this.** Running a refresh against a stubbed `/models` endpoint
  preserves the Muse entries and their Responses transport rather than dropping them or rewriting
  their package back.
- **The mixed state this PR creates — an `@ai-sdk/openai` model under an `@ai-sdk/openai-compatible`
  provider record — stays coherent** across pricing, free-model classification, context handling, and
  the picker/alias/patch surfaces.
- **`reasoning: true` with `supportsReasoningEffort: false` is handled correctly**: classified
  internal-only rather than non-reasoning, patch and default effort controls omitted, and an alias
  cannot re-enable effort because the effort gate reads the catalog's `upstreamModelId`, not the
  inbound alias text. Thinking-block round-tripping through the `@ai-sdk/openai` path restores item
  id and encrypted content.
- **OpenCode Go rejects any request without `x-opencode-session`, and the new route carries it.** The
  header is attached at the routing layer (`src/proxy.ts:659-663`) independently of which package the
  route uses, so it rides the Responses path too — consistent with your live runs succeeding.
- **The new `baseURL` forward is safe for everything that is not Muse.** I chased one real regression
  shape here and then talked myself out of it, so it doesn't get raised later: an `@ai-sdk/openai`
  API-key record storing a *root* URL without `/v1` would now post to `/responses` and 404, where
  `main` silently used the SDK default. It is not reachable — all 60 released tags pin
  `https://api.openai.com/v1`, no `add`/`edit`/`import`/`--base-url` surface can create that shape,
  and the historical importer that could shipped only in upstream relay-ai, wrote to a directory
  clodex never reads, and couldn't emit that URL anyway. More to the point, every *third-party* root
  URL in that class was `main` sending that provider's key to `api.openai.com` — which this PR fixes.
  Only the hand-edited canonical-OpenAI spelling would regress, and special-casing that host would be
  a narrowing predicate around a configuration nobody has observed. Leaving it alone is right.
- **The updater's guardrails still hold**: null-prototype maps (`constructor`, `__proto__` etc. stay
  unmapped and get reported), unmapped feed models reported and excluded, and an effort map naming a
  value the feed does not publish still aborts before any write.
- **Unknown transport strings fall through to Chat Completions silently.** Not an ask — `main`
  already turns a one-character typo in `'anthropic-messages'` into an openai-compatible entry at
  exit 0, so this is pre-existing posture, and the regenerate-then-commit workflow reds three tests
  either way.

